PR #345, opened 2022-09-06
Fix the counter sample

swankjesse:
We lost automatic registration of the iosMain directory when we
introduced the samples/ subproject. This tracks the way we did
similar in the emoji-search example.


oldergod (approved):

JakeWharton:
The cause was actually #337 which stopped using the `ios()` helper as it does not include the ARM simulator target we need.

JakeWharton (approved):

PR #350, opened 2022-09-08
Use lambda receiver for children scope

### LINKED ISSUE #336: Switch children properties back to using lambdas
Now that we're not using the declared schema for testing and are going to take another approach (#335), we should restore the schema to using lambda types to define children. For unscoped children this change has no value, but for scoped children we can change this:
```kotlin
@Children(1, MyScope::Class) children: () -> Unit,
```
into this:
```kotlin
@Children(1) children: MyScope.() -> Unit,
```
which is much more natural and reflective of how it works in generated output.

JakeWharton:
Closes #336.

### DIFF
--- a/redwood-cli/src/test/kotlin/app/cash/redwood/generator/ComposeGenerationTest.kt
+++ b/redwood-cli/src/test/kotlin/app/cash/redwood/generator/ComposeGenerationTest.kt
@@ -33,7 +33,7 @@ class ComposeGenerationTest {
 
   @Widget(1)
   data class Row(
-    @Children(1, RowScope::class) val scoped: () -> Unit,
+    @Children(1) val scoped: RowScope.() -> Unit,
     @Children(2) val unscoped: () -> Unit,
   )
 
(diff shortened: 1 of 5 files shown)


colinrtwhite:
@JakeWharton Were you planning to tackle this for `@LayoutModifier` as well or can I give that a shot?

JakeWharton:
Hmm I didn't think there was a way to change how it works. What were you thinking?

redwood-cli/src/test/kotlin/app/cash/redwood/generator/ComposeGenerationTest.kt
```diff
@@ -33,7 +33,7 @@ class ComposeGenerationTest {
 
   @Widget(1)
   data class Row(
-    @Children(1, RowScope::class) val scoped: () -> Unit,
+    @Children(1) val scoped: RowScope.() -> Unit,
```
adrw, line 36: Nice

adrw (approved):

colinrtwhite:
@JakeWharton Sorry - nevermind. I was thinking it would be:

```kotlin
@LayoutModifier(1)
@RowScope
public data class RowVerticalAlignment(
  /** -1 for top, 0 for middle, 1 for bottom. */
  val direction: Int,
)
```

But of course that wouldn't work since `RowScope` isn't an annotation class in the schema file.

JakeWharton:
We could force them to be. Right now they can be literally anything because we only care about the name. Would be easy to make this work, although we need a way to know which annotation is the scope (a meta `@Scope` annotation on `@RowScope`?).

PR #352, opened 2022-09-08
Move JVM protocol widget test code to commonMain.

colinrtwhite:
Saw there was a TODO for this!


JakeWharton (approved):

JakeWharton:
Nice. Thanks. I think the test code that was preventing this moved to another module.

PR #356, opened 2022-09-08
Update the maven publish plugin. Disable auto component creation.

colinrtwhite:


gradle.properties
```diff
@@ -15,6 +15,7 @@ android.defaults.buildfeatures.aidl=false
 android.defaults.buildfeatures.renderscript=false
 android.defaults.buildfeatures.resvalues=false
 android.defaults.buildfeatures.shaders=false
+android.disableAutomaticComponentCreation=true
```
colinrtwhite, line 18: This disables some legacy behaviour from pre-AGP 7.1 where it would automatically set up publishing for all build variants (I think). It isn't necessary when using the new `publishing` syntax.

JakeWharton (approved):

PR #357, opened 2022-09-08
Move Compose Widget.Children implementation to a common module

JakeWharton:


JakeWharton:
Fun fact: this is NOT specific to Compose UI so it could be used with a UI toolkit like Mosaic or Glance!

JakeWharton:
Shit does that mean it needs to be multiplatform too...

colinrtwhite (approved):

PR #359, opened 2022-09-09
Work-around Android Kotlin multiplatform source sets warning.

colinrtwhite:
Works-around this warning (emitted by the multiplatform plugin) by removing the extra source sets added by AGP 7.2.1:

```
The following Kotlin source sets were configured but not added to any Kotlin compilation:
 * androidAndroidTestRelease
 * androidTestFixtures
 * androidTestFixturesDebug
 * androidTestFixturesRelease
You can add a source set to a target's compilation by connecting it with the compilation's default source set using 'dependsOn'.
See https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#connecting-source-sets
```

Also fixes a minor typo `Pressenter` -> `Presenter`.


build.gradle
```diff
@@ -132,4 +132,20 @@ allprojects {
       }
     }
   }
+
+  // Work-around: https://youtrack.jetbrains.com/issue/KT-48436
```
JakeWharton:
```suggestion
  // Work-around: https://youtrack.jetbrains.com/issue/KT-48436
  // TODO Remove when bumping to Kotlin 1.7.20.
```

JakeWharton: This will cause an issue to be filed to track removing the workaround.

JakeWharton (approved):

.gitignore
```diff
@@ -17,3 +17,6 @@ xcuserdata
 
 # Cocoapods
 Pods
+
+# Mac
+**/.DS_Store
```
JakeWharton: Please move this to a global gitignore on your computer. It's far easier than adding it to every repo.

https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

colinrtwhite:
Haven't merged this since it passes CI, but causes a syncing issue locally. Going to take another quick look today.

JakeWharton:
We could also just wait for Kotlin 1.7.20 which should only be a few weeks away.

colinrtwhite:
Yep, that's a better idea. Going to close this out. Gradle: 1, Colin 0.

PR #362, opened 2022-09-09
Always clear old children before the initial diff

swankjesse:
As-is we need this boilerplate in every implementation of beforeUpdatedCode().

### DIFF
--- a/redwood-treehouse/src/engineMain/kotlin/app/cash/redwood/treehouse/RealViewBinder.kt
+++ b/redwood-treehouse/src/engineMain/kotlin/app/cash/redwood/treehouse/RealViewBinder.kt
@@ -15,6 +15,7 @@
  */
 package app.cash.redwood.treehouse
 
+import app.cash.redwood.protocol.ChildrenDiff.Companion.RootChildrenTag
 import app.cash.redwood.protocol.Diff
 import app.cash.redwood.protocol.EventSink
 import app.cash.redwood.protocol.widget.DiffConsumingWidget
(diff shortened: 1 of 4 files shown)


redwood-treehouse/src/engineMain/kotlin/app/cash/redwood/treehouse/RealViewBinder.kt
```diff
@@ -76,6 +76,7 @@ public class RealViewBinder(
         scope.launch(dispatchers.main) {
           if (firstDiff) {
             firstDiff = false
+            view.protocolDisplayRoot.children(0)!!.clear()
```
JakeWharton: This should be 1, not 0, because it's a tag and not an index. There's a constant for this at `ChildrenDiff.RootChildrenTag` which is used by things like `ProtocolDisplay`.

JakeWharton (approved):

redwood-treehouse/src/engineMain/kotlin/app/cash/redwood/treehouse/RealViewBinder.kt
JakeWharton: Hmmm looks like we unconditionally return the children from our current implementations of the root view. I would prefer if we change those as well to only return when `RootChildrenTag` is requested and to throw otherwise. That change can be separate, but switching to `RootChildrenTag` in this change can still be done.

swankjesse: Perfect

swankjesse: Changed all of our `ProtocolDisplayRoot` implementations to use `RootChildrenTag`.

JakeWharton (approved):
Double approve!

PR #364, opened 2022-09-09
Make the existing MutableListChildren more useful

JakeWharton:
- Implements MutableList as its name suggests (previously it just wrapped one)
- Provide a callback whenever it changes via the Widget.Children mechanism


colinrtwhite (approved):

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/MutableListChildren.kt
```diff
@@ -17,12 +17,23 @@ package app.cash.redwood.widget
...
+ */
 public class MutableListChildren<T : Any>
 @JvmOverloads constructor(
-  public val list: MutableList<T> = mutableListOf(),
-) : Widget.Children<T>, Iterable<T> {
+  private val list: MutableList<T> = mutableListOf(),
+  private val onUpdate: (List<T>) -> Unit = {},
+) : Widget.Children<T>, MutableList<T> by list {
+  /** @param onUpdate Callback invoked when contents change via the [Widget.Children] API. */
+  public constructor(onUpdate: (List<T>) -> Unit) : this(mutableListOf(), onUpdate = onUpdate)
```
swankjesse, line 32: necessary for parameter order?

swankjesse (approved):

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/MutableListChildren.kt
JakeWharton: Primarily for exposing to iOS where things like default argument values (as used in primary constructor) do not work. Although I have not yet tested that it shows up in the header properly.

PR #366, opened 2022-09-13
Improve iOS TreehouseView

JakeWharton:


redwood-treehouse/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIKitView.kt
```diff
@@ -16,37 +16,53 @@
...
-  public val view: UIView = RootUiView(frame = cValue { CGRectZero })
-  private var treehouseApp: TreehouseApp<T>? = null
+  public val view: UIView = RootUiView(this)
+  private var content: TreehouseView.Content<T>? = null
 
-  // TODO(jwilson): track when this view is detached from screen
-  override val boundContent: TreehouseView.Content<T>? = content
+  override val boundContent: TreehouseView.Content<T>?
+    get() {
+      return when {
```
adrw, line 35: `return view.superview?.let { content }`?

adrw (approved):

redwood-treehouse/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIKitView.kt
JakeWharton: Yes! I coped this style from the existing Android one so that they match, for now. There's definitely more terse ways to do this that I will refactor all of them to use in a follow-up.

PR #367, opened 2022-09-13
Remove `ViewBinder`, rename `ViewBinder.Adapter` to `ViewBinder`

JakeWharton:
The `Adpater` is what people implement to bind to native UI.

The old `ViewBinder` was an abstraction with a single implementation and which was not used for testing despite its docs saying so. Moreover, the layer at which it intercepted behavior was not an extremely useful one. This likely was the result of the many refactorings necessary to move this API into the public project.


swankjesse (approved):

swankjesse:
I like it.

Issue #328, opened 2022-08-28
Upstream reusable row/col/box layouts into Redwood module


JakeWharton:
These are foundational things that no one wants to build. Plus it would simplify our samples.

colinrtwhite:
@JakeWharton Is the plan still to implement these in `commonMain`? In existing samples we use native bindings to create these widgets. Also should these live in `redwood-runtime` or `redwood-widget-compose`?

JakeWharton:
Yes. We have a multiplatform paging library that we need to get over the finish line (cc @veyndan) and then it comes down to designing how building the integration into Redwood works.

My original idea was to have a widget which used events to communicate back to the composable when we reach half way to the end of a page so that it could load the next page in that direction. That would re-deliver the list of items back to the UI with the additional data. If you scroll too fast we would have to put placeholders into the UI while waiting for the data.

In the talk I designed the API basically the same way as Compose UI:
```kotlin
@Composable
fun MessageList(messages: List<Message>) {
    LazyColumn {
        items(messages) { message ->
            MessageRow(message)
        }
    }
}
```
We'll know the number of items so we can send that to the UI on first load which will power the `RecyclerView` / Compose UI `LazyColumn` / `UITableView` along with the first page of data. Then, as you scroll, the widget requests more pages which the composable uses to execute the item lambda for those items and then we add those to the list and send it back to the UI to be bound.

But we have basically nothing implemented at the Redwood layer so that's all a guess as to how it can work.

veyndan:
> We have a multiplatform paging library that we need to get over the finish line (cc `@veyndan`) and then it comes down to designing how building the integration into Redwood works.

The priority is currently:

1. Get pagination and `LazyColumn` working internally. The Treehouse `LazyColumn` implementation is helping me refine the best workaround for many of the KMP issues that I'm facing (the issues have already been filed).
2. I'll then work on a public release of multiplatform-paging, and then a public release of `LazyColumn` within Treehouse. Granted that there's nothing technical preventing a public release of multiplatform-paging — just time and priority.

Obviously can rejig the priority if it makes sense.

> My original idea was to have a widget which used events to communicate back to the composable when we reach half way to the end of a page so that it could load the next page in that direction. That would re-deliver the list of items back to the UI with the additional data. If you scroll too fast we would have to put placeholders into the UI while waiting for the data.

My current prototype sends down a serialized instance of `androidx.paging.Pager`. This works around a couple of issues with just doing incremental paging loads based on scroll position:

[...]

PR #372, opened 2022-09-16
Create multiplatform FlexboxEngine implementation.

colinrtwhite:
This PR introduces a `FlexboxEngine` that supports measuring and laying out "nodes" with support for most [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) properties. The code is a port of Google's [Flexbox-layout](https://github.com/google/flexbox-layout) with significant changes to remove all Android/JVM dependencies.

The plan is to use `FlexboxEngine` as an implementation dependency in a future PR to add support for multiplatform [`Column` and `Row` implementations](https://github.com/cashapp/redwood/issues/328).

I still need to clean up references to views and add **a lot** more tests to this PR, but wanted to post this for early feedback.

Refs #328


redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+ */
+package app.cash.redwood
+
+import kotlin.js.JsName
+
+/**
+ * Create a new [FlexboxEngine] instance.
+ */
+@JsName("newFlexboxEngine")
+public fun FlexboxEngine(): FlexboxEngine = RealFlexboxEngine()
```
swankjesse: Nice

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+  public var alignItems: AlignItems
+
+  /**
+   * The padding of the flexbox.
+   */
+  public var padding: Spacing
+
+  /**
+   * The current value of the maximum number of flex lines.
+   * If not set, [Undefined] is returned.
```
swankjesse: [Undefined] --> `-1` ?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Returns the nodes contained in the flexbox.
+   */
+  public val nodes: List<Node>
+
+  /**
+   * Adds the node to the specified index of the flexbox.
+   */
+  public fun addNode(node: Node, index: Int = nodes.size)
```
swankjesse: Can we make `nodes` a `MutableList` and get rid of these?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
```diff
@@ -0,0 +1,198 @@
...
+   */
+  public val measuredWidth: Int
+
+  /**
+   * The measured height after invoking [measure].
+   */
+  public val measuredHeight: Int
+
+  /**
+   * This is called to find out how big this node should be.
```
swankjesse: Can you elaborate? Maybe, ‘After calling this the measuredWidth and measuredHeight attributes will be set.’ ?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * The margin of the node.
+   */
+  public val margin: Spacing
+
+  /**
+   * The measured width after invoking [measure].
+   */
+  public val measuredWidth: Int
```
swankjesse, line 153: Add docs to describe what happens if I invoke this before calling measure()?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
```diff
@@ -0,0 +1,198 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood
+
+/**
+ * An interface that has the common behavior as a node contained in a flexbox.
+ */
+public interface Node {
```
swankjesse, line 21: Can we make this into a class? Perhaps with a user-defined tag value?

swankjesse: I assume the only reason this can’t be an interface today is so we can have a polymorphic `measure` function. But perhaps that can be an external adapter thing?

One thing that I’d like to try is separating the value-object bits from the service bits. If I apply my value/service/glue ideal, this is a platypus. I wonder if we could split out responsibilities differently.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
```diff
@@ -0,0 +1,198 @@
...
+  public val maxHeight: Int
+
+  /**
+   * True if this item is visible and should be laid out.
+   */
+  public val visible: Boolean
+
+  /**
+   * The baseline used for [AlignItems.Baseline] and [AlignSelf.Baseline].
+   * If not specified, [Undefined] is set as a default value.
```
swankjesse: Undefined is -1 ?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
```diff
@@ -0,0 +1,198 @@
...
+   */
+  public val baseline: Int
+
+  /**
+   * The order attribute of the node.
+   *
+   * The attribute can change the order in which the children are laid out.
+   * By default, children are displayed and laid out in the same order as they added to the
+   * [FlexboxEngine]. If not specified, [DefaultOrder] is set as a default value.
```
swankjesse, line 81: Is it better to just say -1 here?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
```diff
@@ -0,0 +1,198 @@
...
+   */
+  public val order: Int
+
+  /**
+   * The flex grow attribute of the node.
+   *
+   * The attribute determines how much this child will grow if positive free space is
+   * distributed relative to the rest of other nodes included in the same flex line.
+   * If not specified, [DefaultFlexGrow] is set as a default value.
```
swankjesse, line 90: Is it better to say 0f here?

swankjesse: It’s also weird for an interface to specify its default value... this feels like a class to me.

swankjesse: ...Do users ever compare anything against `DefaultFlexGrow` ? Or is it just a value they use when creating a new implementation?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
swankjesse: One of the major challenges of this API is the mix of in and out params.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/properties.kt
```diff
@@ -0,0 +1,294 @@
...
+
+/**
+ * This attribute controls the alignment along the cross axis.
+ * The alignment in the same direction can be determined by the
+ * [AlignItems] attribute in the parent, but if this is set to
+ * other than [AlignSelf.Auto], the cross axis alignment is
+ * overridden for this child.
+ */
+@JvmInline
+public value class AlignSelf private constructor(public val ordinal: Int) {
```
swankjesse: I never use value classes. So cool.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/utils.kt
```diff
@@ -0,0 +1,109 @@
...
+
+import app.cash.redwood.Node.Companion.MatchParent
+import app.cash.redwood.Node.Companion.WrapContent
+import kotlin.math.max
+import kotlin.math.min
+
+internal const val Undefined = -1
+
+internal fun packLong(lower: Int, higher: Int): Long {
+  return (higher.toLong() shl 32) or (lower.toLong() and 0xFFFFFFFF)
```
swankjesse: supernit: nicer to do a trailing `L` here, like `0xFFFFFFFFL`. It’s gonna be a long in bytecode and the only reason you get that for free is 'cause it’s bigger than max int.

redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FakeNode.kt
```diff
@@ -0,0 +1,51 @@
...
+
+  override var measuredWidth: Int = -1
+  override var measuredHeight: Int = -1
+
+  override fun measure(widthSpec: MeasureSpec, heightSpec: MeasureSpec) {
+    measuredWidth = MeasureSpec.resolveSize(width, widthSpec)
+    measuredHeight = MeasureSpec.resolveSize(height, heightSpec)
+  }
+
+  override fun layout(left: Int, top: Int, right: Int, bottom: Int) {}
```
swankjesse, line 50: I would have expected this to mutate width and height

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/properties.kt
```diff
@@ -0,0 +1,294 @@
...
+public value class AlignContent private constructor(public val ordinal: Int) {
+
+  override fun toString(): String = when (ordinal) {
+    0 -> "FlexStart"
+    1 -> "FlexEnd"
+    2 -> "Center"
+    3 -> "SpaceBetween"
+    4 -> "SpaceAround"
+    5 -> "Stretch"
+    else -> "Unknown: $ordinal"
```
JakeWharton:
```suggestion
    else -> throw AssertionError()
```
(For all these value classes below, too)

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/properties.kt
```diff
@@ -0,0 +1,294 @@
...
+    public val FlexEnd: AlignContent = AlignContent(1)
+    public val Center: AlignContent = AlignContent(2)
+    public val SpaceBetween: AlignContent = AlignContent(3)
+    public val SpaceAround: AlignContent = AlignContent(4)
+    public val Stretch: AlignContent = AlignContent(5)
+
+    public fun valueOf(ordinal: Int): AlignContent {
+      require(ordinal in 0..5) { "unknown AlignContent: $ordinal" }
+      return AlignContent(ordinal)
+    }
```
JakeWharton: How about public constructor and move this guard to an `init` block (for all these value classes below, too).

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+import kotlin.js.JsName
+
+/**
+ * Create a new [FlexboxEngine] instance.
+ */
+@JsName("newFlexboxEngine")
+public fun FlexboxEngine(): FlexboxEngine = RealFlexboxEngine()
+
+/**
+ * A class that measures and positions its children according to flexbox properties.
```
JakeWharton: This is an interface not a class, but I think I want to advocate for it being a class. There will never be a second implementation, there's no reason to mock/fake this because it operates purely on values, and the behavior is deterministic.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
JakeWharton: The `measure` lambda is effectively a property just like the other properties in this type. When the `flexSpace` property changes, the node is dirty and needs recompouted. When the `measure` lambda changes, the node is dirty and needs recomputed.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
JakeWharton: This one is a public constant, at least, whereas `Undefined` isn't (although maybe it should be). In general I think I prefer pointing users towards constant names rather than values because it's easy to resolve a constant to its value but it's harder to reverse lookup a value to a constant you can rely on.

swankjesse:
@JakeWharton Good point on the benefit of having explicit constants. I’m mostly torn between the discoverability benefits of `Undefined` vs. the pervasive pattern of `-1` to mean an undefined `Int`. When I see `Undefined` I’m inclined so assume there’s gonna be a bunch of other magic values like [WRAP_CONTENT](https://developer.android.com/reference/android/view/ViewGroup.LayoutParams#WRAP_CONTENT). If there’s only ever gonna be one special value, -1 is nice.

swankjesse:
And yes, FlexboxEngine doesn’t get much value from having an interface.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
colinrtwhite: Yep! Maybe better to get rid of the constant?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
colinrtwhite: +1 I'd like to have `measure` return a `Size` (like `FlexboxEngine.measure`) instead of holding onto some mutable state, but it was a bigger change to the code and I want to ensure I have good regression tests set up before making that change. After we get rid of `measuredWidth` and `measuredHeight` we'll only have "in" properties.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
colinrtwhite: Yep! I'm going to inline the constant.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
colinrtwhite: Yep, the API currently doesn't have a way to mark nodes as dirty. It assumes a node's properties don't change once it's added to the layout. The plan is to add this later after adding some more regression tests.

colinrtwhite:
Re: constants discussion I think it's worthwhile to keep `DefaultFlexGrow` and the others since those are mandated defaults by the flexbox spec. Semi-related, but Compose UI defines `DefaultAlpha=1.0`, which imo is useful since I don't need to remember the constant's value. I'm going to inline `Undefined`, though since that's specific to the implementation and -1 is a common default for not set properties.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/ObservableMutableList.kt
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+package app.cash.redwood
+
+internal class ObservableMutableList<T>(
+  private val delegate: MutableList<T> = mutableListOf(),
+  private val onChange: () -> Unit = {},
+  // Delegate to List-only so new mutable methods have to be implemented manually.
+) : MutableList<T>, List<T> by delegate {
```
JakeWharton: Does `AbstractMutableList` help reduce needing to implement some of these functions yourself?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/ObservableMutableList.kt
```diff
@@ -0,0 +1,72 @@
...
+    delegate.clear().also { onChange() }
+  }
+
+  override fun iterator() = delegate.iterator()
+
+  override fun listIterator() = delegate.listIterator()
+
+  override fun listIterator(index: Int) = delegate.listIterator(index)
+
+  override fun subList(fromIndex: Int, toIndex: Int) = delegate.subList(fromIndex, toIndex)
```
JakeWharton: I think this needs wrapped, too. Otherwise mutations to the sublist will be missed.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/ObservableMutableList.kt
colinrtwhite: Yep! It works great, thanks.

redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FlexboxEngineTest.kt
```diff
@@ -0,0 +1,395 @@
...
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+
+class FlexboxEngineTest {
+
+  private lateinit var engine: FlexboxEngine
+
+  @BeforeTest
+  fun before() {
+    engine = FlexboxEngine().apply {
```
JakeWharton:
```suggestion
  private val engine = FlexboxEngine().apply {
```

colinrtwhite:
@swankjesse Feel free to merge this in if you want to iterate on it as well. I still have a few changes I'd like to make, but can do them in follow-up PRs:

- Refactor `Node.measure` to return a `Size` and delete `measuredWidth`/`measuredHeight`.
- Convert `Node` to a concrete class.
- Observe changes to `Node`'s properties to mark them automatically as dirty.
- Many more tests.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+    var indexInFlexLine = 0
+    var flexLine = Line()
+    flexLine.firstIndex = fromIndex
+    flexLine.mainSize = mainPaddingStart + mainPaddingEnd
+    val childCount = nodes.size
+    for (i in fromIndex until childCount) {
+      val child = getReorderedChildAt(i)
+      if (child == null) {
+        if (isLastFlexItem(i, childCount, flexLine)) {
+          addFlexLine(flexLines, flexLine, i, sumCrossSize)
```
swankjesse, line 206: I was expecting this?

```
flexLine = Line()
```

(surprising that it accumulates an element without creating a new element)

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+
+  /**
+   * The computed flex lines after calling [measure].
+   */
+  internal var flexLines = listOf<Line>()
+
+  /**
+   * Holds the reordered indices after [Node.order] has been taken into account.
+   */
+  private var indexToReorderedIndex: IntArray? = null
```
swankjesse, line 83: If we have our own book-keeping class we can avoid all of these ugly parallel arrays. In particular, rather than keeping reorderedIndex/frozen/measureSpec/etc. as their own arrays, we can make each of these properties on our per-node book-keeping class.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+            getFlexItemMarginStartMain(child, false) +
+            getFlexItemMarginEndMain(child, false),
+          childDimension = childMainSize,
+        )
+        child.measure(childCrossMeasureSpec, childMainMeasureSpec)
+        updateMeasureCache(i, childCrossMeasureSpec, childMainMeasureSpec, child)
+      }
+
+      // Check the size constraint after the first measurement for the child
+      // To prevent the child's width/height violate the size constraints imposed by the
```
swankjesse, line 266: supernit: I think this `To` should be lowercase? It’s part of the preceding sentence and not a new sentence??

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
swankjesse: Ahhh. . . . it’s an optimization because it’s the last line. I suspect `continue` aught to be `break`?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+      // {@link Node#getMinWidth()}, {@link Node#getMinHeight()},
```
swankjesse, line 267: supernit: Javadoc `@link` was never doing anything here; needs to be in `/** ... */` comments (which tools interpret). Syntax can be simplified a touch for Kotlin.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+            getFlexItemMarginEndCross(child, true) + sumCrossSize,
+          childDimension = getFlexItemSizeCross(child, true),
+        )
+        child.measure(childMainMeasureSpec, childCrossMeasureSpec)
+        updateMeasureCache(i, childMainMeasureSpec, childCrossMeasureSpec, child)
+      } else {
+        childCrossMeasureSpec = MeasureSpec.getChildMeasureSpec(
+          spec = crossMeasureSpec,
+          padding = crossPaddingStart + crossPaddingEnd +
+            getFlexItemMarginStartCross(child, false) +
```
swankjesse, line 250: I suspect it’d be quite profitable to have an enum or constant for `HORIZONTAL` and `VERTICAL`; these `true` and `false` things are bug magnets.

(Plus if we used an enum those things could be polymorphic!)

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+          flexLinesSize = flexLines.size,
+        )
+      ) {
+        if (flexLine.itemCountVisible > 0) {
+          addFlexLine(flexLines, flexLine, if (i > 0) i - 1 else 0, sumCrossSize)
+          sumCrossSize += flexLine.crossSize
+        }
+        if (isMainAxisHorizontal) {
+          if (child.height == MatchParent) {
+            // This case takes care of the corner case where the cross size of the
```
swankjesse, line 289: The original comments here are generally very good to understand what’s happening! But I feel like this one is weak. No action here, but if we really went to town on this class I’d like to rewrite this, perhaps with ASCII-art

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+              padding = padding.top + padding.bottom + child.margin.top +
+                child.margin.bottom + sumCrossSize,
+              childDimension = child.height,
+            )
+            child.measure(childMainMeasureSpec, childCrossMeasureSpec)
+            checkSizeConstraints(child, i)
+          }
+        } else {
+          if (child.width == MatchParent) {
+            // This case takes care of the corner case where the cross size of the
```
swankjesse, line 307: Off-topic: what kind of monster copy-pastes a 6-line comment? (This is from the original code obvs.)

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+        if (flexWrap != FlexWrap.WrapReverse) {
+          flexLine.maxBaseline = max(
+            flexLine.maxBaseline,
+            child.baseline + child.margin.top,
+          )
+        } else {
+          // if the flex wrap property is WRAP_REVERSE, calculate the
+          // baseline as the distance from the cross end and the baseline
+          // since the cross size calculation is based on the distance from the cross end
+          flexLine.maxBaseline = max(
```
swankjesse, line 368: In a future pass I think we change this to `maxOf()` ?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+        // Let the sumCrossSize start from the negative value of the last flex line's
+        // cross size because otherwise flex lines aren't calculated enough to fill the
+        // visible area.
+        sumCrossSize = -flexLine.crossSize
+        reachedToIndex = true
+      }
+      if (sumCrossSize > needsCalcAmount && reachedToIndex) {
+        // Stop the calculation if the sum of cross size calculated reached to the point
+        // beyond the needsCalcAmount value to avoid unneeded calculation in a
+        // RecyclerView.
```
swankjesse, line 393: We’ll generalize this to avoid using the word `RecyclerView`

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+        break
+      }
+    }
+    return flexLines
+  }
+
+  /**
+   * Returns an array with container's padding in the format: [mainStart, mainEnd, crossStart, crossEnd].
+   */
+  private fun getPadding(isMainAxisHorizontal: Boolean): IntArray {
```
swankjesse, line 407: This temp allocation is most likely going to be a performance tragedy ! (No action to take here; just whining)

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+    } else {
+      intArrayOf(padding.top, padding.bottom, padding.start, padding.end)
+    }
+  }
+
+  /**
+   * Returns the view's measured size in the main axis. Either width or height.
+   *
+   * @param node the view
+   * @param isMainHorizontal is the main axis horizontal
```
swankjesse, line 419: Don’t make me tap the sign
https://publicobject.com/2007/08/05/stop_writing_redundant_javadoc/

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+  }
+
+  /**
+   * Checks if the view's width/height don't violate the minimum/maximum size constraints imposed
+   * by the [Node.minWidth], [Node.minHeight], [Node.maxWidth] and [Node.maxHeight] attributes.
+   *
+   * @param node the view to be checked
+   * @param index index of the view
+   */
+  private fun checkSizeConstraints(node: Node, index: Int) {
```
swankjesse, line 564: I hate this function. In particular, I hate that its name is prefixed `check()` but it neither returns a boolean nor does it throw an exception if the check fails. I think this needs a new name; maybe `measureIfNecessary()` ?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+        } else {
+          getLargestMainSize()
+        }
+        paddingAlongMainAxis = (padding.top + padding.bottom)
+      }
+      else -> error("Invalid FlexDirection: $flexDirection")
+    }
+    var flexLineIndex = 0
+    if (indexToFlexLine != null) {
+      flexLineIndex = indexToFlexLine!![fromIndex]
```
swankjesse, line 635: For follow-up, can do Kotlin things to avoid `!!` after a null check

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+    while (i < size) {
+      val flexLine = flexLines[i]
+      if (flexLine.mainSize < mainSize && flexLine.anyItemsHaveFlexGrow) {
+        expandFlexItems(
+          widthMeasureSpec = widthMeasureSpec,
+          heightMeasureSpec = heightMeasureSpec,
+          flexLine = flexLine,
+          maxMainSize = mainSize,
+          paddingAlongMainAxis = paddingAlongMainAxis,
+          calledRecursively = false,
```
swankjesse, line 648: `calledRecursively = false` really intrigues me.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+        shrinkFlexItems(
+          widthMeasureSpec = widthMeasureSpec,
+          heightMeasureSpec = heightMeasureSpec,
+          flexLine = flexLine,
+          maxMainSize = mainSize,
+          paddingAlongMainAxis = paddingAlongMainAxis,
+          calledRecursively = false,
+        )
+      }
+      i++
```
swankjesse, line 660: This can probably be converted to a for loop?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+    }
+  }
+
+  private fun ensureChildrenFrozen(size: Int) {
```
swankjesse, line 664: Yuck, yeah these manual parallel arrays are so ugly

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+  ) {
+    if (flexLine.totalFlexGrow <= 0 || maxMainSize < flexLine.mainSize) {
+      return
+    }
+    val sizeBeforeExpand = flexLine.mainSize
+    var needsReexpand = false
+    val unitSpace = (maxMainSize - flexLine.mainSize) / flexLine.totalFlexGrow
+    flexLine.mainSize = paddingAlongMainAxis
+
+    // Setting the cross size of the flex line as the temporal value since the cross size of
```
swankjesse, line 701: temporal --> temporary? 

(what’s temporal mean here?)

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+   * @param node      the View to be stretched
+   * @param crossSize the cross size
+   * @param index     the index of the view
+   */
+  private fun stretchViewHorizontally(node: Node, crossSize: Int, index: Int) {
+    var newWidth = crossSize - node.margin.start - node.margin.end
+    newWidth = max(newWidth, node.minWidth)
+    newWidth = min(newWidth, node.maxWidth)
+    val measuredHeight = if (measuredSizeCache != null) {
+      // Retrieve the measured height from the cache because there
```
swankjesse, line 1366: Oooooh this is very interesting. Looks like there’s a case we might want to track where the `View` object gets replaced with another one, and we prefer cache data?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -0,0 +1,2041 @@
...
+  /**
+   * Returns a View, which is reordered by taking [Node.order] parameters into account.
+   *
+   * @param index the index of the view
+   * @return the reordered view, which [Node.order] is taken into account.
+   * If the index is negative or out of bounds of the number of contained views, returns `null`.
+   */
+  private fun getReorderedChildAt(index: Int): Node? {
+    if (indexToReorderedIndex == null) {
+      val sorted = nodes.withIndex().sortedWith(compareBy({ -it.value.order }, { it.index }))
```
swankjesse, line 2032: This will likely be another needless performance bomb . . . . accessing each child sorts all the nodes? Yuck

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/ObservableMutableList.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood
+
+internal class ObservableMutableList<T>(
```
swankjesse, line 18: Neato

redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FlexboxEngineTest.kt
```diff
@@ -0,0 +1,389 @@
...
+ */
+package app.cash.redwood
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+
+class FlexboxEngineTest {
```
swankjesse, line 24: These are our tests? Do we have any of Google’s tests?

redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/testUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood
+
+import kotlin.test.assertTrue
+
+internal fun assertAlmostEquals(expected: Int, actual: Int, tolerance: Int = 1) {
```
swankjesse, line 20: SCREAMS

swankjesse (approved):

swankjesse:
Merging 'cause although I have comments, it’s much better for all of us to iterate on trunk vs. this branch!

redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FlexboxEngineTest.kt
colinrtwhite: @swankjesse They're pretty far off Google's code, but good point. I'll update the header.

Issue #361, opened 2022-09-09
Restore all native targets


JakeWharton:
Once https://github.com/JetBrains/androidx/pull/292 ships. Refs #284 and #299.

JakeWharton:
Okay so after #2070 we can probably make the following changes:

- `Common` + `CommonWithAndroid` to gain Mac OS, Linux, Mingw, tvOS, watchOS, and Wasm JS.
- `TreehouseGuest` to gain Wasm JS.
- `ToolkitComposeUi` to gain Wasm JS.

PR #373, opened 2022-09-16
JetBrains Compose runtime beta01

JakeWharton:
Drop their custom repo which is no longer needed as this now on central (finally).


JakeWharton:
Unblocks #361

swankjesse (approved):

PR #375, opened 2022-09-17
Move `redwood-gradle-plugin` to a subproject

### LINKED ISSUE #297: Attempt to pull our plugin into its own Gradle project
It only depends on the version which we can pull separately. This will allow us to use the plugin on the test schema which is used to test the various first-party libraries. In order to make this work, we'll need dependency substitution (similar to the sample's includeBuild mapping) so that we point to the in-project modules, but otherwise it _should_ work. Complicates publishing a bit, but that will all be done by CI so who cares. As the plugin functionality gets more complicated, duplicating implementation is becoming more and more of a pain (see, for example, #287).

JakeWharton:
This allows it to be used both within our build for the test-schema but also published as an artifact for external users. The samples can also now be part of the main build rather than using includeBuild which avoids some bugs with including across Kotlin multiplatform projects.

Closes #297


JakeWharton:
I have done some serious crimes in this PR and I love them so much. Just need one final sanity check tomorrow and we're good to go.

redwood-gradle-plugin/build.gradle
```diff
@@ -1,8 +1,16 @@
 apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'java-gradle-plugin'
-apply plugin: 'com.vanniktech.maven.publish'
 apply plugin: 'com.github.gmazzo.buildconfig'
 
+// This module is included in two projects:
+// - In the root project where it's released as one of our artifacts
+// - In build-support project where we can use it for the test-schema and samples.
+//
+// We only want to publish when it's being built in the root project.
```
swankjesse, line 9: SERIOUS CRIMES

swankjesse (approved):

PR #382, opened 2022-09-26
Add a test that lays out strings visually

swankjesse:
Testing flexbox should be easy. With this its a picnic.

### DIFF
--- a/redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FlexboxTest.kt
+++ b/redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FlexboxTest.kt
@@ -1,0 +1,158 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood
+
+import app.cash.redwood.FlexDirection.Companion.Row
+import app.cash.redwood.FlexDirection.Companion.RowReverse
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class FlexboxTest {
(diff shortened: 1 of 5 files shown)


redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FlexboxTest.kt
```diff
@@ -0,0 +1,156 @@
...
+ */
+package app.cash.redwood
+
+import app.cash.redwood.FlexDirection.Companion.Row
+import app.cash.redwood.FlexDirection.Companion.RowReverse
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class FlexboxTest {
+  private val imdbTop4 = listOf(
```
swankjesse, line 24: Definitely open to suggestions for another theme here. I want 3-5 short phrases with variable word length.

redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FlexboxTest.kt
```diff
@@ -0,0 +1,156 @@
...
+ * limitations under the License.
+ */
+package app.cash.redwood
+
+import app.cash.redwood.FlexDirection.Companion.Row
+import app.cash.redwood.FlexDirection.Companion.RowReverse
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class FlexboxTest {
```
swankjesse, line 23: This test isn’t exhaustive! But I think it’s more real than the FakeNode one we inherited! In particular this is starting to prove out which parameters are inputs and which ones are outputs.

redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/FlexboxTest.kt
```diff
@@ -0,0 +1,156 @@
...
+
+  private fun FlexboxEngine.layout(
+    width: Int,
+    height: Int,
+  ): String {
+    val canvas = StringCanvas(width, height)
+    val widthSpec = MeasureSpec.from(width, MeasureSpecMode.Exactly)
+    val heightSpec = MeasureSpec.from(height, MeasureSpecMode.Exactly)
+
+    flexLines = when (flexDirection) {
```
swankjesse, line 144: This mutates the FlexBoxEngine. I don’t think that requiring callers to do this is a good API. We should fix that. Also callers shouldn’t need to interrogate the flex direction to call the right function!

JakeWharton (approved):
Probably want to rebase on #383, although may want to wait until its mutable property follow-up.

swankjesse:
Rebased!

colinrtwhite (approved):
This is awesome 🙏

PR #383, opened 2022-09-26
Convert Node to a concrete class.

colinrtwhite:
Converts `Node` from an interface to a concrete class. At the moment only a couple of its properties are mutable, but longer term all of them will be.


JakeWharton (approved):
What's the plan for when these values change over time? Are the properties going to become mutable?

JakeWharton:
Oh you say that in the PR description 🤦

swankjesse (approved):

PR #386, opened 2022-09-26
Parameterize Node on its widget type

swankjesse:


redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
```diff
@@ -158,20 +145,20 @@ public class Node(
...
+  public var top: Int = -1
+  public var right: Int = -1
+  public var bottom: Int = -1
 
   /**
-   * A callback to place the node inside the given `left`, `top`, `right`, and `bottom` coordinates.
+   * Assign the node's position.
    */
-  public var layout: (left: Int, top: Int, right: Int, bottom: Int) -> Unit = { _, _, _, _ -> }
+  public fun layout(left: Int, top: Int, right: Int, bottom: Int) {
```
swankjesse, line 156: Open question – who should own this state? It’s definitely an output of the layout engine and so there’s a simple case to be made that this function belongs on `FlexboxEngine.Adapter` instead.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Node.kt
```diff
@@ -18,7 +18,8 @@ package app.cash.redwood
 /**
  * A node with properties that can be measured and laid out inside a flexbox.
  */
-public class Node(
+public class Node<T>(
+  public val widget: T,
```
swankjesse, line 22: My hypothesis is that the responsibilities are better separated like so:

 - Widget: Opaque object from the native widget system.
 - Node: Flexbox-specific configuration of each widget.
 - Adapter: Answers measurement questions about a widget.

The work in front of us is to decide what should be state on Node (such as flexGrow) and what should be queried  (already `minWidth()`, perhaps `visible()` and others?).

JakeWharton:
I remain opposed to this is a design choice. The library makes no use of this reference in a way that cannot also be accomplished by capturing a reference into the measure lambda.

swankjesse:
@JakeWharton yeah, @colinrtwhite and I discussed offline. We’re gonna split node’s responsibilities.

swankjesse:
Discussed offline further! Punting this for now.

PR #391, opened 2022-09-27
Split Measurable from Node

swankjesse:
The state related to measurements should be its own type.

### DIFF
--- a/redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
+++ b/redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
@@ -17,9 +17,9 @@
 
 package app.cash.redwood
 
+import app.cash.redwood.Measurable.Companion.MatchParent
 import app.cash.redwood.Node.Companion.DefaultFlexBasisPercent
 import app.cash.redwood.Node.Companion.DefaultFlexGrow
-import app.cash.redwood.Node.Companion.MatchParent
 import app.cash.redwood.Node.Companion.UndefinedFlexShrink
 import kotlin.math.roundToInt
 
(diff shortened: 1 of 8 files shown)


redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Measurable.kt
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood
+
+public open class Measurable {
```
swankjesse, line 18: This has a naming conflict with Compose. I assume a typealias on that import is enough? Suggestions for other names also welcome!

redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/StringWidget.kt
```diff
@@ -21,29 +21,20 @@ package app.cash.redwood
...
-) {
+) : Measurable() {
   private val words = text.split(" ")
 
   private var left = -1
   private var top = -1
   private var right = -1
   private var bottom = -1
 
-  // TODO: fix the Node API to not conflate layout parameters with measurements.
```
swankjesse, line 32: Fixing this todo is the big motivation here. StringWidget does all the measurement, Node does all the layout parameters.

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/Measurable.kt
colinrtwhite, line 18: I like this! Later we can add `FlexProperties` so `Node` will just be properties + measurable + layout.

colinrtwhite (approved):

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
```diff
@@ -17,9 +17,9 @@
 
 package app.cash.redwood
```
JakeWharton, line 18: All of these types need moved into a `flexbox` subpackage at some point.

JakeWharton (approved):

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/FlexboxEngine.kt
JakeWharton: https://github.com/cashapp/redwood/issues/395

PR #398, opened 2022-09-27
Move flexbox classes into a flexbox subpackage.

### LINKED ISSUE #395: Flexbox module should use a `.flexbox` subpackage

colinrtwhite:
Fixes: https://github.com/cashapp/redwood/issues/395


redwood-flexbox-engine/src/commonTest/kotlin/app/cash/redwood/flexbox/FlexboxEngineTest.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Google Inc. All rights reserved.
+ * Copyright (C) 2022 Square, Inc.
```
JakeWharton: Are we retaining these? Need to update Spotless config and revert if so.

JakeWharton (approved):
Thanks!

PR #399, opened 2022-09-27
Renames Direction -> Orientation, Node -> FlexNode.

colinrtwhite:


redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/flexbox/Orientation.kt
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.flexbox
+
+internal fun FlexDirection.toOrientation(): Orientation {
```
colinrtwhite, line 18: Github wants to consider this a new file even though it was `git mv`'d unfortunately.

JakeWharton (approved):

JakeWharton:
Can you squash / auto-squash these multi-commit PRs instead of merge / auto-merge? Multiple commits (whose intermediate ones fail to build) make doing a `git bisect` much more challenging.

colinrtwhite:
@JakeWharton Ah sorry that was my intention! Should we disable "Allow merge commits" under settings as well? I only have "Allow squash merging" enabled in my personal repos.

JakeWharton:
My biggest problem with that is it breaks stacking PRs. I truly dislike every option GitHub provides. They all have some fault in what they do which breaks some workflow. That being said, I have quite a few repos set up the same way (squash only) as it's the safest bet despite it breaking things like `--merged` on the CLI.

JakeWharton:
Okay squash only is on here for now. Let's see how it goes.

PR #402, opened 2022-09-28
Add LazyColumn

veyndan:
There are still a few things to resolve, but the `LazyColumn` API is identical to that of Compose UI's!

[demo.webm](https://user-images.githubusercontent.com/6900601/192835829-2d4b4c2e-1603-4342-b4f2-7d3354c33059.webm)

#### Still to resolve:

- [ ] We need to display a placeholder for each element while the elements are being loaded, otherwise we end up trying to load [all the elements](https://developer.android.com/jetpack/compose/lists#avoid-0-size-items). A placeholder isn't necessarily bad, but having it hardcoded in the client probably isn't a good idea.
- [x] There's a bug where when we try to scroll to the top, where it thinks we've hit the top, but when we try to scroll up again, it shows there are more items (demonstrated in the beginning of the video).
- [ ] There are other parameters to `items(…)` in Compose UI that we'll want to support, most importantly `key`. A snag with this is that Compose UI expects `key` to be passed synchronously. The current architecture of the code doesn't support that yet.
- [ ] Pull out the LazyColumn stuff to a separate module.

I've not used paging here, as we're not actually paging anything. I have a local branch where I added pagination, but it was just going through one long trip through the pagination API without getting any benefit from it.


samples/emoji-search/android/src/main/java/app/cash/zipline/samples/emojisearch/ComposeUiLazyColumn.kt
```diff
@@ -0,0 +1,97 @@
...
+import example.schema.widget.LazyColumn
+import example.values.IntervalList
+import example.values.LazyListIntervalContent
+import example.values.MutableIntervalList
+
+class ComposeUiLazyColumn<T : Any>(
+  treehouseApp: TreehouseApp<T>,
+  widgetFactory: AndroidEmojiSearchWidgetFactory<T>,
+) : LazyColumn<@Composable () -> Unit> {
+  private var intervals by mutableStateOf<IntervalList<LazyListIntervalContent>>(MutableIntervalList())
```
swankjesse: This can’t be a mutableStateList, right?

samples/emoji-search/android/src/main/java/app/cash/zipline/samples/emojisearch/ComposeUiLazyColumn.kt
```diff
@@ -0,0 +1,97 @@
...
+    LazyColumn(
+      horizontalAlignment = Alignment.CenterHorizontally,
+      modifier = Modifier
+        .fillMaxWidth(),
+    ) {
+      itemsIndexed(intervals) { _, localIntervalIndex, interval ->
+        val item = interval.value.item
+        var shown by remember { mutableStateOf(false) }
+        if (!shown) Text("Waiting…", Modifier.size(64.dp))
+        AndroidView(
```
swankjesse: For follow up it’d be nice to do this without any views anywhere

samples/emoji-search/presenters/src/commonMain/kotlin/app/cash/zipline/samples/emojisearch/serializers.kt
```diff
@@ -0,0 +1,91 @@
...
+    return MutableVector(content.size) { content[it] }
+  }
+
+  override fun serialize(encoder: Encoder, value: MutableVector<LazyListIntervalContent>) {
+    val content = List(value.size) { value[it] }
+    serializer.serialize(encoder, content)
+  }
+}
+
+object MutableVectorIntervalLazyListIntervalContentSerializer : KSerializer<MutableVector<IntervalList.Interval<LazyListIntervalContent>>> {
```
swankjesse: That’s a  .. . long class name

samples/emoji-search/presenters/src/jsMain/kotlin/app/cash/zipline/samples/emojisearch/LazyColumn.kt
```diff
@@ -0,0 +1,79 @@
...
+  ) : LazyListIntervalContent.Item {
+
+    override fun get(index: Int): ZiplineTreehouseUi {
+      val treehouseUi = object : TreehouseUi {
+        @Composable
+        override fun Show() {
+          content(index)
+        }
+      }
+      return treehouseUi.asZiplineTreehouseUi(DiffProducingEmojiSearchWidgetFactory(json), widgetVersion = 1u)
```
swankjesse: For follow-up, lets make `DiffProducingEmojiSearchWidgetFactory` composition-scoped so you can pluck it out of the environment.

samples/emoji-search/values/src/commonMain/kotlin/example/values/IntervalList.kt
```diff
@@ -0,0 +1,211 @@
...
+ * limitations under the License.
+ */
+package example.values
+
+import androidx.compose.runtime.collection.MutableVector
+import androidx.compose.runtime.collection.mutableVectorOf
+import kotlinx.serialization.Serializable
+
+/**
+ * The list consisting of multiple intervals.
```
swankjesse: This is in emoji-search . . . . is that right long-term?

swankjesse: I guess right now we’re pinned in that package until we have Redwood schemas that can import other schemas

samples/emoji-search/values/src/commonMain/kotlin/example/values/IntervalList.kt
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * It is useful when you want to define your own dsl similar to the one used by LazyColumn where
+ * list items can be defined via multiple item/items calls.
+ *
+ * This interface is read only, in order to create a list you need to use [MutableIntervalList].
+ *
+ * @param T type of values each interval contains in [Interval.value].
+ */
+sealed interface IntervalList<T> {
```
swankjesse: I wonder if there’s a better name for this thing. Aggregate or Concatenated both come to mind.

swankjesse: I only dislike the current name as it has `List` as a suffix, but doesn’t extend from the List interface.

samples/emoji-search/values/src/commonMain/kotlin/example/values/IntervalList.kt
```diff
@@ -0,0 +1,211 @@
...
+   * Allows to iterate through all the intervals starting from the one containing [fromIndex]
+   * until the one containing [toIndex].
+   *
+   * @param fromIndex we will start iterating from the interval containing this index.
+   * @param toIndex the last interval we iterate through will contain this index. This index
+   * should be not smaller than [fromIndex].
+   *
+   * @throws IndexOutOfBoundsException if the indexes are not within 0..[size] - 1 range.
+   */
+  override fun forEach(fromIndex: Int, toIndex: Int, block: (IntervalList.Interval<T>) -> Unit) {
```
swankjesse: Would it be simpler for callers for this to implement `Iterable<Interval>` ? Then they can iterate without necessarily using a for loop to do so

samples/emoji-search/values/src/commonMain/kotlin/example/values/IntervalList.kt
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }
+
+  override fun get(index: Int): IntervalList.Interval<T> {
+    checkIndexBounds(index)
+    return getIntervalForIndex(index)
+  }
+
+  private fun getIntervalForIndex(itemIndex: Int): IntervalList.Interval<T> {
+    val lastInterval = lastInterval
```
swankjesse: I don’t like that `last` here means _most recently requested_ which is different from _end of list_

Maybe use the word _cached_ ?

samples/emoji-search/values/src/commonMain/kotlin/example/values/LazyColumnAdapter.kt
```diff
@@ -0,0 +1,31 @@
...
+ */
+package example.values
+
+import app.cash.redwood.treehouse.ZiplineTreehouseUi
+import app.cash.zipline.ZiplineService
+import kotlinx.serialization.Contextual
+import kotlinx.serialization.Serializable
+
+@Serializable
+class LazyListIntervalContent(
```
swankjesse, line 24: I _think_ we might be able to inline this now that we have `@Contextual` parameters on `ZiplineService`.

samples/emoji-search/presenters/src/commonMain/kotlin/app/cash/zipline/samples/emojisearch/serializers.kt
```diff
@@ -0,0 +1,91 @@
...
+
+val treehouseSerializersModule = SerializersModule {
+  contextual(ziplineServiceSerializer<LazyListIntervalContent.Item>())
+  contextual(MutableVectorLazyListIntervalContentSerializer)
+  polymorphic(IntervalList::class) {
+    subclass(MutableIntervalListSerializer)
+  }
+}
+
+object MutableVectorLazyListIntervalContentSerializer : KSerializer<MutableVector<LazyListIntervalContent>> {
```
swankjesse: These can all be `internal` ?

samples/emoji-search/values/src/commonMain/kotlin/example/values/IntervalList.kt
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  private fun IntervalList.Interval<T>.contains(index: Int) = index in startIndex until startIndex + size
+}
+
+/**
+ * Finds the index of the interval which contains the highest value of
+ * [IntervalList.Interval.startIndex] that is less than or equal to the given [itemIndex].
+ */
+private fun <T> MutableVector<IntervalList.Interval<T>>.binarySearch(itemIndex: Int): Int {
```
swankjesse: Maybe this? Fewer tests to write!
```
return asMutableList().binarySearchBy(...)
```

samples/emoji-search/values/src/commonMain/kotlin/example/values/IntervalList.kt
swankjesse: (Fine to keep it as-is for now)

swankjesse (approved):
# SHIP IT

samples/emoji-search/values/src/commonMain/kotlin/example/values/IntervalList.kt
veyndan: It's definitely not right long term — it's already one of my TODOs in the PR description:

> Pull out the LazyColumn stuff to a separate module.

samples/emoji-search/android/src/main/java/app/cash/zipline/samples/emojisearch/ComposeUiLazyColumn.kt
veyndan: Much of this PR has been looking through how `LazyColumn` is implemented in Compose UI. They use an `IntervalList` (which I had to copy to our codebase as it's in `androidx.compose.foundation.lazy.layout` yet is KMP-compatible). I'm looking into ways of solving keys (per my TODOs in the PR description) as in Android it assumes that keys are retrieved synchronously (and not in a composable).

samples/emoji-search/values/src/commonMain/kotlin/example/values/IntervalList.kt
veyndan: This class is a copy-paste of `androidx.compose.foundation.lazy.layout.IntervalList` so I'll probably keep everything as it's what `LazyColumn` in Compose UI is using under the hood too. I thought I changed the license of this file to Android's, but I think the `.idea/copyright/profiles_settings.xml` keeps on overwriting it as soon as I leave the file…

100% agree with the `List` suffix not being great as we're not extending from `List` though. I just think it might be more confusing having differing implementations of the class.

samples/emoji-search/values/src/commonMain/kotlin/example/values/LazyColumnAdapter.kt
veyndan: It's not just a wrapper class this time! I'm still planning to shove stuff in here. I'm attempting to mimic that of https://github.com/androidx/androidx/blob/99014a24aac1ec28a82081ae741b15f8df883a0d/compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/lazy/LazyListScopeImpl.kt#L76-L81

swankjesse (approved):

veyndan:
I'll finish up the remaining unresolved things in follow up PRs. I don't think it's necessary to block the PR for them.

samples/emoji-search/presenters/src/jsMain/kotlin/app/cash/zipline/samples/emojisearch/LazyColumn.kt
veyndan: Replaced `JsonProvider` with `DiffProducingWidgetFactoryProvider`.

veyndan:
To prevent a runtime error when running the iOS version of EmojiSearch, we must wait for https://github.com/cashapp/zipline/pull/737 to be released.

PR #403, opened 2022-09-28
Create redwood-layout with common layout properties.

colinrtwhite:
Pulling these into a separate PR from the implementations to make that PR cleaner.


redwood-layout/README.md
```diff
@@ -0,0 +1,102 @@
...
+    return then(HorizontalAlignmentLayoutModifier(alignment))
+  }
+}
+```
+
+Then copy the following classes into the `commonMain` of your `widget` module. These classes connect
+the generated schema with the platform `RowWidget` and `ColumnWidget` implementations:
+
+```kotlin
+class RowBridge<T : Any>(private val delegate: RowWidget<T>) : Row<T>, Widget<T> by delegate {
```
colinrtwhite, line 65: Was hoping to avoid copying and pasting these classes, but the implementation classes can't implement an interface generated by the schema. Could potentially make the implementations open and mix in the generated interface (i.e. `fun Column() = object : ComposeColumn(), Column<() -> Unit> {}`), but I'd rather not make the implementations `open`.

redwood-layout/README.md
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /** https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink */
+  fun LayoutModifier.shrink(value: Int): LayoutModifier {
+    return then(ShrinkLayoutModifier(value))
+  }
+
+  /** https://developer.mozilla.org/en-US/docs/Web/CSS/align-self */
+  fun LayoutModifier.horizontalAlignment(alignment: CrossAxisAlignment): LayoutModifier {
+    return then(HorizontalAlignmentLayoutModifier(alignment))
```
colinrtwhite, line 56: We need to have separate horizontal and vertical layout modifiers since `Box` has two `CrossAxisAlignment`s.

redwood-layout/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+kotlin {
+  apply from: "${rootDir}/addAllTargets.gradle"
+
+  sourceSets {
+    commonMain {
+      dependencies {
+        implementation projects.redwoodFlexboxEngine
+        api projects.redwoodRuntime
+        api projects.redwoodWidget
+        api libs.jetbrains.compose.runtime
```
colinrtwhite, line 17: Should this module be renamed to `redwood-compose-layout` since it imports the Compose runtime? It only needs the annotations, but those aren't in a separate artifact, unfortunately.

redwood-layout/README.md
```diff
@@ -0,0 +1,102 @@
+Redwood Layout
+=======
+
+This artifact includes `Row` and `Column` widget implementations for Android Views, Android Compose UI, and iOS UIKit.
+
+To being, add the follow entries to your `schema.kt`:
```
JakeWharton:
```suggestion
To begin, add the follow entries to your `schema.kt`:
```

redwood-layout/README.md
```diff
@@ -0,0 +1,102 @@
+Redwood Layout
+=======
```
JakeWharton:
```suggestion
# Redwood Layout
```
For whatever reason the underline-based header is like soft-deprecated or something and the prefix-based one is preferred.

redwood-layout/README.md
```diff
@@ -0,0 +1,102 @@
...
+To being, add the follow entries to your `schema.kt`:
+
+```kotlin
+@Widget(1)
+data class Row(
+  @Property(1) @Default("Padding.Zero") val padding: Padding,
+  @Property(2) @Default("Overflow.Clip") val overflow: Overflow,
+  @Property(3) @Default("MainAxisAlignment.Start") val horizontalAlignment: MainAxisAlignment,
+  @Property(4) @Default("CrossAxisAlignment.Start") val verticalAlignment: CrossAxisAlignment,
+  @Children(5) val children: RowScope.() -> Unit,
```
JakeWharton:
```suggestion
  @Children(1) val children: RowScope.() -> Unit,
```
Children are namespaced (numberspaced?) differently than properties

redwood-layout/README.md
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+
+@Widget(2)
+data class Column(
+  @Property(1) @Default("Padding.Zero") val padding: Padding,
+  @Property(2) @Default("Overflow.Clip") val overflow: Overflow,
+  @Property(3) @Default("CrossAxisAlignment.Start") val horizontalAlignment: CrossAxisAlignment,
+  @Property(4) @Default("MainAxisAlignment.Start") val verticalAlignment: MainAxisAlignment,
+  @Children(5) val children: ColumnScope.() -> Unit,
```
JakeWharton:
```suggestion
  @Children(1) val children: ColumnScope.() -> Unit,
```

redwood-layout/README.md
```diff
@@ -0,0 +1,102 @@
...
+  @Property(3) @Default("MainAxisAlignment.Start") val horizontalAlignment: MainAxisAlignment,
+  @Property(4) @Default("CrossAxisAlignment.Start") val verticalAlignment: CrossAxisAlignment,
+  @Children(5) val children: RowScope.() -> Unit,
+)
+
+object RowScope {
+  /** https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow */
+  fun LayoutModifier.grow(value: Int): LayoutModifier {
+    return then(GrowLayoutModifier(value))
+  }
```
JakeWharton, line 21: In the schema layout modifiers are defined like widgets and scopes are empty types.

So this should look something like
```kotlin
object RowScope

@LayoutModifier(1, RowScope::class)
data class Grow(
  val value: Int,
)
```

redwood-layout/build.gradle
JakeWharton: This seems fine. Minimal dependencies FTW.

redwood-layout/src/commonMain/kotlin/app/cash/redwood/layout/properties.kt
```diff
@@ -0,0 +1,125 @@
...
+      "Invalid Padding: [$start, $end, $top, $bottom]"
+    }
+  }
+
+  public companion object {
+    public val Zero: Padding = Padding()
+  }
+}
+
+@Stable
```
JakeWharton, line 116: Does this have an effect on functions? I thought it was only for types.

JakeWharton (approved):

redwood-layout/src/commonMain/kotlin/app/cash/redwood/layout/properties.kt
colinrtwhite: Yep! When applied to functions is means same inputs == same output. It's added in Compose UI's `PaddingValues` class as well.

redwood-layout/README.md
colinrtwhite: @JakeWharton I think we might need to define these manually since they can't be generated or the implementation classes can't reference those modifiers. Will it cause issues if those don't have an associated tag?

JakeWharton: I suspect we just need to solve the case of including a sub-schema along with its set of generated types as a dependency. Almost like a `dependencies=` thing similar to Dagger where we can assume there's already be codegen.

```kotlin
@Schema(
  dependencies = [
    Flexbox::class,
  ],
  components = [ .. ]
)
```

Eventually we can get clever and wire in all the dependencies automatically through Gradle module metadata, but for now it just means you'll have to ship a couple dependencies here (schema artifact, layout modifier artifact, composable artifact, and widget artifact).

It also means the generated widget factory would have a property for each dependency
```kotlin
interface ExampleWidgetFactory {
  val flexbox: FlexboxWidgetFactory

  fun YourWidget(): YourWidget
}
```
so that you'd do something like
```kotlin
class AndroidViewExampleWidgetFactory(
  private val context: Context,
  override val flexbox: FlexboxWidgetFactory,
)
// ...
val factory = AndroidViewExampleWidgetFactory(
  context,
  AndroidViewFlexboxWidgetFactory(context), // <-- we ship this type!
)
```

I can work on minimally bootstrapping this dependency mechanism if this all sounds good.

colinrtwhite: @JakeWharton I like that idea - let's do it! I'm guessing `Flexbox` in this example is a reference to the child `@Schema`-annotated interface? This would also let me get rid of the `Bridge` classes in the `README`.

If it's OK I'll also leave the layout modifiers as is for the moment so I can reference them in the implementation PR. I'll update it to the correct way once it's possible to add dependencies.

JakeWharton: Sounds good. I'm out tomorrow so this'll be a next week thing.

Issue #394, opened 2022-09-27
`TreehouseComposeView` should not wrap content in a `Column`


JakeWharton:
Removing should give `Box`-like behavior of stacking children on the Z axis.

JakeWharton:
Didn't auto close because it was merged into a non-trunk branch erroneously. But is done!

PR #405, opened 2022-09-30
Require only a `Widget.Children` to create a `ProtocolDisplay`

JakeWharton:
This simplifies setup because generally you have an existing container into which you want to place content and can wrap it in one of our children helpers. Internally, we synthesize a root widget in the same way the Treehouse code used to do.

Since we no longer need a widget, we no longer need to expose the public 'wrap' functions from the diff-consuming widget factory. These are now private and changed slightly to keep the 'when'-containing function body as small as possible.


redwood-treehouse-composeui/src/main/kotlin/app/cash/redwood/treehouse/composeui/TreehouseComposeView.kt
```diff
@@ -23,26 +23,30 @@ import androidx.compose.foundation.layout.Column
...
   context: Context,
   private val treehouseApp: TreehouseApp<T>,
   public val widgetFactory: Widget.Factory<@Composable () -> Unit>,
 ) : FrameLayout(context), TreehouseView<T> {
-  public val composeView: ComposeView = ComposeView(context)
+  private val _children = ComposeWidgetChildren()
+  override val children: Widget.Children<*> get() = _children
 
   init {
+    val composeView = ComposeView(context)
```
JakeWharton, line 41: This class should extend `AbstractComposeView` rather than using a pointless `FrameLayout` to hold this `ComposeView`. Future change.

redwood-treehouse-composeui/src/main/kotlin/app/cash/redwood/treehouse/composeui/TreehouseComposeView.kt
```diff
@@ -23,26 +23,30 @@ import androidx.compose.foundation.layout.Column
...
+    composeView.setContent {
+      Column(
+        horizontalAlignment = Alignment.CenterHorizontally,
+      ) {
```
JakeWharton, line 45: This is #394

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolDisplay.kt
```diff
@@ -15,25 +15,23 @@
...
 import app.cash.redwood.protocol.Diff
 import app.cash.redwood.protocol.DiffSink
 import app.cash.redwood.protocol.EventSink
+import app.cash.redwood.protocol.PropertyDiff
+import app.cash.redwood.widget.Widget
+import kotlinx.serialization.json.JsonArray
 
 public class ProtocolDisplay<T : Any>(
-  private val root: DiffConsumingWidget<T>,
+  container: Widget.Children<T>,
```
swankjesse, line 30: Ooooh, nice

swankjesse (approved):

PR #406, opened 2022-09-30
Remove assorted caches in FlexboxEngine.

colinrtwhite:
Removes `indexToReorderedIndex`, `childrenFrozen`, `indexToFlexLine`, `measureSpecCache`, and `measuredSizeCache` member properties from `FlexboxEngine` so now it only keeps the `flexLines` state after measurement.

These member properties are mostly unused for the measurement/layout algorithms and were only important when `FlexboxEngine` was still coupled to `View`s.


redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/flexbox/FlexboxEngine.kt
```diff
@@ -67,59 +67,27 @@ public class FlexboxEngine {
...
-   * Cache the measured spec. The first 32 bit represents the height measure spec, the last
-   * 32 bit represents the width measure spec of each flex item.
-   * E.g. an entry is created like `(long) heightMeasureSpec << 32 | widthMeasureSpec`
-   *
-   * To retrieve a widthMeasureSpec, call [unpackLower] or [unpackHigher] for a heightMeasureSpec.
-   */
-  private var measureSpecCache: LongArray? = null
+  private var reorderedNodes: List<FlexNode> = listOf()
+    get() {
+      if (field.size != nodes.size) {
```
JakeWharton, line 78: What if I remove and add an item between calls to this function?

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/flexbox/FlexboxEngine.kt
```diff
@@ -1149,31 +1019,21 @@ public class FlexboxEngine {
...
     if (alignItems == AlignItems.Stretch) {
-      var flexLineIndex = 0
-      if (indexToFlexLine != null) {
-        flexLineIndex = indexToFlexLine!![fromIndex]
-      }
-      var i = flexLineIndex
+      var i = 0
       val size = flexLines.size
       while (i < size) {
         val flexLine = flexLines[i]
```
JakeWharton: Nothing manipulates `i` so I think this can be a normal for-in loop.

JakeWharton (approved):

redwood-flexbox-engine/src/commonMain/kotlin/app/cash/redwood/flexbox/FlexboxEngine.kt
colinrtwhite: The field is reset to an empty list every time `nodes` changes: https://github.com/cashapp/redwood/pull/406/files#diff-df8627befd69b95c6ee8f3b8240fd22ea54df1060b9fbf8c167bc5a4360c4ff1L70

swankjesse (approved):

PR #407, opened 2022-09-30
Fix snapshot publishing.

colinrtwhite:
Fixes an issue with the `publish` task introduced by this PR: https://github.com/cashapp/redwood/pull/356

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,6 +20,9 @@ android.disableAutomaticComponentCreation=true
 # Signals to our own plugin that we are building within the repo.
 app.cash.redwood.internal=true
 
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -20,6 +20,9 @@ android.disableAutomaticComponentCreation=true
 # Signals to our own plugin that we are building within the repo.
 app.cash.redwood.internal=true
 
+SONATYPE_HOST=DEFAULT
+RELEASE_SIGNING_ENABLED=true
```
colinrtwhite, line 24: This only attempts to sign the artifacts when we start publishing to maven central.

JakeWharton (approved):

PR #408, opened 2022-09-30
Rename FlexboxEngine -> FlexContainer, FlexNode -> FlexItem.

colinrtwhite:
This brings us in line with the [CSS flexbox spec](https://www.w3.org/TR/css-flexbox-1/).


JakeWharton (approved):
I remain opposed to this naming, but I won't stand in the way of it merging.

The engine noun makes sense to me. The spec uses container as the noun. I think layout also makes sense.

The spec uses flex as the adjective.

I would strongly prefer to adhere to the spec since we are already doing so for the rest of the API. Moreover, we are (probably) the only consumers of this API making the need to be cute or short minimal. Everyone else will only see the Row/Column names and the value API that adheres to the spec.

colinrtwhite:
Read back through that Slack thread and I agree that we should follow [the spec](https://www.w3.org/TR/css-flexbox-1/). My initial problem with `FlexLayout` was we could end up with a `layout.layout()` API, which is confusing. What do you think about the following renames:

`FlexboxEngine` -> `FlexContainer`
`FlexNode` -> `FlexItem`

I might be missing something in the spec, but it looks like [`FlexContainer`](https://www.w3schools.com/css/css3_flexbox_container.asp) is used pretty interchangeably with `FlexLayout` and it avoids the `layout.layout` problem. Those two renames would bring us back in line with the spec.

JakeWharton:
I am definitely comfortable with those!

swankjesse:
Yep, that works for me.

PR #409, opened 2022-10-01
Fix maven username and password environment variables.

colinrtwhite:
The properties were renamed in a previous publish plugin release.


JakeWharton (approved):
This is why I never update this plugin...

colinrtwhite:
Yeah, a few too many breaking changes recently.

PR #410, opened 2022-10-02
Removing an old article about React Native

moetouban:
I don't think the reasons for not using React Native in a brownfield app are still applicable, the framework and tooling have gone far since then. Even the author of that article has admitted multiple times that he didn't mean for the article to land in this way.


JakeWharton:
Our opinions on React Native are unchanged. However, I don't think it's important to mention other tools.

The entire README is nearly 2 years old and likely will be rewritten from scratch when we're ready to publicize the project more anyway.

JakeWharton (approved):

PR #419, opened 2022-10-05
Migrate layout module to have a schema

JakeWharton:
This is the mechanism by which it will be consumed downstream, but it's also useful for testing the forthcoming implementation of schema dependencies.


redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
```diff
@@ -15,17 +15,16 @@
  */
 @file:Suppress("MemberVisibilityCanBePrivate")
 
-package app.cash.redwood.layout
+package app.cash.redwood.layout.api
```
JakeWharton, line 18: Open to suggestions for a better name for this module/package. It's types shared by the schema, generated code, and widget implementations.

redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
```diff
@@ -15,17 +15,16 @@
...
 import androidx.compose.runtime.Stable
-import kotlin.jvm.JvmInline
 import kotlinx.serialization.Serializable
 
 /** Equivalent to `justify-content`. */
-@JvmInline
+@Immutable
 @Serializable
-public value class MainAxisAlignment private constructor(public val ordinal: Int) {
+public /*value*/ class MainAxisAlignment private constructor(private val ordinal: Int) {
```
JakeWharton: Non-JVM Compose can't handle these as value classes. Upstream issue forthcoming.

settings.gradle
```diff
@@ -4,7 +4,12 @@ include ':redwood-cli'
 include ':redwood-compose'
 include ':redwood-flexbox-engine'
 include ':redwood-gradle-plugin'
-include ':redwood-layout'
+include ':redwood-layout-api'
+include ':redwood-layout-schema'
+include ':redwood-layout-schema:compose'
+include ':redwood-layout-schema:layout-modifiers'
+include ':redwood-layout-schema:widget'
+include ':redwood-layout-view'
```
JakeWharton, line 12: Renamed this module because we'll need a `-composeui` version. I'm not sure it makes sense to ship a MPP artifact with Android views + iOS UIKit then have a separate Compose UI artifact, compared to just have three separate artifacts. Of course, the design of the existing widget and treehouse artifacts does that so maybe I'm wrong.

redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
colinrtwhite: Can we work-around it by making the constructor `internal` or does it fail for all value classes?

settings.gradle
colinrtwhite: I think I'd prefer separate `view`, `uikit`, and `composeui` artifacts as well since there will probably be consumers who are 100% compose UI and don't need the separate view-based implementations.

colinrtwhite (approved):

redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
colinrtwhite: `layout-api` sounds good to me :+1:

redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
JakeWharton: `internal` works!

PR #420, opened 2022-10-06
Convert ViewFlexContainerTest to a parameterized test.

colinrtwhite:
This lets us test all the combination of parameters at once and should make adding tests for new parameters easy.


redwood-layout/src/androidTest/kotlin/app/cash/redwood/layout/ViewFlexContainerTest.kt
```diff
@@ -15,129 +15,115 @@
...
-      horizontalAlignment(MainAxisAlignment.End)
-      verticalAlignment(CrossAxisAlignment.Center)
-      value.background = ColorDrawable(Color.LTGRAY)
-    }
-
-    textViews.forEachIndexed { index, textView ->
-      row.children.insert(index, textView)
-    }
-    paparazzi.snapshot(row.value)
+private inline fun <reified T> cartesianProduct(vararg lists: List<T>): List<Array<T>> {
```
colinrtwhite: Copied from [here](https://github.com/cashapp/paparazzi/issues/180#issuecomment-729871227), thanks!

redwood-layout/src/androidTest/kotlin/app/cash/redwood/layout/ViewFlexContainerTest.kt
```diff
@@ -15,129 +15,115 @@
...
 import app.cash.paparazzi.Paparazzi
+import app.cash.redwood.flexcontainer.AlignItems
+import app.cash.redwood.flexcontainer.FlexDirection
+import app.cash.redwood.flexcontainer.JustifyContent
 import org.junit.Rule
 import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+
+@RunWith(Parameterized::class)
```
JakeWharton: I tend to use https://github.com/google/TestParameterInjector nowadays, but since the types and values here aren't actually enums it may not actually save any code like it normally should.

JakeWharton (approved):

redwood-layout/src/androidTest/kotlin/app/cash/redwood/layout/ViewFlexContainerTest.kt
colinrtwhite: Looks like it has the [same issue](https://github.com/google/TestParameterInjector/issues/22) with value classes as `Parameterized ` unfortunately. We'd still have to wrap all our params and do the cross product ourselves.

PR #422, opened 2022-10-06
Reduce FlexContainer.measure complexity. Remove calls to measure.

colinrtwhite:
Reduces the number of branching paths based on `FlexDirection.isHorizontal` and uses `Orientation` instead.

Also removes several calls to `measure`. I think the remaining calls (except the main `measure` call) can be replaced by `minIntrinsicWidth(height: Int)`, etc. calls in a separate PR.

### DIFF
--- a/redwood-flex-container/src/commonMain/kotlin/app/cash/redwood/flexcontainer/FlexContainer.kt
+++ b/redwood-flex-container/src/commonMain/kotlin/app/cash/redwood/flexcontainer/FlexContainer.kt
@@ -98,8 +98,6 @@ public class FlexContainer {
     mainMeasureSpec = widthMeasureSpec,
     crossMeasureSpec = heightMeasureSpec,
     needsCalcAmount = Int.MAX_VALUE,
-    fromIndex = 0,
-    toIndex = -1,
   )
 
   /**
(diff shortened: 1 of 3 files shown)


redwood-flex-container/src/commonMain/kotlin/app/cash/redwood/flexcontainer/FlexContainer.kt
```diff
@@ -1625,8 +1502,8 @@ public class FlexContainer {
    * with the result.
    */
   private fun FlexItem.applyMeasure(widthSpec: MeasureSpec, heightSpec: MeasureSpec) {
-    val size = measurable.measure(widthSpec, heightSpec)
-    this.measuredWidth = size.width
-    this.measuredHeight = size.height
+    val (width, height) = measurable.measure(widthSpec, heightSpec)
+    this.measuredWidth = width.coerceIn(measurable.minWidth, measurable.maxWidth)
+    this.measuredHeight = height.coerceIn(measurable.minHeight, measurable.maxHeight)
```
colinrtwhite, line 1504: We can remove `measureWithConstraints` and enforce the min/max values here.

JakeWharton (approved):

PR #427, opened 2022-10-06
Fix scheduling 2 CI jobs for PRs.

colinrtwhite:
Currently, we're scheduling 2 identical `build` CI jobs for PRs that are from branches inside the Redwood repo as they match both the `pull_request` and `push` triggers.

Unfortunately it doesn't look like there's a straightforward way to fix this duplication so I suggest we do the same as OkHttp and only build pull requests and `trunk`. This means CI won't build branches unless there's an associated PR. Open to suggestions!

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,9 +4,7 @@ on:
   pull_request: {}
   push:
     branches:
-      - '**'
-    tags-ignore:
-      - '**'
+      - trunk
 
 env:
   GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx6g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"


JakeWharton:
So this comes up a lot but I think it's more of a theoretical problem than one in practice. Here's another recent example: https://github.com/cashapp/paparazzi/issues/572

I would prefer not changing it if only because I regularly push branches to test because I don't run many if any tests locally. I don't like the solution of draft PRs because it still sends notifications to watchers (which I try to minimize).

We have never had a problem with minutes, and the build config isn't duplicated or anything. Additionally, we have this config on nearly every repo so if we change it we should be changing it everywhere. I just don't think we're really causing any problems with it and it supports at least one useful workflow.

colinrtwhite:
Sounds good! If we're not worried about queuing then let's leave it as. The only thing I noticed it because both checks need to pass CI ends up taking the max of the two jobs to finish, which makes CI slightly slower.

JakeWharton:
Ah I guess because they have the same name and we can only guard it on the name, not the trigger type.

Best we can do is probably parallelize the build to make it faster. Removing the todo sample was an easy win. We can probably remove the counter sample too if we add a web target for emoji search.

There's probably room to do more such as splitting the various test targets.

PR #439, opened 2022-10-10
Implement schema dependencies

JakeWharton:
A schema may now depend on others and use their widgets and layout modifiers.

To produce a schema as a library you need to have the schema itself as well as the generated artifacts for layout modifiers and widgets. You do not need to generate compose code or any protocol code.

To consume a schema, add it as a dependency to your `@Schema` annotation along with a tag (1-2000). The layout modifier and widget artifacts for your schema will need to depend on the associated artifacts from the dependency. The code generation for the compose and protocol artifacts will automatically include all dependency types.


JakeWharton:
This is a functional first effort with some rough edges and limitations.

Two things greatly simplify the implementation:
- Dependencies can only be one level deep
- The Compose and protocol code is deferred to the final schema. For the Compose code it's asymmetric to the widgets which feels weird.

But hey it works and we can iterate from here.

colinrtwhite (approved):
💯

PR #448, opened 2022-10-12
Remove release build type from iOS apps

JakeWharton:
This keeps only the debug build type through various semi-hacks. Reduces the amount of tasks that run locally and on CI when you do an 'assemble'.

Before:

    $ time gca -q 2> /dev/null

    real	6m16.328s
    user	0m1.360s
    sys	0m0.263s

After:

    $ time gca -q 2> /dev/null

    real	2m59.198s
    user	0m1.239s
    sys	0m0.204s

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,6 @@
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+import org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType
+
 buildscript {
   dependencies {
     classpath libs.kotlin.gradlePlugin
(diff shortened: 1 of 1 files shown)


JakeWharton:
Obviously a non-scientific measurement, but the link tasks are last and they are serial so they suck so hard. There are tasks to parallelize them such as https://youtrack.jetbrains.com/issue/KT-49385/KotlinNative-Parallelize-link-tasks-for-different-targets and https://youtrack.jetbrains.com/issue/KT-49662/KotlinNative-Use-Gradle-Workers-API-for-Link-and-Compile-Tasks that will hopefully come soon (for the others).

colinrtwhite (approved):
Awesome!

PR #452, opened 2022-10-13
Implement LayoutModifier.toString().

### LINKED ISSUE #430: Layout modifier impl's need `toString` overrides

colinrtwhite:
Closes: https://github.com/cashapp/redwood/issues/430


JakeWharton (approved):

JakeWharton:
In the future we might want to migrate to using `+` instead of `StringBuilder` explicitly so that we can take advantage of string concat optimizations.

colinrtwhite:
Updated to use string interpolation: e.g. output is `"""CustomTypeWithDefault(customType=$customType, string=$string)"""`

PR #454, opened 2022-10-13
Expose scope as sealed interface

### LINKED ISSUE #451: Scoped LayoutModifier functions are accessible outside of their scope.
The test schema defines the following modifier:

```kotlin
object RowScope

@LayoutModifier(1, RowScope::class)
data class RowVerticalAlignment(
  val direction: Int,
)
```

However, we can access it from outside a `RowScope` (this compiles):

```kotlin
import example.redwood.compose.RowScope.rowVerticalAlignment

val modifier = LayoutModifier.rowVerticalAlignment(0)
```

To fix, we should probably generate `RowScope` as an `interface` and generate an internal `object RowScopeImpl`. This is similar to how Compose UI defines scopes.

`RowScope` could be defined as an `interface` in the schema as well, though this is optional.

JakeWharton:
To avoid using functions outside of the scope.

Closes #451 

This was the minimal change to fix the problem. I don't think there's much/any value to actually splitting the function across these types unless I've missed something.

### DIFF
--- a/redwood-cli/src/test/kotlin/app/cash/redwood/generator/ComposeGenerationTest.kt
+++ b/redwood-cli/src/test/kotlin/app/cash/redwood/generator/ComposeGenerationTest.kt
@@ -56,10 +56,7 @@ class ComposeGenerationTest {
     val fileSpec = generateComposable(schema, schema.widgets.single())
     assertThat(fileSpec.toString()).apply {
       contains("scoped: @Composable @ScopedAndUnscopedSchemaComposable RowScope.() -> Unit")
-      contains("RowScope.scoped()")
-
       contains("unscoped: @Composable @ScopedAndUnscopedSchemaComposable () -> Unit")
-      contains("unscoped()")
     }
   }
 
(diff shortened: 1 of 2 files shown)


colinrtwhite (approved):
LGTM - yeah probably not much reason to split out the functions.

redwood-cli/src/test/kotlin/app/cash/redwood/generator/ComposeGenerationTest.kt
```diff
@@ -56,10 +56,7 @@ class ComposeGenerationTest {
     val fileSpec = generateComposable(schema, schema.widgets.single())
     assertThat(fileSpec.toString()).apply {
       contains("scoped: @Composable @ScopedAndUnscopedSchemaComposable RowScope.() -> Unit")
-      contains("RowScope.scoped()")
```
JakeWharton, line 59: Removed these because they're testing implementation details. You can't invoke this lambda without a valid receiver so there's no point testing that we emitted code which does so.

PR #457, opened 2022-10-14
Fix Paparazzi tests. Add more tests for larger lists of items.

colinrtwhite:
Most of the lines changes are due to adding/removing screenshots.


.github/workflows/build.yaml
```diff
@@ -17,14 +17,16 @@ jobs:
...
+          lfs: true
       - uses: gradle/wrapper-validation-action@v1
       - uses: actions/setup-java@v3.5.1
         with:
           distribution: 'zulu'
           java-version: 18
       - uses: gradle/gradle-build-action@v2
 
-      - run: ./gradlew build dokkaHtmlMultiModule --parallel
+      - run: ./gradlew build dokkaHtmlMultiModule verifyPaparazziDebug --parallel
```
JakeWharton, line 29: Oof the non-validating default behavior strikes again.

redwood-layout-composeui/src/test/kotlin/app/cash/redwood/layout/composeui/ComposeFlexContainerTest.kt
```diff
@@ -65,27 +69,31 @@ class ComposeFlexContainerTest(
...
       val justifyContent: JustifyContent,
       val padding: Padding,
     ) {
       override fun toString() = "" +
         "FlexDirection.$flexDirection, " +
+        "Items(${items.size}), " +
         "AlignItems.$alignItems, " +
         "JustifyContent.$justifyContent, " +
-        "$padding"
+        "Padding(${padding.start})"
```
JakeWharton, line 96: We could change the `toString` on `Padding` to have this behavior if all four values match

JakeWharton (approved):

redwood-layout-composeui/src/test/kotlin/app/cash/redwood/layout/composeui/ComposeFlexContainerTest.kt
colinrtwhite: Should also do the same if horizontal padding and vertical padding match since there's a `Padding(horizontal, vertical)` overload? Personally I'd vote to keep the standard data class `toString`.

JakeWharton: Well I'm mostly just guarding against the case where some future traveler adds an asymmetric padding to this test (or copies this helper into another with asymmetric padding) and the test names accidentally collide. I believe paparazzi does nothing to guard against this case. But I also think the case of uniform padding is probably the most common usage.

colinrtwhite: Make sense! Uniform padding is the most likely. I'll update the `toString` in a subsequent PR.

.github/workflows/build.yaml
colinrtwhite: Yep, it caught me by surprise since Paparazzi still runs as part of `test`.

PR #458, opened 2022-10-14
Handle uniform values in Padding.toString.

colinrtwhite:


redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
```diff
@@ -102,6 +102,18 @@ public data class Padding(
     }
   }
 
+  override fun toString(): String {
+    return if (start == end && top == bottom) {
+      if (start == top) {
+        "Padding(all=$start)"
```
JakeWharton, line 108: I thought you'd do `Padding(0)` so all the existing tests would match. But you already changed them all so whichever you prefer.

JakeWharton (approved):

redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
colinrtwhite: Going to leave it as is to follow the standard data class format (`property=$value`), but we can revisit it later if needed.

PR #461, opened 2022-10-17
Support scrolling in ViewFlexContainer.

colinrtwhite:
Also removes the use of internal methods in tests.


redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
```diff
@@ -37,16 +39,31 @@ internal class ViewFlexContainer(context: Context, direction: FlexDirection) {
...
+
+  private val scrollView = if (direction.isHorizontal) {
+    HorizontalScrollView(context).apply {
+      isFillViewport = true
+      setTouchEnabled(false)
+      addView(hostView)
+    }
+  } else {
+    VerticalScrollView(context).apply {
+      isFillViewport = true
```
colinrtwhite, line 52: `HorizontalScrollView` and `VerticalScrollView`'s common parent is `FrameLayout`, which doesn't define `isFillViewport` so we need to set it on each view.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
```diff
@@ -37,16 +39,31 @@ internal class ViewFlexContainer(context: Context, direction: FlexDirection) {
     flexDirection = direction
   }
 
-  private val _view = HostView(context)
-  val view: View get() = _view
+  private val hostView = HostView(context)
+
+  private val scrollView = if (direction.isHorizontal) {
+    HorizontalScrollView(context).apply {
```
colinrtwhite, line 45: Maybe we can eventually extend `HorizontalScrollView`/`VerticalScrollView` to avoid creating an extra view, but it's not necessary at the moment.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
```diff
@@ -37,16 +39,31 @@ internal class ViewFlexContainer(context: Context, direction: FlexDirection) {
     flexDirection = direction
   }
 
-  private val _view = HostView(context)
-  val view: View get() = _view
+  private val hostView = HostView(context)
+
+  private val scrollView = if (direction.isHorizontal) {
```
JakeWharton, line 44: Is direction fixed? I thought it was a property that could change over time.

JakeWharton (approved):

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
colinrtwhite: Yep! It's only set on initialization of the column/row.

JakeWharton: Oh right I guess you would have to use a relocatable content lambda and then switch between `Row` and `Column` composables.

Issue #418, opened 2022-10-05
Make protocol values all value types


JakeWharton:
~There's no reason to use signed values.~ Just going to use value types instead. Unsigned values inside value types bad. See below..

JakeWharton:
I can't tell if this is worth it or not. It's slightly more code because Kotlin has semi-lame coercion rules, but I think it's probably worth it.

JakeWharton:
Actually seems like value classes might make the most sense. No one should be doing math on these. And one's value cannot be used for another (and we've had a bug or two on this in the past).

We can retain the use of ints in the schema definition and then have conversion functions to the value classes where ranges are validated.

JakeWharton:
I have managed to convince myself that this is all a bad idea.

We want two things: value types to differentiate between the semantic meaning of each value and unsigned values so that the range is entirely positive. The problem is that unsigned types box during _both_ serialization and deserialization.

Consider this type:
```kotlin
@JvmInline
@Serializable
value class PropertyTag(val value: UInt)
```

Let's look at the generated serializer:
```js
  $serializer.prototype.serialize_pzq1au_k$ = function (encoder, value) {
    var tmp0_inlineEncoder = encoder.encodeInline_8gn4q6_k$(this.descriptor_1);
    if (tmp0_inlineEncoder == null)
      null;
    else {
      tmp0_inlineEncoder.encodeSerializableValue_bps9ot_k$(UIntSerializer_getInstance(), new UInt(_PropertyTag___get_value__impl__y7w42g(value)));
    }
  };
```
Ignoring the encode bit, we first call `_PropertyTag___get_value__impl__y7w42g` with the value which "unwraps" the value. This is an identity function because the value is already the underlying type.
```js
  function _PropertyTag___get_value__impl__y7w42g($this) {
    return $this;
  }
```
But then we immediately call `new UInt` with that value which allocates an object only to pass it to `UIntSerializer` (which will unwrap and serialize the value).

Similarly, in the deserialization path:
```js
  $serializer.prototype.deserialize_2wnycf_k$ = function (decoder) {
    return _PropertyTag___init__impl__yg2cb0(decoder.decodeInline_k1q7ba_k$(this.descriptor_1).decodeSerializableValue_xpp80o_k$(UIntSerializer_getInstance()).data_1);
  };
```
[...]

PR #464, opened 2022-10-18
Use a value class to encapsulate a widget ID in the protocol

JakeWharton:
Refs #418

### DIFF
--- a/redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
+++ b/redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
@@ -15,6 +15,7 @@
  */
 package app.cash.redwood.protocol
 
+import kotlin.jvm.JvmInline
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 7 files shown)


JakeWharton:
The other ones are a bit harder, but also want to rename some shit in the protocol first as it's currently not super symmetric

redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
```diff
@@ -29,10 +30,19 @@ public fun interface DiffSink {
   public fun sendDiff(diff: Diff)
 }
 
+/** Identifies a widget instance. */
+@JvmInline
+@Serializable
+public value class Id(public val value: ULong) {
```
adrw, line 36: nice

adrw (approved):

colinrtwhite (approved):

redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
urica12: Ok

PR #465, opened 2022-10-19
Generate method references for composable property setters.

colinrtwhite:
Not sure there's a benefit to this considering the code isn't run on the JVM, but figured I would propose the change.


redwood-cli/src/main/kotlin/app/cash/redwood/generator/composeGeneration.kt
```diff
@@ -164,14 +164,14 @@ internal fun generateComposable(
           }
 
           val updateLambda = CodeBlock.builder()
-            .add("set(layoutModifier) { layoutModifiers = layoutModifier }\n")
+            .add("set(layoutModifier) { layoutModifiers = it }\n")
```
JakeWharton: Revert?

JakeWharton (approved):
It's still a good change because it reduces debug info where none is needed. Function references don't get line number info or source file information because the lambda body is generated by the compiler instead of in your source.

PR #467, opened 2022-10-19
Namespace known types into rough objects

JakeWharton:
More collisions are coming soon so doing this preemptively.


adrw (approved):
I like the look of this, makes it easier to distinguish generated/namespaced from regular camelCase vals.

colinrtwhite (approved):

PR #470, opened 2022-10-21
Generate method references for composable layout modifier setters.

colinrtwhite:
Following the same logic as: https://github.com/cashapp/redwood/pull/465

### DIFF
--- a/redwood-cli/src/main/kotlin/app/cash/redwood/generator/composeGeneration.kt
+++ b/redwood-cli/src/main/kotlin/app/cash/redwood/generator/composeGeneration.kt
@@ -87,6 +87,7 @@ fun SunspotButton(
   RedwoodComposeNode<SunspotWidgetFactory<*>, SunspotButton<*>>(
     factory = SunspotWidgetFactory<*>::SunspotButton,
     update = {
+      set(layoutModifier, SunspotButton<*>::layoutModifiers::set)
       set(text, SunspotButton<*>::text)
       set(enabled, SunspotButton<*>::enabled)
       set(onClick, SunspotButton<*>::onClick)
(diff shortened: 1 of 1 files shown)


JakeWharton:
Unfortunately this seems to be worse: https://kotlin.godbolt.org/z/PGaEEeWe3

The old version is `ExampleKt$updater$1` which generates a fairly normal class to implement the lambda which takes the values and does the call to the setter in the `invoke` function.

The new version is `ExampleKt$updater$2` which actually takes a reference to another lambda class, `ExampleKt$updater$3` which is more like the first lambda, and then wraps it.

Now it's possible these get optimized away. I'm working on https://github.com/cashapp/zipline/issues/738 which would allow dumping the resulting QuickJS bytecode to see what it looks like.

redwood-cli/src/main/kotlin/app/cash/redwood/generator/composeGeneration.kt
```diff
@@ -164,7 +165,7 @@ internal fun generateComposable(
           }
 
           val updateLambda = CodeBlock.builder()
-            .add("set(layoutModifier) { layoutModifiers = layoutModifier }\n")
```
JakeWharton, line 167: Huh I didn't realize your other change to switch this to `it` was intentional and not a typo. I was not aware that the value was also supplied to this lambda (although I guess it makes sense since the function references work).

colinrtwhite:
Ah yep, looks like it generates a `kotlin/reflect/KMutableProperty1` for the property as opposed to just getters/setters? Also cool website!

redwood-cli/src/main/kotlin/app/cash/redwood/generator/composeGeneration.kt
colinrtwhite: Yep, I could update it to use `it` again for consistency with the other properties. That would make it non-capturing, right?

JakeWharton: Yeah let's update to non-capturing. I'm actually surprised it worked at all in the capturing form. Maybe a pre-1.0 thing that got changed and I never tracked it. I was pretty sure that I copied their examples from Compose UI but maybe I just got it wrong.

Issue #371, opened 2022-09-15
Widget IDs


swankjesse:
I’d like to have a way to refer to widgets in a composition. My specific use case is so I can position a pop-up window adjacent to a widget in a composition. I can also imagine use cases where I use IDs to drive shared element transitions, possibly even transitions that span non-Redwood widgets!

### Compose

Every widget gains a new optional `id` parameter that defaults to null. 

Sample generated code:

```
@Composable
public fun Image(
  url: String, 
  onClick: (() -> Unit)? = null,
  layoutModifier: LayoutModifier = LayoutModifier,
  id: Long? = null,
): Unit { ... }
```

A user gets the ID for a widget by calling a new `nextWidgetId()` function and passing that in. Note that a `remember` call is necessary. 

```
class PopUpTreehouseUi {
  val popUpApplication: PopUpApplication = ...

  @Composable
  override fun Show() {
    val imageId = remember { nextWidgetId() }
    Image(
      url = ...
      id = imageId,
      onClick = {
        popUpApplication.showPopUpBesideWidget(imageId)
      }
    )
  }
```

If no ID is passed in, an ID is generated internally.

### Display

Users can retrieve their platform-specific widget object via the `ProtocolDisplay`.

```
public class ProtocolDisplay<T : Any>(...) : DiffSink {

  /**
   * Returns the widget on the current display for [id], or null if
   * no such widget exists. This may be the case if the widget has
   * since been removed from the display.
   */
  public fun widget(id: Long): T?
}
```

Perhaps the easiest way to make this value accessible to application code is to make another accessor on `TreehouseApp`. We may also want to make that member public on `TreehouseView`.

```
class PopUpApplication(
  val treehouseApp: TreehouseApp<*>,
) : ZiplineService {
  fun showPopUpBesideWidget(id: Long) {
    val imageView = (treehouseApp.widget(id) ?: return) as View
    showPopUp(widget.boundsInWindow())
  }
}
```

### Unique IDs

It’d be convenient to find a way to make widget IDs in the system unique, even if the `TreehouseApp` updates its code while it’s running. Probably the best way to bootstrap this is to make `ProtocolApplier` take its initial ID in its constructor, and to wire that in somehow.

JakeWharton:
Now that I've had a few hours to think about this, I do not think it's a good path forward.

[...]

PR #471, opened 2022-10-21
Refactor Widget.Children to support accessing child layout modifiers.

colinrtwhite:
Changes `Widget.Children` to support accessing + being notified about changes to child modifiers.

I left `LayoutModifier`s as part of `Widget` for the moment as it's not clear how we plan to extend layout modifiers in the future (e.g. [IDs](https://github.com/cashapp/redwood/issues/371)).

Not super satisfied with the design so looking for feedback!


redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/utils.kt
```diff
@@ -0,0 +1,74 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.protocol.widget
+
+internal fun <T> MutableList<T>.move(fromIndex: Int, toIndex: Int, count: Int) {
+  /*
+   * Copyright 2019 The Android Open Source Project
```
JakeWharton: You can move this to the file-level, delete these function-level ones, and add it to spotless exclude.

I keep meaning to expose these upstream.

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/utils.kt
```diff
@@ -0,0 +1,74 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.widget
+
+internal fun <T> MutableList<T>.move(fromIndex: Int, toIndex: Int, count: Int) {
+  /*
+   * Copyright 2019 The Android Open Source Project
```
JakeWharton: Same thing here.

Also two copies, woof. Time to try and expose upstream today, I think!

JakeWharton (commented):
Need to make another pass. Still thinking about it.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/utils.kt
JakeWharton: Sent http://r.android.com/2265406

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/Widget.kt
```diff
@@ -38,8 +41,11 @@ public interface Widget<T : Any> {
    * list. No additional validation needs to be performed (for example, checking index bounds).
    */
   public interface Children<T : Any> {
-    /** Insert child [widget] at [index]. */
-    public fun insert(index: Int, widget: T)
+    /** Insert child [widget] with [LayoutModifier] at [index]. */
+    public fun insert(index: Int, widget: T, layoutModifier: LayoutModifier)
+
+    /** Set the [LayoutModifier] for the widget at [index]. */
+    public fun set(index: Int, layoutModifier: LayoutModifier)
```
JakeWharton: `set` feels a little terse here since this is primarily a collection of widgets and this does not set a widget (in the same way a list's set function does). I could maybe swing with `update` or something wordy like `setLayoutModifier` / `updateLayoutModifier`.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolDisplay.kt
```diff
@@ -27,57 +27,72 @@ import app.cash.redwood.widget.Widget
...
         }
       }
     }
 
     for (layoutModifier in diff.layoutModifiers) {
-      val widget = checkNotNull(widgets[layoutModifier.id]) {
-        "Unknown widget ID ${layoutModifier.id}"
+      val node = node(layoutModifier.id)
+      val childIndex = node(node.parentId).childIds.indexOf(layoutModifier.id)
+      if (childIndex != -1) {
```
JakeWharton: I think this is a throw-worthy case. If the parent node is found but it's missing a reference to the child we want to crash as that's an invalid state.

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolApplier.kt
```diff
@@ -189,17 +189,11 @@ internal class ProtocolApplier(
     val current = current as DiffProducingChildrenWidget
     val children = current._children
 
-    // TODO We should not have to track this and send it as part of the protocol.
-    //  Ideally this would be entirely encapsulated on the display-side with additional bookkeeping.
-    //  For now, we track it here and send it in the protocol as a simple solution.
-    val removedIds = ArrayList<Id>(count)
     for (i in index until index + count) {
-      removedIds.add(children[i].id)
+      nodes.keys.remove(children[i].id)
```
JakeWharton: Don't think you need the `.keys` call here. Remove by key is on the map directly, too.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolDisplay.kt
JakeWharton: I think you can just do nothing and let line 74 throw AIOOBE

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolDisplay.kt
```diff
@@ -27,57 +27,72 @@ import app.cash.redwood.widget.Widget
...
+        node.widget.updateLayoutModifier(layoutModifier.elements)
+        node.parentChildren.set(childIndex, node.widget.layoutModifiers)
```
JakeWharton: Definitely feels like we only should be doing one of these. Did adding the `Widget` instance to the children not pan out?

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/HTMLElementChildren.kt
```diff
@@ -15,20 +15,33 @@
...
 import kotlinx.dom.clear
 import org.w3c.dom.HTMLElement
 import org.w3c.dom.get
 
 public class HTMLElementChildren(
   private val parent: HTMLElement,
 ) : Widget.Children<HTMLElement> {
-  override fun insert(index: Int, widget: HTMLElement) {
+  private val _children = MutableListChildren<HTMLElement>()
+  public val children: List<Child<HTMLElement>> get() = _children
```
JakeWharton: Where is the public half of this used? I didn't see anything. Seems to be on all of the built-in types but I can't see a usage.

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolApplier.kt
```diff
@@ -189,17 +189,11 @@ internal class ProtocolApplier(
     val current = current as DiffProducingChildrenWidget
     val children = current._children
 
-    // TODO We should not have to track this and send it as part of the protocol.
-    //  Ideally this would be entirely encapsulated on the display-side with additional bookkeeping.
-    //  For now, we track it here and send it in the protocol as a simple solution.
```
JakeWharton, line 194: Well this is nice

redwood-widget-compose/src/commonMain/kotlin/app/cash/redwood/widget/compose/ComposeWidgetChildren.kt
```diff
@@ -17,55 +17,39 @@ package app.cash.redwood.widget.compose
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.mutableStateListOf
+import app.cash.redwood.LayoutModifier
+import app.cash.redwood.widget.MutableListChildren
+import app.cash.redwood.widget.MutableListChildren.Child
```
JakeWharton: Not sure this is the right place for this type if it's being exposed by all these varying subtypes. Sure they all use MLC as implementation detail, but it probably belongs next to `Widget.Children` in `Widget`.

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/HTMLElementChildren.kt
JakeWharton: Oh I guess it's for the parent to see the instances and their layout modifiers.

JakeWharton (commented):
Good start. Do you have any big concerns or questions that make it a draft and not a normal PR? We can talk about it Monday in person, I guess.

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/Widget.kt
colinrtwhite: Agreed, was hoping to keep the one word method, but being more descriptive is probably best. Using just `update` might be confusing given `MutableListChildren`'s `onUpdate` callback.

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/HTMLElementChildren.kt
colinrtwhite: Yep, the plan is to expose this list so the parent can access the layout modifiers using the same index for the view.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolDisplay.kt
colinrtwhite: It does work, but I wanted to plan for a future where `Widget` doesn't have access to its own layout modifiers. I think for the moment I'll revert to using `Widget` and kick the "removing `Widget.layoutModifiers`" can down the road.

colinrtwhite:
@JakeWharton My main concern was around using `MutableListChildren` inside the other children implementations, but on second look this is probably ok.

Also I kept going back and forth on whether widgets should have access to their own layout modifiers. At the moment I'm leaning toward no, but didn't want to make that big change without discussing it. My thoughts:

- Only scoped layout modifiers currently make sense because they have to be explicitly handled by the parent.
- If we want to have unscoped layout modifiers they probably shouldn't be handled by the widget that they're applied on. Instead, we probably need some kind of `UnscopedLayoutModifierApplier<T>` interface that's provided to the `WidgetFactory` and handles applying a `LayoutModifier` on `T` at a global level. In this case `LayoutModifier`s still don't need to exist on `Widget`.

Either way I think I'll leave that change for a separate PR since my main goal is to unblock the per-item `padding`/`alignment` modifiers in `Column` and `Row` so Jesse can start using them in the Activity and Money tabs.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolDisplay.kt
colinrtwhite: Changed my mind and updated the PR to use `Widget<T>`.

redwood-widget-compose/src/commonMain/kotlin/app/cash/redwood/widget/compose/ComposeWidgetChildren.kt
```diff
@@ -17,55 +17,37 @@ package app.cash.redwood.widget.compose
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.mutableStateListOf
+import app.cash.redwood.widget.MutableListChildren
 import app.cash.redwood.widget.Widget
 
 public class ComposeWidgetChildren : Widget.Children<@Composable () -> Unit> {
-  private val children = mutableStateListOf<@Composable () -> Unit>()
+  private val _widgets = MutableListChildren<@Composable () -> Unit>(mutableStateListOf())
```
JakeWharton: I don't think you need the `MutableListChildren` wrapper here. Just using the `mutableStateListOf()` value should be enough. And I think that allows the Composition to partially re-render for children changes, whereas now it likely re-renders the whole list of children because it can't see through this wrapper type.

JakeWharton (approved):

redwood-widget-compose/src/commonMain/kotlin/app/cash/redwood/widget/compose/ComposeWidgetChildren.kt
```diff
@@ -17,22 +17,22 @@ package app.cash.redwood.widget.compose
...
 import androidx.compose.runtime.mutableStateListOf
-import app.cash.redwood.widget.MutableListChildren
+import app.cash.redwood.LayoutModifier
 import app.cash.redwood.widget.Widget
 
 public class ComposeWidgetChildren : Widget.Children<@Composable () -> Unit> {
-  private val _widgets = MutableListChildren<@Composable () -> Unit>(mutableStateListOf())
-  public val widgets: List<Widget<@Composable () -> Unit>> get() = _widgets
+  private val _widgets = mutableStateListOf<Pair<Widget<@Composable () -> Unit>, LayoutModifier>>()
+  public val widgets: List<Pair<Widget<@Composable () -> Unit>, LayoutModifier>> get() = _widgets
```
JakeWharton: Hmm why the `Pair`? Aren't the `LayoutModifier` instances available here the same as those you can get to from each `Widget` instance?

colinrtwhite: They are, but we need a way to trigger recomposition of an item if its layout modifiers change.

JakeWharton: Ah, yeah. Should that be a separate list with dummy reads in the composable? I guess the real version of my original question is: do we need to expose the pairs publicly to the parents?

colinrtwhite: Ah yep, we should probably avoid exposing the `Pair`s. I'll try using a separate list

PR #473, opened 2022-10-22
Add Theme, with dark mode support, to Emoji Sample App

### LINKED ISSUE #434: Emoji sample dark mode is broken
Text colors don't change leading to the illegible black on off-black combination.

dgmltn:
Closes #434

### DIFF
--- a/samples/emoji-search/android/src/main/java/app/cash/zipline/samples/emojisearch/theme.kt
+++ b/samples/emoji-search/android/src/main/java/app/cash/zipline/samples/emojisearch/theme.kt
@@ -15,18 +15,23 @@
  */
 package app.cash.zipline.samples.emojisearch
 
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.LocalContentColor
 import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Shapes
 import androidx.compose.material.Typography
 import androidx.compose.material.darkColors
+import androidx.compose.material.lightColors
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+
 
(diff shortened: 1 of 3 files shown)


samples/emoji-search/android/src/main/java/app/cash/zipline/samples/emojisearch/theme.kt
```diff
@@ -54,12 +65,22 @@ private val Shapes = Shapes(
...
-    colors = DarkColorPalette,
+    colors = colors,
     typography = Typography,
-    shapes = Shapes,
-    content = content
-  )
+    shapes = Shapes
+  ) {
+    CompositionLocalProvider(
+      LocalContentColor provides MaterialTheme.colors.onBackground,
```
JakeWharton: This is surprising. Where can I learn about why this is needed?

JakeWharton (approved):

samples/emoji-search/android/src/main/java/app/cash/zipline/samples/emojisearch/theme.kt
dgmltn: This surprised me too. In the [definition of Text()](https://androidx.tech/artifacts/compose.material/material/1.3.0-rc01-source/androidx/compose/material/Text.kt.html), the default value for `color` is `Color.Unspecified`, and:
```
 * @param color [Color] to apply to the text. If [Color.Unspecified], and [style] has no color set,
 * this will be [LocalContentColor].
```

But I've never noticed this problem before. In a Hello World compose app, even in dark mode, the default background for the activity is still white. But when the activity's theme in the manifest is based on `Theme.MaterialComponents.DayNight` (like it is in emoji-search), the activity background is automatically changed to darkgray/white without Compose.

dgmltn: The other solution is obviously to specify the color in the `Text()` composable. Do you have plans on how to bake this into redwood?
```
Text(
    text = "Hello world",
    color = MaterialTheme.colors.onBackground
)
```

JakeWharton: We can add that to this sample's Compose UI Text binding

dgmltn: Yep ok. As long as we always want `ComposeUiText` in the sample app to be the `onBackground` color that works too. Updated the PR.

PR #474, opened 2022-10-24
Hook up layout modifiers in View/ComposeUI flex containers.

colinrtwhite:
**Wait!** Github struggles to render the entire diff of this PR due to the number of image changes (~500). [Click here to view a diff of only the code changes](https://github.com/cashapp/redwood/pull/474/files?file-filters%5B%5D=.kt&show-viewed-files=true).

This PR hooks up layout modifiers in the View and ComposeUI flex containers so the `grow`/`shrink`/`padding`/`horizontalAlignment`/`verticalAlignment` layout modifiers affect the size and positioning of those children.


redwood-runtime/src/commonMain/kotlin/app/cash/redwood/LayoutModifier.kt
```diff
@@ -46,6 +46,11 @@ public interface LayoutModifier {
    */
   public fun <R> foldOut(initial: R, operation: (Element, R) -> R): R
 
+  /**
+   * Iterates over all [Element]s in this [LayoutModifier].
+   */
+  public fun forEach(block: (Element) -> Unit)
```
JakeWharton, line 52: The generated serialization code can also use this. Currently it's doing a `foldIn(Unit)` to emulate a for loop.

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/LayoutModifier.kt
```diff
@@ -111,6 +119,11 @@ public class CombinedLayoutModifier(
   override fun <R> foldOut(initial: R, operation: (LayoutModifier.Element, R) -> R): R =
     outer.foldOut(inner.foldOut(initial, operation), operation)
 
+  override fun forEach(block: (LayoutModifier.Element) -> Unit) {
+    outer.forEach(block)
+    inner.forEach(block)
+  }
```
JakeWharton, line 125: We should probably start adding tests for these. There's probably a bunch for existing functions that I never imported when I pulled in the type, too.

redwood-layout-schema/src/main/kotlin/app/cash/redwood/layout/modifiers.kt
```diff
@@ -22,13 +22,13 @@ import app.cash.redwood.schema.LayoutModifier
 /** Grow the item relative to [value] along the main axis. */
 @LayoutModifier(1, scopes = [RowScope::class, ColumnScope::class])
 public data class Grow(
-  val value: Int,
+  val value: Float,
```
JakeWharton, line 31: This and `Shrink` will need docs about the valid range of values. Should it throw outside of [0,1]?

JakeWharton (approved):

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/LayoutModifier.kt
JakeWharton: The only test I can find is https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/test/kotlin/androidx/compose/ui/ModifierTest.kt;l=22?q=ModifierTest&sq=&ss=androidx%2Fplatform%2Fframeworks%2Fsupport and it's for `foldIn` and `foldOut` which I suspect we'll be deleting once this merges and the generated serialization code is migrated to `forEach`. So, yeah. Feel free to just test your new changes.

redwood-layout-schema/src/main/kotlin/app/cash/redwood/layout/modifiers.kt
colinrtwhite: Yep, they both need to be positive, `[0, ∞)`. I'll document them for the moment, but eventually we should probably add runtime validation.

JakeWharton: Oh right. It's like LinearLayout, right? They're portions of their sum.

colinrtwhite: Yep! Exactly

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/LayoutModifier.kt
colinrtwhite: Sounds good. `forEach` isn't present in Compose UI, but if we're OK with diverging from Compose UI's `Modifier` then I'll remove `foldIn` and `foldOut` in a subsequent PR + add tests.

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/LayoutModifier.kt
colinrtwhite: Updated the generated code.

JakeWharton (approved):
:shipit:

PR #477, opened 2022-10-27
Add UIViewChildren.

colinrtwhite:
Adds a `Widget.Children` implementation for UIView in UIKit.

EDIT: Confirmed that this works in the emoji sample app.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,7 +26,7 @@ jobs:
           java-version: 18
       - uses: gradle/gradle-build-action@v2
 
-      - run: ./gradlew assemble dokkaHtmlMultiModule allTests verifyPaparazziDebug --parallel
+      - run: ./gradlew assemble lint allTests dokkaHtmlMultiModule verifyPaparazziDebug --parallel
 
       - name: Build Counter iOS sample
         run: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,7 +26,7 @@ jobs:
           java-version: 18
       - uses: gradle/gradle-build-action@v2
 
-      - run: ./gradlew assemble lint allTests dokkaHtmlMultiModule verifyPaparazziDebug --parallel
+      - run: ./gradlew assemble lint allTests verifyPaparazziDebug dokkaHtmlMultiModule --parallel
 
       - name: Build Counter iOS sample
         run: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,7 +26,7 @@ jobs:
           java-version: 18
       - uses: gradle/gradle-build-action@v2
 
-      - run: ./gradlew build dokkaHtmlMultiModule verifyPaparazziDebug --parallel
+      - run: ./gradlew assemble dokkaHtmlMultiModule allTests verifyPaparazziDebug --parallel
 
       - name: Build Counter iOS sample
         run: |
--- a/redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt
+++ b/redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Square, Inc.
+ * Copyright (C) 2022 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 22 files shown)


redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/HTMLElementChildren.kt
```diff
@@ -68,6 +68,8 @@ public class HTMLElementChildren(
   }
 
   override fun onLayoutModifierUpdated(index: Int) {
-    _widgets.onLayoutModifierUpdated(index)
+    val element = parent.children[index] as HTMLElement
+    parent.removeChild(element)
+    parent.insertBefore(element, parent.children[index])
```
colinrtwhite, line 73: Forcing the element to be redrawn. Not sure if there's a better way.

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt
```diff
@@ -0,0 +1,73 @@
...
+    subviews.forEachIndexed { offset, view ->
+      insert(view, newIndex + offset)
+    }
+  }
+
+  override fun remove(index: Int, count: Int) {
+    _widgets.remove(index, count)
+
+    for (i in 0 until count) {
+      // Subviews aren't removed immediately from the list.
```
JakeWharton: Is this documented? For safety I think I would prefer to iterate backwards.

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt
```diff
@@ -0,0 +1,73 @@
...
+    _widgets.insert(index, widget)
+    insert(widget.value, index)
+  }
+
+  override fun move(fromIndex: Int, toIndex: Int, count: Int) {
+    _widgets.move(fromIndex, toIndex, count)
+
+    val subviews = Array(count) { offset ->
+      parent.typedSubviews[fromIndex + offset].also(UIView::removeFromSuperview)
+    }
```
JakeWharton: In the old one I always pulled from `fromIndex` (https://github.com/cashapp/redwood/blob/0a69b7b5af28d98aab61f26ed4d8ee1f4fd9b6c8/redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt#L33-L37) but maybe we never exercised it properly? Although I swear there are tests...

I'm still very wary of this implicit behavior of things not actually being removed from the collection and would prefer to do a second iteration pass or iterate backwards.

Or, even better, just ensure there's tests! I suspect there already are, actually.

JakeWharton (approved):
Two things we should also do:
- Re-add `UIViewChildrenTest`: https://github.com/cashapp/redwood/blob/0a69b7b5af28d98aab61f26ed4d8ee1f4fd9b6c8/redwood-widget/src/iosTest/kotlin/app/cash/redwood/widget/UIViewChildrenTest.kt
- Remove the update lambda from `MutableListChildren` which was only added to support this case by reverting this patch: https://github.com/cashapp/redwood/commit/7f652ab013bb4ca19aedd55c227d95eced73f7d0#diff-46782cd85b7679ccea1be95ee8f7e1a58051637f23e1ae7542b4571c5bf5e843

JakeWharton:
We should also update emoji search sample which is written in Swift and confirm this works there. I just realized that I never did this with the `MutableListChildren` update lambda. Oops!

colinrtwhite:
@JakeWharton Just curious, but how come we removed the old `UIViewChildren`? Were there issues you ran into?

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt
colinrtwhite: Yeah, let me double check this behaviour. I thought it was the case, but I'll verify when migrating the emojisearch Swift impl to use `UIViewChildren`.

JakeWharton:
It wasn't useful in its current form. It didn't work with UIStackView or UITableView because it adds views in a different way. That's why we changed `MutableListChildren` to expose the lambda so they could add the views in the correct way for each subtype. I think your  new version will be sufficient, however.

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt
colinrtwhite: My bad - looks like children are immediately removed!

.github/workflows/build.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
           java-version: 18
       - uses: gradle/gradle-build-action@v2
 
-      - run: ./gradlew build dokkaHtmlMultiModule verifyPaparazziDebug --parallel
+      - run: ./gradlew assemble lint allTests verifyPaparazziDebug dokkaHtmlMultiModule --parallel
```
colinrtwhite, line 29: `build` doesn't run the `UIViewChildren` tests, but `allTests` does. Replaced `build` with separate `assemble` and `lint` since `build` also runs the tests which isn't necessary with `allTests`.

JakeWharton: We need to keep build or at least run check (but build is just assemble+check so might as well keep build). With this change you've removed spotless from the CI codepath, for example.

Seems like that's a severe bug if so and needs filed on Kotlin. check task needs to run all tests!

colinrtwhite: Gotcha, I'll revert back to `build` and submit an issue to the Kotlin issue tracker.

PR #478, opened 2022-10-27
Add websocket hot reloading

yissachar:
Adds websocket hot reloading for Android. Shaves 400-500ms off of hot reload times. For now leaving iOS as 500ms polling but will follow-up with another PR once that's working.

Note that right now there is no websocket on the server. This code will fallback to 500ms polling when we can't connect to a websocket, but also tries to make a websocket connection on each poll, in case the websocket is only down temporarily (e.g. in the case of a dev leaving emulator running, stopping hot reload server, then restarting, we want to make sure we always upgrade to websocket where possible to get the fastest speeds).


redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
```diff
@@ -0,0 +1,91 @@
...
+      .build()
+
+    var currentlyConnectedToWebSocket = false
+
+    val uri = URI(manifestUrl)
+    val request: Request = Request.Builder()
+      .url("ws://${uri.host}:${uri.port}/ws")
+      .build()
+
+    val wsFlow:Flow<String> = callbackFlow {
```
yissachar: All of this code feels super wonky. I'm not a flows expert but surely there's a better way to do this?

redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
```diff
@@ -0,0 +1,91 @@
...
+    val wsClient: OkHttpClient = OkHttpClient.Builder()
+      .readTimeout(0, TimeUnit.MILLISECONDS)
+      .connectTimeout(1, TimeUnit.SECONDS)
+      .build()
+
+    var currentlyConnectedToWebSocket = false
+
+    val uri = URI(manifestUrl)
+    val request: Request = Request.Builder()
+      .url("ws://${uri.host}:${uri.port}/ws")
```
yissachar: I don't love this approach of hardcoding the expected websocket URL, but I'm not sure where we would expose it as a config, and it doesn't seem like a huge problem since we're not likely to change the Zipline server often.

redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
```diff
@@ -0,0 +1,91 @@
...
+import kotlinx.coroutines.flow.transformLatest
+import kotlinx.coroutines.launch
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.Response
+import okhttp3.WebSocket
+import okhttp3.WebSocketListener
+
+@OptIn(ExperimentalCoroutinesApi::class)
+internal actual fun hotReloadFlow(flow: Flow<String>): Flow<String> {
```
yissachar, line 38: Would like to add some tests to validate this behaviour if I get the time.

redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
```diff
@@ -0,0 +1,91 @@
...
+          }
+          delay(500.milliseconds)
+        }
+      }
+
+      awaitClose {
+        ws?.close(1000, null)
+      }
+    }
+    wsFlow.onEach { emit(it) }.collect()
```
JakeWharton: You can delete this and switch to `flattMapLatest` from which you can return the inner flow and it will be automatically subscribed to.

redwood-treehouse/build.gradle
```diff
@@ -42,6 +42,9 @@ kotlin {
     nativeMain {
       dependsOn(hostMain)
     }
+    iosMain {
+      dependsOn(nativeMain)
+    }
```
JakeWharton: Was this needed? The `iosMain` source set and its dependence on `nativeMain` should already have been configured globally for all multiplatform modules.

redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
```diff
@@ -0,0 +1,91 @@
...
+internal actual fun hotReloadFlow(flow: Flow<String>): Flow<String> {
+  return flow.transformLatest { manifestUrl ->
+    val wsClient: OkHttpClient = OkHttpClient.Builder()
+      .readTimeout(0, TimeUnit.MILLISECONDS)
+      .connectTimeout(1, TimeUnit.SECONDS)
+      .build()
+
+    var currentlyConnectedToWebSocket = false
+
+    val uri = URI(manifestUrl)
```
JakeWharton: Jesee would probably want you to use OkHttp's `HttpUrl` here!

redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
```diff
@@ -0,0 +1,91 @@
...
+import okhttp3.WebSocket
+import okhttp3.WebSocketListener
+
+@OptIn(ExperimentalCoroutinesApi::class)
+internal actual fun hotReloadFlow(flow: Flow<String>): Flow<String> {
+  return flow.transformLatest { manifestUrl ->
+    val wsClient: OkHttpClient = OkHttpClient.Builder()
+      .readTimeout(0, TimeUnit.MILLISECONDS)
+      .connectTimeout(1, TimeUnit.SECONDS)
+      .build()
```
JakeWharton, line 43: A new client is being created for every upstream emission when what you probably want is to share this instance for the lifetime of the downstream subscription. You can do that with a level of nesting such as
```kotlin
return flow {
  val client = OkHttpClient.Builder()...build()

  emitAll(upstream.flatMapLatest { logic... })
}
```

redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
JakeWharton: Plus you don't actually have to use the `ws` scheme with OkHttp so you can do `HttpUrl.get(manifestUrl).resolve("/ws")` and be done.

JakeWharton (approved):

redwood-treehouse/build.gradle
yissachar: I had to add it to resolve some error, but doesn't look like that's the case anymore - possibly because I was actually working on a really old branch, and then rebased against latest.

yissachar:
Looks like MockWebServer doesn't have proper support for testing websockets (or at least I couldn't figure it out), so I'll just merge and try to talk to Jesse when he's back to see if there's a different way to test this.

redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
```diff
@@ -0,0 +1,96 @@
...
+          .url(manifestUrl.toHttpUrl().resolve("/ws")!!)
+          .build()
+
+        return@flatMapLatest callbackFlow {
+          var ws: WebSocket? = null
+          launch {
+            while (true) {
+              if (!currentlyConnectedToWebSocket) {
+                trySendBlocking(manifestUrl)
+                println("we are here")
```
swankjesse, line 59: this snuck in

swankjesse (commented):
NICE!

redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/hotReloadAndroid.kt
JakeWharton: It was removed in a follow-up

PR #483, opened 2022-10-30
Revert to using 'build' on CI.

colinrtwhite:
Follow up: https://github.com/cashapp/redwood/pull/477#discussion_r1008771892

`gradle build` and `gradle test` do not run `:redwood-widget:iosSimulatorArm64Test` whereas `gradle check` does. Investigating...

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,7 +26,7 @@ jobs:
           java-version: 18
       - uses: gradle/gradle-build-action@v2
 
-      - run: ./gradlew build allTests verifyPaparazziDebug dokkaHtmlMultiModule --parallel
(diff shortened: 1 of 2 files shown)


JakeWharton:
Are they being incorrectly cached maybe?

JakeWharton (approved):

colinrtwhite:
Yep, I'm testing locally and it seems like a caching issue. I can get `gradle build` to pass, but `gradle build --rerun-tasks` to fail.

JakeWharton:
We can sprinkle in some `upToDateWhen { false }` for all the test tasks for now, assuming they have some common type we can target.

colinrtwhite:
That "fixed" it!

PR #486, opened 2022-11-01
Add initial UIView Column/Row implementation.

colinrtwhite:


redwood-layout-uiview/README.md
```diff
@@ -0,0 +1,39 @@
+# Redwood Layout UIKit
+
+This artifact includes `Row` and `Column` widget implementations for UIKit.
+
+Currently, `UIViewRedwoodLayoutWidgetFactory` requires a `UIScrollView` factory due to Kotlin-Swift interop issues. Here's an implementation you can copy into your Swift project:
+
+```swift
+class UIScrollViewFactoryImpl: Redwood_layout_uiviewRedwoodUIScrollViewFactory {
```
colinrtwhite, line 8: Unfortunately, as far as I can tell at the moment it's impossible to create a `UIView` subclass from Kotlin as `intrinsicContentSize` (and other Swift properties) can't be overridden. This class delegates the actual `UIView` creation to Swift.

JakeWharton: You can do a fake override via an annotation if you do not have to invoke the super implementation. See the ProtocolDisplayRoot implementation in the Treehouse module.

I'm on mobile. Can link directly later.

colinrtwhite: @JakeWharton Unfortunately `@ObjCAction` can only be used for methods that return `Unit`/`Void`.

JakeWharton: Maybe we can use cklib to package it in the klib but write it in Objective-C.

@kpgalligan is it possible to use Objective-C with cklib? Or do you have any other ideas?

colinrtwhite: I also tried the `cinterop` solution described [here](https://stackoverflow.com/questions/69266205/subclassing-uiview-from-kotlin-native), but I ran into compilation issues unfortunately.

Specifically, it said that function signatures didn't match as I was using `app.cash.redwood.layout.uikit.CGSize` as an argument, but the C definition used the correct platform `CGSize`.

JakeWharton: Can you file an issue to track attempting to fix this? No reason to block the PR on it, but I really, really don't want to force this onto users long-term.

JakeWharton (approved):

redwood-layout-uiview/README.md
colinrtwhite: Yep, agreed - definitely want to avoid shipping this publicly. Looks like there's [this ticket](https://youtrack.jetbrains.com/issue/KT-40426/Incorrect-Objective-C-extensions-importing-that-prevents-UIKit-usage) already.

In the ticket folks seem to be having success with the cinterop solution so I'm going to take a second run at that, though I suspect I'll still run into issues with Swift properties like `intrinsicContentSize`.

JakeWharton: I meant on this repo. I'm sure we can figure something out or at least track an upstream ticket.

colinrtwhite: Tracked here: https://github.com/cashapp/redwood/issues/489

PR #487, opened 2022-11-02
Fix dependency layout modifiers not being included in diff producing/consuming serialization.

colinrtwhite:
Currently layout modifiers from schema dependencies throw an exception if they're applied to a widget that isn't part of that schema dependency. For example:

```kotlin
Column {
    Image("watch", layoutModifier = LayoutModifier.padding(5))
}
```

will throw because the `Padding` layout modifier is part of the `RedwoodLayout` schema whereas `Image` is from the main schema. This is because the generated code uses separate `toJsonElement` and `toLayoutModifier` functions for the main schema and each child dependency.

This PR updates the generated code to use one `toJsonElement` and one `toLayoutModifier` for all layout modifiers. Example generated output:

<details>
    <summary>diff producing generated code</summary>

  ```kotlin
  internal fun LayoutModifier.toJsonArray(json: Json): JsonArray = buildJsonArray {
    forEach { element -> add(element.toJsonElement(json)) }
  }

  private fun LayoutModifier.Element.toJsonElement(json: Json): JsonElement = when (this) {
    is RowVerticalAlignment -> RowVerticalAlignmentSurrogate.encode(json, this)
    is AccessibilityDescription -> AccessibilityDescriptionSurrogate.encode(json, this)
    is CustomType -> CustomTypeSurrogate.encode(json, this)
    is CustomTypeStateless -> CustomTypeStatelessSurrogate.encode()
    is CustomTypeWithDefault -> CustomTypeWithDefaultSurrogate.encode(json, this)
    is CustomTypeWithMultipleScopes -> CustomTypeWithMultipleScopesSurrogate.encode()
    is Grow -> GrowSurrogate.encode(json, this)
    is Shrink -> ShrinkSurrogate.encode(json, this)
    is Padding -> PaddingSurrogate.encode(json, this)
    is HorizontalAlignment -> HorizontalAlignmentSurrogate.encode(json, this)
    is VerticalAlignment -> VerticalAlignmentSurrogate.encode(json, this)
  }
  ```
</details>

<details>
    <summary>diff consuming generated code</summary>

  ```kotlin
  internal fun JsonArray.toLayoutModifier(json: Json, mismatchHandler: ProtocolMismatchHandler):
      LayoutModifier = fold<JsonElement, LayoutModifier>(LayoutModifier) { modifier, element ->
    modifier then element.toLayoutModifier(json, mismatchHandler)
  }
  
  private fun JsonElement.toLayoutModifier(json: Json, mismatchHandler: ProtocolMismatchHandler):
      LayoutModifier {
    val array = jsonArray
    require(array.size == 2) { "Layout modifier JSON array length != 2: ${array.size}" }
  
    val serializer = when (val tag = array[0].jsonPrimitive.int) {
      1 -> RowVerticalAlignmentImpl.serializer()
      2 -> AccessibilityDescriptionImpl.serializer()
      3 -> CustomTypeImpl.serializer()
      4 -> return CustomTypeStatelessImpl
      5 -> CustomTypeWithDefaultImpl.serializer()
      6 -> return CustomTypeWithMultipleScopesImpl
      1000001 -> GrowImpl.serializer()
      1000002 -> ShrinkImpl.serializer()
      1000003 -> PaddingImpl.serializer()
      1000004 -> HorizontalAlignmentImpl.serializer()
      1000005 -> VerticalAlignmentImpl.serializer()
      else -> {
        mismatchHandler.onUnknownLayoutModifier(tag)
        return LayoutModifier
      }
    }
  
    val value = array[1].jsonObject
    return json.decodeFromJsonElement(serializer, value)
  }
  ```
</details>

### DIFF
--- a/redwood-cli/src/main/kotlin/app/cash/redwood/generator/diffConsumingGeneration.kt
+++ b/redwood-cli/src/main/kotlin/app/cash/redwood/generator/diffConsumingGeneration.kt
@@ -475,8 +475,8 @@ private fun generateJsonElementToLayoutModifier(schema: Schema): FunSpec {
             .build(),
         )
       } else {
-        for ((layoutModifierSchema, layoutModifier) in layoutModifiers) {
-          val typeName = ClassName(layoutModifierSchema.widgetPackage(schema), layoutModifier.type.simpleName!! + "Impl")
+        for ((localSchema, layoutModifier) in layoutModifiers) {
+          val typeName = ClassName(localSchema.widgetPackage(schema), layoutModifier.type.simpleName!! + "Impl")
           if (layoutModifier.properties.isEmpty()) {
             addStatement("%L -> return %T", layoutModifier.tag, typeName)
           } else {
--- a/redwood-cli/src/main/kotlin/app/cash/redwood/generator/diffConsumingGeneration.kt
+++ b/redwood-cli/src/main/kotlin/app/cash/redwood/generator/diffConsumingGeneration.kt
@@ -25,7 +25,6 @@ import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.FileSpec
 import com.squareup.kotlinpoet.FunSpec
 import com.squareup.kotlinpoet.INT
-import com.squareup.kotlinpoet.KModifier.DATA
 import com.squareup.kotlinpoet.KModifier.INTERNAL
 import com.squareup.kotlinpoet.KModifier.OVERRIDE
 import com.squareup.kotlinpoet.KModifier.PRIVATE
@@ -371,16 +370,7 @@ internal fun generateDiffConsumingWidget(schema: Schema, widget: Widget, host: S
           FunSpec.builder("updateLayoutModifier")
             .addModifiers(OVERRIDE)
             .addParameter("value", KotlinxSerialization.JsonArray)
-            .addStatement(
-              """
-              |layoutModifiers = value.fold<%1T, %2T>(%2T) { modifier, element ->
-              |  modifier then element.%3M(json, mismatchHandler)
-              |}
-              """.trimMargin(),
-              KotlinxSerialization.JsonElement,
-              Redwood.LayoutModifier,
-              host.toLayoutModifier,
-            )
+            .addStatement("layoutModifiers = value.%M(json, mismatchHandler)", host.toLayoutModifier)
             .build(),
         )
         .build(),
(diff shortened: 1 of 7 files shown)


redwood-cli/src/main/kotlin/app/cash/redwood/generator/diffConsumingGeneration.kt
```diff
@@ -371,79 +370,28 @@ internal fun generateDiffConsumingWidget(schema: Schema, widget: Widget, host: S
...
 
       for (layoutModifier in schema.layoutModifiers) {
-        val typeName =
-          ClassName(schema.widgetPackage(host), layoutModifier.type.simpleName!! + "Impl")
+        val typeName = ClassName(schema.widgetPackage(host), layoutModifier.type.simpleName!! + "Impl")
         val typeBuilder = if (layoutModifier.properties.isEmpty()) {
           TypeSpec.objectBuilder(typeName)
         } else {
           TypeSpec.classBuilder(typeName)
-            .addModifiers(DATA)
```
colinrtwhite, line 446: I removed the `data` class modifier here since we generate our own `equals`/`hashCode`/`toString` for layout modifiers.

redwood-cli/src/main/kotlin/app/cash/redwood/generator/sharedHelpers.kt
```diff
@@ -111,6 +111,18 @@ internal fun Schema.layoutModifierImpl(layoutModifier: LayoutModifier): ClassNam
 internal val Schema.toLayoutModifier: MemberName get() =
   MemberName(widgetPackage(this), "toLayoutModifier")
 
+internal val Schema.toJsonArray: MemberName get() =
+  MemberName(composePackage(this), "toJsonArray")
+
+internal val Schema.toJsonElement: MemberName get() =
+  MemberName(composePackage(this), "toJsonElement")
+
+internal fun Schema.allLayoutModifiers(): List<Pair<Schema, LayoutModifier>> {
```
colinrtwhite, line 120: Was thinking about making this a map, but decided against it since iteration order is important.

redwood-cli/src/main/kotlin/app/cash/redwood/generator/diffConsumingGeneration.kt
JakeWharton: 👍 . All `data` classes will have to be removed eventually since they're not suitable for public API / API evolution.

redwood-cli/src/main/kotlin/app/cash/redwood/generator/sharedHelpers.kt
JakeWharton: Someday https://openjdk.org/jeps/431#Sequenced-maps

JakeWharton: If it's all internal code you can use the implementation type `LinkedHashMap` to guarantee iteration order is preserved / exposed.

JakeWharton (approved):

redwood-cli/src/main/kotlin/app/cash/redwood/generator/sharedHelpers.kt
colinrtwhite: Ooh yeah sequenced collections would be perfect.

colinrtwhite: Opted not to use a `Map` in the end since we need to know the total number of all layout modifiers.

PR #491, opened 2022-11-04
Add samples/counter/android-composeui

### LINKED ISSUE #490: Add counter sample variant for Android + compose
It might be nice to have a simple example with compose widgets instead of views, and to be able to compare the two.

(I've already done the work, but do you want it added to this repo?)

dgmltn:
Closes #490 

Adds a simple composeui-based sample module, seen here on the left. On the right is the existing views-based module:

<p>

</p>

My `initComposition` method seems like quite a bit of boilerplate, is that the minimum that's required? It seems like a good candidate for an extension function in the consumer library.


JakeWharton:
> My initComposition method seems like quite a bit of boilerplate, is that the minimum that's required? It seems like a good candidate for an extension function in the consumer library.

I don't think this meets the definition of boilerplate!

It mixes many things we don't know anything about: your chosen UI toolkit, your widget implementation, the fact that you're using the protocol to do seriazliation, your logging mechanism. And it cuts across the layers of the system which don't know about each other: the composition which produces the diffs and the display which ingests and renders the diffs.

Plus you only write this once for your entire application so it does not scale linearly with the size of the app, the number of widgets in your schema, or the number of composables behind the screens.

There are two paths to reduced verbosity: use the Treehouse system which moves the composition into a JS engine so that it can be updated without restarting the app, or remove the protocol serialization which allows the composition to talk directly to the widgets. 

I wouldn't worry about it for now. The other three apps for the counter are all still using in-process protocol serialization. We can switch them all to skipping that later.

JakeWharton (approved):

JakeWharton:
Great stuff! Thanks!

dgmltn:
Thanks!

PR #496, opened 2022-11-09
Treehouse EventListener

swankjesse:
This combines features of Zipline's EventListener and Redwood's ProtocolMismatchHandler.

It changes the Zipline APIs to use Redwood's types.

I expect distinct implementations on each target platform, and I expect end-users to adapt this during debugging.

### DIFF
--- a/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
+++ b/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package app.cash.redwood.treehouse
+
+import app.cash.zipline.Call
+import app.cash.zipline.CallResult
+import app.cash.zipline.ZiplineService
+
+public abstract class EventListener {
+  public open fun appCreated(
+    app: TreehouseApp<*>,
+  ) {
+  }
+
+  public open fun appCanceled(
+    app: TreehouseApp<*>,
+  ) {
+  }
+
+  public open fun codeLoadStart(
+    app: TreehouseApp<*>,
+    manifestUrl: String?,
+  ): Any? = Unit
+
+  public open fun codeLoadEnd(
+    app: TreehouseApp<*>,
+    manifestUrl: String?,
+    startValue: Any?,
+  ) {
+  }
+
+  public open fun codeLoadFailed(
+    app: TreehouseApp<*>,
+    manifestUrl: String?,
+    exception: Exception,
+    startValue: Any?,
+  ) {
+  }
+
+  public open fun onUnknownWidget(
+    app: TreehouseApp<*>,
+    kind: Int,
+  ) {
+  }
+
+  public open fun onUnknownLayoutModifier(
+    app: TreehouseApp<*>,
+    tag: Int,
+  ) {
+  }
+
+  public open fun onUnknownChildren(
+    app: TreehouseApp<*>,
+    kind: Int,
+    tag: Int,
+  ) {
+  }
+
+  public open fun onUnknownProperty(
+    app: TreehouseApp<*>,
+    kind: Int,
+    tag: Int,
+  ) {
+  }
+
+  public open fun downloadStart(
+    app: TreehouseApp<*>,
+    url: String,
+  ): Any? = Unit
+
+  public open fun downloadEnd(
+    app: TreehouseApp<*>,
+    url: String,
+    startValue: Any?,
+  ) {
+  }
+
+  public open fun downloadFailed(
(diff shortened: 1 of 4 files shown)


redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseLauncher.kt
```diff
@@ -37,6 +37,7 @@ import okio.Path
 public class TreehouseLauncher internal constructor(
   private val platform: TreehousePlatform,
   public val dispatchers: TreehouseDispatchers,
+  private val eventListener: EventListener,
```
swankjesse, line 40: For follow-up: actually invoke methods on these things.

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -0,0 +1,139 @@
...
+
+  public open fun appCanceled(
+    app: TreehouseApp<*>,
+  ) {
+  }
+
+  public open fun codeLoadStart(
+    app: TreehouseApp<*>,
+    manifestUrl: String?,
+  ): Any? = Unit
```
underscoretang, line 36: Can we replace this with a strong type / alias like a `typealias TreehouseCodeLoadValue = Any?` just so it's more obvious what this does. Not sure what to name it

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -0,0 +1,139 @@
...
+    name: String,
+    service: ZiplineService,
+  ) {
+  }
+
+  public open fun callStart(
+    call: Call,
+  ): Any? = Unit
+
+  public open fun callEnd(
```
underscoretang, line 122: should we have a `callFailed`? or is that going to be a part of the CallResult? Either way, will we have an array of the stacktrace here?

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
colinrtwhite: What would we expect to return from this method? Shouldn't it be `{}` like the others?

EDIT: Ah I see it's a token that's passed to subsequent lifecycle methods. I like it! Should this method return `null` by default?

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -0,0 +1,138 @@
...
+
+  public open fun takeService(
+    name: String,
+    service: ZiplineService,
+  ) {
+  }
+
+  public open fun callStart(
+    call: Call,
+  ): Any? = Unit
```
colinrtwhite, line 120: Same for this method.

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  public open fun appCanceled(
+    app: TreehouseApp<*>,
+  ) {
+  }
+
+  public open fun codeLoadStart(
+    app: TreehouseApp<*>,
+    manifestUrl: String?,
```
colinrtwhite, line 35: In what case is `manifestUrl` `null`?

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
colinrtwhite: Yep - `CallResult` provides a `kotlin.Result` so you can do `result.result.exceptionOrNull()` to get the exception if the call failed.

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseLauncher.kt
colinrtwhite: We should also add docs for the listener methods to the follow up 🙏

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -0,0 +1,138 @@
...
+  ): Any? = Unit
+
+  public open fun codeLoadEnd(
```
colinrtwhite, line 38: Is this called for both the success and fail cases? If it's only for success I might rename the methods to `codeLoadSuccess`/`codeLodeFail` to indicate they're mutually exclusive.

colinrtwhite (approved):

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
JakeWharton: Was going to make the same comment. `null` seems best.

JakeWharton (approved):

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
swankjesse: Good call y’all. Changed to null.

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
swankjesse: When a network load attempt fails, we fall back to either cached or embedded app. That doesn’t have a URL.

Issue #14, opened 2021-04-02
Polymorphic applier/scope to avoid protocol indirection


JakeWharton:
Right now if you're running a composition we go through our ProtocolApplier which creates protocol models (for optional serialization) which are then parsed by the display layer and applied to the display nodes. If you are targeting the same Kotlin backend with both the Compose layer and the display layer then this indirection does nothing but create garbage.

The working idea for fixing this is to make the Compose code generate depend on the Display generation. Then the protocol indirection can be created by generating implementations of the display nodes which create protocol models.

When setting up the TreehouseComposition, we would then have the ability to supply either the protocol-based display system or the real display system.

This also would be the mechanism by which #1 is accomplished. You supply an applier/scope/whatever that creates instances of your schema models for testing purposes.

PR #500, opened 2022-11-10
Remove knowledge of protocol from generated code

JakeWharton:
Instead of an intermediate children node which is aware of the protocol tag, use an intermediate node which can select the appropriate `Widget.Children` instance from its parent. Then, the diff-producing widgets now return `Widget.Children` instances which record operations to the diff.

This completely unblocks using Redwood without the protocol. All that's left is to decide whether we duplicate the applier and composition setup or try to share it.

Refs #14

### DIFF
--- a/redwood-cli/src/main/kotlin/app/cash/redwood/generator/composeGeneration.kt
+++ b/redwood-cli/src/main/kotlin/app/cash/redwood/generator/composeGeneration.kt
@@ -77,19 +77,23 @@ internal fun generateComposableTargetMarker(schema: Schema): FileSpec {
 /*
 @Composable
 @SunspotComposable
-fun SunspotButton(
-  text: String?,
-  enabled: Boolean = true,
-  onClick: (() -> Unit)? = null,
+fun Row(
+  padding: Padding = Padding.Zero,
+  overflow: Overflow = Overflow.Clip,
   layoutModifier: LayoutModifier = LayoutModifier,
+  children: @Composable @SunspotComposable RowScope.() -> Unit,
 ): Unit {
-  RedwoodComposeNode<SunspotWidgetFactory<*>, SunspotButton<*>>(
-    factory = SunspotWidgetFactory<*>::SunspotButton,
+  _RedwoodComposeNode<SunspotWidgetFactory<*>, Row<*>>(
+    factory = { it.RedwoodLayout.Row() },
     update = {
       set(layoutModifier) { layoutModifiers = it }
-      set(text, SunspotButton<*>::text)
-      set(enabled, SunspotButton<*>::enabled)
-      set(onClick, SunspotButton<*>::onClick)
+      set(padding, Row<*>::padding)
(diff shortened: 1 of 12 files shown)


redwood-cli/src/main/kotlin/app/cash/redwood/generator/composeGeneration.kt
```diff
@@ -77,19 +77,23 @@ internal fun generateComposableTargetMarker(schema: Schema): FileSpec {
...
-      set(text, SunspotButton<*>::text)
-      set(enabled, SunspotButton<*>::enabled)
-      set(onClick, SunspotButton<*>::onClick)
+      set(padding, Row<*>::padding)
+      set(overflow, Row<*>::overflow)
+    },
+    content = {
+      into(Row<*>::children) {
+        RowScopeImpl.children()
+      }
```
JakeWharton, line 96: This the real change. Before we did
```kotlin
_SyntheticChildrenNode(1U) {
  RowScopeImpl.children()
}
```

redwood-cli/src/main/kotlin/app/cash/redwood/generator/diffProducingGeneration.kt
```diff
@@ -227,10 +227,11 @@ internal fun generateDiffProducingWidget(schema: Schema, widget: Widget, host: S
               }
               is Children -> {
                 addProperty(
-                  PropertySpec.builder(trait.name, NOTHING, OVERRIDE)
+                  PropertySpec.builder(trait.name, RedwoodWidget.WidgetChildren.parameterizedBy(NOTHING))
+                    .addModifiers(OVERRIDE)
                     .getter(
                       FunSpec.getterBuilder()
-                        .addStatement("throw %T()", Stdlib.AssertionError)
+                        .addStatement("return diffProducingWidgetChildren(%LU)", trait.tag)
```
JakeWharton, line 234: Then here is where the diff-producing widgets return an instance of children that produce diffs, and this is where that tag value moved to.

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/utils.kt
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.protocol.compose
+
+internal fun <T> MutableList<T>.move(fromIndex: Int, toIndex: Int, count: Int) {
```
colinrtwhite, line 18: We're now at 4 copies of these functions! 😅

colinrtwhite (approved):
Awesome!

PR #501, opened 2022-11-10
Support execution without the protocol layer

### LINKED ISSUE #14: Polymorphic applier/scope to avoid protocol indirection

JakeWharton:
Update all counter clients to remove protocol usage.

https://user-images.githubusercontent.com/66577/201180223-40a4d081-4e85-4335-bfef-322c73dc8733.mp4

Closes #14


JakeWharton:
Lots of duplication here (from the protocol-compose module).

I think we probably want to expose an abstraction that controls the output of the composition (direct or protocol) and have that be a thing you feed into the factory function. This would replace the protocol-specific factory function and the type could serve as holding the protocol node map and sending the diffs when the applier end-changes callback is invoked.

Going to think about it for a bit. I would be fine landing this as-is and doing that in a follow-up so we'll see what I come up with.

samples/counter/ios/shared/src/commonMain/kotlin/example/ios/CounterViewControllerDelegate.kt
```diff
@@ -37,27 +33,16 @@ class CounterViewControllerDelegate(
...
     )
-    val factory = DiffConsumingSunspotWidgetFactory(IosSunspotNodeFactory)
-    val display = ProtocolDisplay(
+    val composition = RedwoodComposition(
+      scope = scope,
       container = children,
-      factory = factory,
-      eventSink = composition,
+      factory = IosSunspotNodeFactory,
     )
```
dgmltn, line 44:
```suggestion
    ).apply {
      setContent {
        Counter()
      }
    }
```

samples/counter/android-composeui/src/main/java/example/android/MainActivity.kt
```diff
@@ -37,41 +31,20 @@ class MainActivity : AppCompatActivity() {
...
 
-    val widgets = initComposition {
-      Counter()
-    }
+    val children = ComposeWidgetChildren()
+    RedwoodComposition(
+      scope = scope,
+      container = children,
+      factory = AndroidSunspotWidgetFactory(),
+    )
```
dgmltn, line 39:
```suggestion
    ).apply {
      setContent {
        Counter()
      }
    }
```

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
```diff
@@ -71,14 +156,67 @@ public inline fun <F : Widget.Factory<*>, W : Widget<*>> RedwoodComposeNode(
...
+  _RedwoodComposeContent.Instance.content()
 
   currentComposer.endNode()
 }
 
 /**
  * @suppress For generated code usage only.
  */
-public interface RedwoodApplier<F : Widget.Factory<*>> {
+@Suppress("ClassName") // Hiding from auto-complete.
```
swankjesse: Jam it in an `internal` package?

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
```diff
@@ -71,14 +156,67 @@ public inline fun <F : Widget.Factory<*>, W : Widget<*>> RedwoodComposeNode(
...
+ * Compose's tree assumes each node only has single list of children. Or, put another way, even if
+ * you apply multiple children Compose treats them as a single list of child nodes. In order to
+ * differentiate between these children lists we introduce synthetic nodes. Every real node which
+ * supports one or more groups of children will have one or more of these synthetic nodes as its
+ * direct descendants. The nodes which are produced by each group of children will then become the
+ * descendants of those synthetic nodes.
+ *
+ * @suppress For generated code usage only.
+ */
+// TODO Make this type private when the applier moves into this module.
```
swankjesse: 👍🏻

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/AbstractDiffProducingWidget.kt
```diff
@@ -0,0 +1,52 @@
...
+package app.cash.redwood.protocol.compose
+
+import app.cash.redwood.protocol.Event
+import app.cash.redwood.protocol.Id
+import app.cash.redwood.protocol.LayoutModifiers
+import app.cash.redwood.protocol.PropertyDiff
+import app.cash.redwood.widget.Widget.Children
+
+/**
+ * @suppress For generated code usage only.
```
swankjesse: What are you suppressing?

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/AbstractDiffProducingWidget.kt
```diff
@@ -0,0 +1,52 @@
...
+public abstract class AbstractDiffProducingWidget(
+  public val type: Int,
+) : DiffProducingWidget {
+  override val value: Nothing
+    get() = throw AssertionError()
+
+  public var id: Id = Id(ULong.MAX_VALUE)
+    internal set
+
+  @Suppress("PropertyName") // Avoiding potential collision with subtype properties.
```
swankjesse: Can’t be private eh?

swankjesse (approved):

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/AbstractDiffProducingWidget.kt
JakeWharton: Visibility in docs

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
JakeWharton: I don't think that helps since auto-complete doesn't know anything about that convention. I finished this refactoring and most types become private. The ones which stay public I'm going to put behind an `@OptIn` annotation instead.

JakeWharton:
I have mostly finished this refactoring which I'll push to this branch shortly. I'm going to land it all in separate PRs, however. The refactoring was backwards (start with what i want and work back so it fits) and I want to instead land the changes with forward (and reviewable) chunks.

PR #503, opened 2022-11-10
Remove the concept of clear

JakeWharton:
For Compose, this Applier function is only invoked when the Composition is going away and a new one is going to attach to an existing Applier. We never do that, thus, this function will never be invoked. Clear was used by Treehouse so a similar mechanism is introduced at that layer.


JakeWharton:
This will simplify the abstraction talked about in https://github.com/cashapp/redwood/pull/501#issuecomment-1310761350 as clearing was a weird case which behaved differently than any other function.

colinrtwhite (approved):

JakeWharton:
I'm going to add a boolean to the applier to ensure there are no interactions after clear to be safe. The docs and implementation current will not do this, but better to guard it with a crash than silently continue in a corrupted state.

PR #505, opened 2022-11-11
Fix the internal publication process.

colinrtwhite:
We need to disable release signing for now + set the credentials properly.


JakeWharton (approved):
What's the story for how we get signing back enabled then? I would prefer to always keep the repo ready for Maven central releases, but of course we can relax that for the near term.

colinrtwhite:
@JakeWharton I don't think we've been publishing signed redwood artifacts so far (release signing is disabled by default) so we'll need to set up the signing config.

JakeWharton:
Currently, yes, but the code in the repo should be optimizing for release to Sonatype for Maven central. Internal publishing is a secondary concern that should eventually trend to zero and I'm okay if you have to specify overrides like `-DRELEASE_SIGNING_ENABLED=false` as part of that.

colinrtwhite:
Ah gotcha - agreed signing should be the default. I'll add a `RELEASE_SIGNING_ENABLED` param and set us back to default true.

JakeWharton:
We should do a 0.1 soon if only to say there's non-snapshot bits out there

PR #506, opened 2022-11-11
Migrate to doubles for flex item coordinates.

colinrtwhite:
iOS uses doubles for its coordinate system so we need to represent item positions as doubles inside `FlexContainer`. It's not ideal since we use extra memory on Android, but we don't have much choice if we want to preserve precision.

[Link to the file changes without screenshot test updates.](https://github.com/cashapp/redwood/pull/506/files?file-filters%5B%5D=.kt&file-filters%5B%5D=.md&file-filters%5B%5D=.swift&show-viewed-files=true) The rounding changes alter the screenshots imperceptibly.


redwood-flex-container/src/commonMain/kotlin/app/cash/redwood/flexcontainer/properties.kt
```diff
@@ -185,29 +184,38 @@ public value class JustifyContent private constructor(private val ordinal: Int)
 /**
  * A MeasureSpec encapsulates the layout requirements passed from parent to child.
  * Each MeasureSpec represents a requirement for either the width or the height.
- * A MeasureSpec is comprised of a size and a mode.
+ * A MeasureSpec is composed of a size and a mode.
  */
-@JvmInline
-public value class MeasureSpec private constructor(private val value: Int) {
+public class MeasureSpec private constructor(
```
colinrtwhite, line 189: We can't occupy the top 2 bits with a `Double` unfortunately (at least easily?). Will circle back on seeing if this can be an inline class later.

JakeWharton: You'll have to interpret the bits as a Long, mask the bits, and then interpret back as a double. File a tracking issue?

JakeWharton (approved):

swankjesse (approved):

PR #509, opened 2022-11-11
Restore the TreehouseView parameter in ViewBinder.widgetFactory

swankjesse:
This is necessary for View-based factories which need a Context to create the child views.

### DIFF
--- a/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
+++ b/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
@@ -173,6 +173,7 @@ public class TreehouseApp<T : Any> internal constructor(
       val display = ProtocolDisplay(
         container = view.children as Widget.Children<Any>,
         factory = spec.viewBinder.widgetFactory(
+          view,
           zipline.json,
           eventPublisher.protocolMismatchHandler(this@TreehouseApp),
         ) as Factory<Any>,
--- a/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/ViewBinder.kt
+++ b/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/ViewBinder.kt
@@ -31,6 +31,7 @@ public interface ViewBinder {
 
   /** Returns a widget factory for encoding and decoding changes to the contents of [view]. */
   public fun widgetFactory(
+    view: TreehouseView<*>,
     json: Json,
     mismatchHandler: ProtocolMismatchHandler,
   ): DiffConsumingWidget.Factory<*>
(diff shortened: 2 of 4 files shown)


JakeWharton:
This seems like a really weird way to get a context other than just bringing one you capture yourself.

colinrtwhite:
+1 to what Jake said. `view` is only necessary for its context when using `View`-based widget factories. Compose UI and UiKit don't need this param. I wonder if there's a better place for `widgetFactory` to live?

Could you grab the `view` from the other `ViewBinder` methods and hold a reference to it?

swankjesse:
I don’t follow?

In our application, the `TreehouseApp` can be an host-application-scoped object. Ie. it can be a Dagger singleton. The `TreehouseView` a screen-scoped object. It has a narrower `Context` that may have scoped things in it, like a theme.

We need a way for the `TreehouseApp` to create child views for the `TreehouseView`. How should it get a Context for that? It used to work!

JakeWharton:
I think it's weird to have to unsafe cast this to a subtype to get access to one of its properties that's effectively a dependency of the factory. What about if you need an image loader or a feature flag manager or something else? Do we have a missing layer that sits between app and each view?

swankjesse:
This is specifically about `Context` in Android. It travels through the view hierarchy, and we can’t/shouldn’t get it from elsewhere.

swankjesse:
(Previously the `TreehouseView` had a `WidgetFactory` parameter, but you deleted that also. Bringing that back instead would also unblock.)

colinrtwhite:
@swankjesse This is what I did to get access to the `view`, though it's not ideal:

```kotlin
class ViewBinderImpl : ViewBinder {
  private lateinit var view: TreehouseView<*>

  override fun codeLoading(view: TreehouseView<*>) {
    this.view = view
  }
}
```

Hmm I don't have better solution so I'd be up for re-adding the parameter until we figure out a better way to pass view-scoped dependencies to `Widget.Factory` instances.

colinrtwhite (approved):

JakeWharton:
Yeah things flow through the other hierarchies too. With Compose UI you can't cast a subtype to access anything since it's all contained within the functions. So this is a band-aid for now and we're missing something to support these usages.

swankjesse:
I don’t expect we’ll need anything for Compose UI. It can pass themes etc. down the call stack while rendering.

It’s a quirk of Android’s API that you need a Context to construct a View.

What we had previously was even more powerful than this, cause we could configure the factory for each view. But this should be sufficient to unblock Android views.

JakeWharton:
My point is more that the motivation to include these things is usually a sign of broken layering.

Your refactor when we open sourced it was a first step to fixing things. My refactor helped clean that up further in order to pipe through view-level configuration but it wasn't everything. In that refactor I noted that this specific callback seemed misplaced as it's conflated with event-listener-style callbacks (perhaps now completely obviated by your actual event listener).

I don't think it makes sense for the `TreehouseApp` to own a `ViewBinder` and that responsibility should be on the `TreehouseView`. In doing so, that means the specific subtype of the `TreehouseVIew` is known without an unsafe cast across unrelated types. But even better, the `Context` used to create the `TreehouseView` is directly available (as well as any other dependencies you want to pull).

Perhaps it's easier for me to show you what I mean with a PR. I'll file an issue about it for follow up post-hackweek.

PR #516, opened 2022-11-14
Delete ViewBinder

swankjesse:
Promote the widgetFactory function to TreehouseApp.Spec.


JakeWharton:
I don't think it makes sense for this to live at the "app" layer. It's not tied to running the application. It's part of the rendering/view layer. I think the `TreehouseView` should own this callback and that eliminates the need to have `TreehouseView` as an argument and for doing unsafe downcasts.

swankjesse:
I gave it a shot here:
https://github.com/cashapp/redwood/pull/518

The thing that I don’t like about this is now every view that used to need a `TreehouseApp` now also needs a `TreehouseView.WidgetSystem`. The benefit of this are abstract (run Treehouse without a widget system?) but the cost is concrete (pass 2x as many parameters to create each view). Is there another benefit I’m not seeing?

JakeWharton:
It just seems like the correct location for it with respect to the layering. The current layering is like if SQLDelight owned a factory for an Adapter for every RecyclerView in your app. Would it work? Maybe, but it's a little weird.

Does your factory need a `Context`? Once moved now at that layer and have direct access to it rather than downcasting a parameter you otherwise shouldn't need. It also means you can inject anything else your widget factory needs at that layer such analytics dependencies or a view recycler.

That 2x is a multiplier on a value of 1 meaning the result is still only 2. It's not like 60 parameters turned into 120. I'm pretty comfortable with both Redwood and Treehouse having a decent amount of glue because of the complexity it offloads. It's not like an app will have 100 of these.

By having this be a property of the view layer it more easily facilitates migration from a view-based system to a Compose UI-based system. Each TreehouseView would owns the respective factory for the children which are placed inside of it. With the change you could even run two views side-by-side with different rendering systems for fun.

The biggest evidence that these things belong at the same layer is that the type parameters go from being `<*>` to `<T>`. We are creating the relationship between them at a place where they are able to agree on the parameter.

swankjesse:
> By having this be a property of the view layer it more easily facilitates migration from a view-based system to a Compose UI-based system. Each TreehouseView would owns the respective factory for the children which are placed inside of it. With the change you could even run two views side-by-side with different rendering systems for fun.

Yeah, this is compelling!

PR #521, opened 2022-11-17
Make padding consistent between platforms.

colinrtwhite:
[Link to diff without screenshot test changes.](https://github.com/cashapp/redwood/pull/521/files?file-filters%5B%5D=.kt&show-viewed-files=true)

Updates the implementations to ensure generic padding values look the same on both platforms. I standardized on iOS's scale since their coordinates are always density independent.


redwood-layout-composeui/src/main/kotlin/app/cash/redwood/layout/composeui/utils.kt
```diff
@@ -68,7 +69,20 @@ internal fun CrossAxisAlignment.toAlignSelf() = when (this) {
   else -> throw AssertionError()
 }
 
-internal fun Padding.toSpacing() = Spacing(start.toDouble(), end.toDouble(), top.toDouble(), bottom.toDouble())
+internal fun Padding.toSpacing(context: Context): Spacing {
```
colinrtwhite, line 72: I'd like to pull a lot of the methods from this file into a `redwood-layout-common` module so I can stop duplicating code. It'll be published, but it'll be considered private API that can change at any time. What do you think?

JakeWharton: We could try to set up repackaging so it's not actually published but just locally shared. Failing that, we could annotate everything with an `@OptIn` annotation to convey that it's unstable API. All of that can come later, though. Extracting a published module to eliminate duplication seems fine.

JakeWharton (approved):

redwood-layout-composeui/src/main/kotlin/app/cash/redwood/layout/composeui/utils.kt
colinrtwhite: Sounds good, I'll try sharing it locally + repackaging. Do you know if we've done something similar previously in any of Square/Cash App's repos?

JakeWharton: I've only ever done it for a jar-producing module. Since the ones we want to use it on are Android modules that complicates things outside the realm of what I've done in the past.

Here's the jar one: https://github.com/cashapp/redwood/blob/b8aac3644337e557d434d0dbb5e6dba90c324bec/treehouse/compose/compiler/build.gradle

PR #523, opened 2022-11-17
Remove support for flex item order.

colinrtwhite:
Removes support for the flex item `order` param. I don't think we should expose this in the redwood layout API since it's confusing and removing it allows us to clean up the flex container code. Also you can always change the order in the `items` list to solve the same use case.


adrw (approved):
Good call, simpler API will be easier to maintain

JakeWharton (approved):

PR #526, opened 2022-11-18
Move ID generation out of applier

JakeWharton:
An `IdAllocator` type creates IDs and is owned by the root factory. This allows IDs to be created synchronously at construction time.

Some other small changes that are tangentially related are included in this change:
- The type is not longer held in a backing field and instead each subtype has a property getter to return its value.
- Elements of the diff-producing widgets have been moved to the `DiffProducingWidget` interface such as ID, type, and the ability to send an event.

This is the first in many steps toward sharing an Applier implementation between protocol-based and non-protocol-based usage.

Refs #14.

### DIFF
--- a/redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/AbstractDiffProducingWidget.kt
+++ b/redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/AbstractDiffProducingWidget.kt
@@ -15,8 +15,6 @@
  */
 package app.cash.redwood.protocol.compose
 
-import app.cash.redwood.protocol.Event
-import app.cash.redwood.protocol.Id
 import app.cash.redwood.protocol.LayoutModifiers
 import app.cash.redwood.protocol.PropertyDiff
 import app.cash.redwood.widget.Widget.Children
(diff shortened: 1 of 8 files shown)


redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/AbstractDiffProducingWidget.kt
```diff
@@ -15,24 +15,14 @@
...
-public abstract class AbstractDiffProducingWidget(
-  public val type: Int,
-) : DiffProducingWidget {
-  override val value: Nothing
-    get() = throw AssertionError()
-
-  public var id: Id = Id(ULong.MAX_VALUE)
-    internal set
-
+public abstract class AbstractDiffProducingWidget : DiffProducingWidget {
```
JakeWharton, line 25: This type will eventually disappear altogether.

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/IdAllocator.kt
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package app.cash.redwood.protocol.compose
+
+import app.cash.redwood.protocol.Id
+
+/**
+ * @suppress For generated code usage only.
+ */
+public class IdAllocator {
```
JakeWharton, line 23: This type will get a new name and a lot more responsibility over the coming PRs. It becomes the Compose-side version of `ProtocolDisplay`.

colinrtwhite (approved):

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/IdAllocator.kt
colinrtwhite: Will this eventually get the responsibility of converting Compose changes into diffs? Or will that still live in `ProtocolApplier`?

PR #527, opened 2022-11-18
Rename `ProtocolDisplay` to `ProtocolBridge`

JakeWharton:
We are going to have a symmetric type on the Compose side soon which will share its name.

Going for
```mermaid
flowchart LR;
    Compose-->DiffProducingWidgetFactory;
    DiffProducingWidgetFactory-->PB1[ProtocolBridge];
    PB1[ProtocolBridge] --> Protocol;
    Protocol --> ProtocolBridge;
    ProtocolBridge -->DiffConsumingWidgetFactory;
    DiffConsumingWidgetFactory-->Widgets;

```

### DIFF
--- a/redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
+++ b/redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
@@ -22,17 +22,25 @@ import app.cash.redwood.protocol.Diff
 import app.cash.redwood.protocol.DiffSink
 import app.cash.redwood.protocol.EventSink
 import app.cash.redwood.protocol.Id
+import app.cash.redwood.protocol.LayoutModifiers
 import app.cash.redwood.protocol.PropertyDiff
 import app.cash.redwood.widget.Widget
 import kotlinx.serialization.json.JsonArray
 
-public class ProtocolDisplay<T : Any>(
+/**
+ * Bridges the serialized Redwood protocol back to widgets on the display side.
+ *
+ * This type will consume [Diff]s and apply their [ChildrenDiff] operations to the widget tree.
+ * [PropertyDiff]s and [LayoutModifiers]s are forwarded to their respective widgets. Events from
+ * widgets are forwarded to [eventSink].
+ */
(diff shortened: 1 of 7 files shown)


colinrtwhite (approved):
Just to make sure I understand - so the plan is to eventually be able to pass a custom `Protocol` to `ProtocolRedwoodComposition` so we can skip diff serialization/deserialization if we're not using Zipline?

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
```diff
@@ -22,17 +22,25 @@ import app.cash.redwood.protocol.Diff
...
 
-public class ProtocolDisplay<T : Any>(
+/**
+ * Bridges the serialized Redwood protocol back to widgets on the display side.
+ *
+ * This type will consume [Diff]s and apply their [ChildrenDiff] operations to the widget tree.
+ * [PropertyDiff]s and [LayoutModifiers]s are forwarded to their respective widgets. Events from
+ * widgets are forwarded to [eventSink].
+ */
+public class ProtocolBridge<T : Any>(
```
colinrtwhite, line 37: Why have two types with the same name if one takes diffs as inputs and the other as outputs?

JakeWharton:
If you don't want to use the protocol to serialize the values then the Compose code will talk directly to the Widget code. It already does this today, as the diff-producing factory implements the same widget factory that you would implement. So you choose whether it talks directly to your widget factory or whether it talks to a widget factory that records operations as diffs in the protocol bridge. Then, on the other side, there's an inverse protocol bridge that unwraps diffs, talks to a diff-consuming widget factory, and then it talks to your real widget factory.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
JakeWharton: I couldn't think of a good-enough descriptive name. They each have inputs and outputs–they're just mirrors of each other. The compose-side has diff outputs and event inputs and the widget-side has event outputs and diff inputs.

PR #528, opened 2022-11-21
Add TextFieldState to fix edit text race conditions.

### LINKED ISSUE #380: EmojiSearch sample needs a TextFieldState type
To defend against racy edits.

colinrtwhite:
Resolves: https://github.com/cashapp/redwood/issues/380

### DIFF
--- a/samples/emoji-search/android-composeui/src/main/java/app/cash/zipline/samples/emojisearch/composeui/ComposeUiTextInput.kt
+++ b/samples/emoji-search/android-composeui/src/main/java/app/cash/zipline/samples/emojisearch/composeui/ComposeUiTextInput.kt
@@ -15,30 +15,54 @@
  */
 package app.cash.zipline.samples.emojisearch.composeui
 
-import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.padding
 import androidx.compose.material.Text
 import androidx.compose.material.TextField
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
-import androidx.compose.ui.Modifier
-import androidx.compose.ui.unit.dp
+import androidx.compose.ui.text.TextRange
+import androidx.compose.ui.text.input.TextFieldValue
 import app.cash.redwood.LayoutModifier
 import example.schema.widget.TextInput
+import example.values.TextFieldState
 
 class ComposeUiTextInput : TextInput<@Composable () -> Unit> {
   private var hint by mutableStateOf("")
-  private var text by mutableStateOf("")
-  private var onTextChanged: ((String) -> Unit)? = null
+  private var composeState by mutableStateOf(TextFieldValue())
+  private var redwoodState by mutableStateOf(TextFieldState())
+  private var onTextChanged: ((TextFieldState) -> Unit)? = null
 
   override var layoutModifiers: LayoutModifier = LayoutModifier
 
   override val value = @Composable {
+    // not sure which of these are correct
+    // the first has one source of truth, but latency in the UI and a lot of TextField state bouncing around internally
+    // the second results in less latency in the UI but two sources of truth
+    val composeStateWithLastTextValue = composeState.copy(text = redwoodState.text)
+    // val composeStateWithLastTextValue = composeState
+
+    println("!!!Android - Recomposing with $composeStateWithLastTextValue")
     TextField(
-      value = text,
-      onValueChange = { onTextChanged?.invoke(it) },
+      value = composeStateWithLastTextValue,
+      onValueChange = { newComposeState ->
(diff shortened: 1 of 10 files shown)


samples/emoji-search/android-composeui/src/main/java/app/cash/zipline/samples/emojisearch/composeui/ComposeUiTextInput.kt
```diff
@@ -15,44 +15,91 @@
...
 
   override val value = @Composable {
     TextField(
-      value = text,
-      onValueChange = { onTextChanged?.invoke(it) },
-      label = { Text(hint) },
+      value = TextFieldValue(
+        text = state.text,
+        selection = TextRange(state.selectionStart, state.selectionEnd)
+      ),
```
eygraber: `CoreTextField` will call `onValueChange` when the IME updates the `composition` property on `TextFieldValue`. Preserving that value is why I was keeping track of the `TextFieldValue` coming back from `onValueChanged` (similar to what the `String` overload of `BasicTextField` does) but that also led to some more tricky state handling.

eygraber: Running the sample on this branch shows that the `composition` is getting dropped (e.g. the IME underlines the text that is being typed). I don't know if that's a big deal or not, but I thought it would be good to try and keep it.

colinrtwhite: @eygraber Great callout - I added a commit to remember the latest `TextFieldValue` to preserve that state without recomposing.

samples/emoji-search/ios/app/EmojiSearchApp/TextInputBinding.swift
```diff
@@ -24,24 +24,24 @@ class TextInputBinding: WidgetTextInput {
...
     }
 
-    func onTextChanged(onTextChanged: ((String) -> Void)? = nil) {
+    func onChange(onChange: ((ValuesTextFieldState) -> Void)? = nil) {
         let identifier = UIAction.Identifier("TextInputBinding.onTextChanged")
 
         root.removeAction(identifiedBy: identifier, for: .editingChanged)
         root.addAction(UIAction(identifier: identifier, handler: { [unowned self] _ in
-            onTextChanged?(self.root.text ?? "")
+            onChange?(ValuesTextFieldState(text: self.root.text ?? "", selectionStart: 0, selectionEnd: 0, userEditCount: 0))
```
colinrtwhite, line 41: We'll eventually need to handle this properly once https://github.com/cashapp/redwood/issues/370 is resolved.

samples/emoji-search/android-composeui/src/main/java/app/cash/zipline/samples/emojisearch/composeui/ComposeUiTextInput.kt
eygraber: This causes an infinite loop of `onValueChange` being called when any text is typed.

eygraber: Making `textFieldValue` a mutable state fixes the issue (which led to my questions about multiple sources of truth). After experimenting a bunch, I found that `TextField` gets really unhappy if you leave the `onValueChange` block without using the `TextFieldValue` parameter to update `TextField.value` (unless you don't plan on updating the value at all).

eygraber: In fact reverting `1f53621b7d` will also lead to the same infinite loop of `onValueChange` invocations.

colinrtwhite: @eygraber Ah yep, you're right - as far as I can tell `onValueChange` loops forever if you don't use its value. We could probably create some higher order state object that combines the values from `TextFieldValue` and `TextFieldState`, but that's probably overkill. Since we use `textFieldValue.copy` the `text` and `selection` values on `TextFieldValue` are always ignored.

eygraber: :+1: 

In theory there can be a case where the presenter mutates the text under certain conditions that can be skipped if the user edit takes precedence over the Treehouse update (as a contrived example, a presenter that truncates the text to 9 characters if the text is 10 characters long). Can't think of any real world examples at the moment, and also having trouble deciding if that's a bug or just a consequence of text input :man_shrugging:

JakeWharton (approved):

PR #530, opened 2022-11-21
Drop the event and diff side-effects

JakeWharton:
You can always wrap if you need these.

Do not lament the loss of the log statements in the samples, for these are the ones which will be migrated to not use the protocol altogether in a follow-up!


swankjesse (approved):

swankjesse:
Good riddance!

PR #532, opened 2022-11-21
Move Compose state management out of the applier

JakeWharton:
And into a new(ish) type, `ProtocolBridge`, a mirror image of the `ProtocolBridge` that sits on the widget side.

Refs #14. This unblocks the ability to move the applier into the compose module, break the dependency between the compose and protocol-compose module, and run composition without the protocol!


redwood-protocol-compose/src/commonTest/kotlin/app/cash/redwood/protocol/compose/DiffProducingWidgetFactoryTest.kt
```diff
@@ -41,23 +41,18 @@ class DiffProducingWidgetFactoryTest {
...
     textInput.customType(10.seconds)
-    diffAppender.trySend()
 
     val expected = Diff(
       propertyDiffs = listOf(
         PropertyDiff(Id(1U), 2U, JsonPrimitive("PT10S")),
       ),
     )
-    assertEquals(expected, diffSink.diffs.single())
+    assertEquals(expected, bridge.createDiffOrNull())
```
JakeWharton, line 55: This test diff really demonstrates what a superior design this is.

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolRedwoodComposition.kt
```diff
@@ -25,15 +25,15 @@ import androidx.compose.runtime.snapshots.ObserverHandle
...
+import app.cash.redwood.protocol.Id
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 
-public interface ProtocolRedwoodComposition : RedwoodComposition, EventSink {
+public interface ProtocolRedwoodComposition : RedwoodComposition {
   public fun start(diffSink: DiffSink)
 }
```
JakeWharton, line 38: This type will be going away soon! We won't need it with the new design. Knowledge of the protocol is entirely separated from knowledge of Compose.

swankjesse (approved):

colinrtwhite (approved):

PR #535, opened 2022-11-22
Extract Lazy Layout properties to publishable module

veyndan:


redwood-treehouse-lazylayout-api/gradle.properties
```diff
@@ -0,0 +1,2 @@
+POM_ARTIFACT_ID=redwood-treehouse-lazylayout-api
```
veyndan, line 1: Instead of putting this within `redwood-layout-api`, I moved this to a separate module, as technically `LazyColumn` only works with Treehouse as it utilizes Zipline.

I named it `lazylayout` as that's the terminology that they use [here](https://developer.android.com/jetpack/compose/lists), as we're not restricted to lazy lists, as we could have lazy grids too.

Open to changing either of these things.

JakeWharton (approved):

Issue #536, opened 2022-11-22
UI tests for counter sample


JakeWharton:
* [x] Android Compose UI
* [x] Android views
* [ ] Browser
* [ ] iOS

I wrote some Android ones two years ago here: https://github.com/cashapp/redwood/commit/f777de3dcf7f949175ac35d52d4d24beddf4c900. Should be easy to update and land for Android. Although we should ideally share the tests between both Android frontends.

JakeWharton:
I wrote a UI Automator test which works for both Android frontends. Might want to try something like Appium for a test that works on all four frontends, though.

PR #539, opened 2022-11-22
Add UI Automator tests for Android Counter samples

JakeWharton:
Also refactor CI build to build samples separately.

Refs #536


colinrtwhite (approved):
Awesome! Looks like this shaves 10 minutes off `build`

JakeWharton:
Yeah and there's more to be had. Dokka is an atrocity and takes like 10m on its own. I'm tempted to remove it completely from PR/branch builds. The change that it fails is extremely low.

PR #545, opened 2022-11-23
Support running `RedwoodComposition` without the protocol

### LINKED ISSUE #14: Polymorphic applier/scope to avoid protocol indirection

### LINKED ISSUE #492: JS counter app broken
Is this the main entrypoint for the JS counter app?
```
./gradlew browserRun
```
It compiles and opens the browser with html that looks reasonable to me, but a JS error happens. The Chrome console shows an error I don't understand:
```
[HMR] Waiting for update signal from WDS...
counter.js:495 Uncaught TypeError: p0.SunspotBox_9mogxp_k$ is not a function
    at l (SunspotBox.kt?b9ba:16:17)
    at eval (SunspotBox.kt?b9ba:27:1634)
    at eval (Composer.kt?cb50:1537:24)
    at eval (Composer.kt?cb50:33:2)
    at applyChangesInLocked (Composition.kt?3de3:34:9)
    at CompositionImpl.applyChanges_yo70f0_k$ (Synchronized.kt?bedf:22:31650)
    at Recomposer.composeInitial$composable_e2dr50_k$ (Recomposer.kt?99ae:827:21)
    at CompositionImpl.setContent$composable_jlpbvd_k$ (Composition.kt?3de3:519:16)
    at DiffProducingRedwoodComposition.setContent$composable_jlpbvd_k$ (ProtocolRedwoodComposition.kt?435c:131:17)
    at main (main.kt?11a0:52:15)
index.js?b39d:551 [webpack-dev-server] Hot Module Replacement enabled.
index.js?b39d:551 [webpack-dev-server] Live Reloading enabled.
```

JakeWharton:
We did it! Migrate the counter sample frontends to this new API and drop the protocol modules.

Closes #14. Closes #492.


samples/counter/android-views/src/main/java/example/android/MainActivity.kt
```diff
@@ -38,25 +32,11 @@ class MainActivity : AppCompatActivity() {
...
 
-    val composeBridge = ComposeProtocolBridge()
-
-    val root = FrameLayout(this).apply {
-      layoutParams = LayoutParams(MATCH_PARENT, MATCH_PARENT)
-    }
-    setContentView(root)
-
-    val factory = DiffConsumingSunspotWidgetFactory(AndroidSunspotWidgetFactory(this))
-    val widgetBridge = ProtocolBridge(
```
swankjesse, line 49: Nice

swankjesse (approved):

swankjesse:
# 💎

PR #547, opened 2022-11-23
Add OptIn annotation for generated code APIs

### LINKED ISSUE #542: An `@OptIn` annotation for generated-use only
A small speed-bump to prevent anyone from using our APIs which target generated code only. Generate `@OptIn` on all our generated code to suppress.

This may have to live in the runtime since it needs to be accessible to the Compose code, the diff-producing widget code, and the diff-consuming widget code.

JakeWharton:
Speed bumps to try and prevent manual usage since interaction is delicate.

Closes #542. Even though it's only used by Compose code, I expect to use it in widget code after some refactorings so I chose to put it in 'runtime'.

### DIFF
--- a/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/annotations.kt
+++ b/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/annotations.kt
@@ -20,3 +20,11 @@ package app.cash.redwood
  */
 @DslMarker
 public annotation class LayoutScopeMarker
+
+/**
+ * Denote an API which should only be used by Redwood's generated code.
+ *
+ * @suppress
+ */
+@RequiresOptIn("This API is for use by Redwood's generated code only")
+public annotation class RedwoodCodegenApi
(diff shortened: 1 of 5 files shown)


swankjesse (approved):

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/annotations.kt
```diff
@@ -20,3 +20,11 @@ package app.cash.redwood
  */
 @DslMarker
 public annotation class LayoutScopeMarker
+
+/**
+ * Denote an API which should only be used by Redwood's generated code.
+ *
+ * @suppress
+ */
+@RequiresOptIn("This API is for use by Redwood's generated code only")
```
colinrtwhite, line 29: Should we use `level = RequiresOptIn.Level.ERROR` for this annotation? It forces the user to add the opt in annotation whereas a warning can be ignored. `@InternalComposeApi` also uses `Level.ERROR`.

JakeWharton: I believe that's the default.

colinrtwhite: Ah yep it is 🤦 . For some reason I thought it warned by default.

PR #550, opened 2022-11-23
Do not double wrap native widgets for diff consumption

JakeWharton:
Move responsibilities up into a base type for the generated diff consuming node. Additionally, there is no reason for this generated node to actually implement `Widget` and so that is dropped.

This is, admittedly, a little sloppy. It's merely a step on a larger refactoring that I wanted to fire off before the long weekend.

### DIFF
--- a/redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/DiffConsumingNode.kt
+++ b/redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/DiffConsumingNode.kt
@@ -16,17 +16,29 @@
 package app.cash.redwood.protocol.widget
 
 import app.cash.redwood.protocol.EventSink
+import app.cash.redwood.protocol.Id
 import app.cash.redwood.protocol.PropertyDiff
 import app.cash.redwood.widget.Widget
 import kotlinx.serialization.json.JsonArray
 
 /**
  * A [Widget] which consumes protocol diffs and applies them to a platform-specific representation.
+ *
+ * @suppress
  */
-public interface DiffConsumingWidget<W : Any> : Widget<W> {
-  public fun apply(diff: PropertyDiff, eventSink: EventSink)
+public abstract class DiffConsumingNode<W : Any>(
+  public val parentId: Id,
+  public val parentChildren: Widget.Children<W>,
+) {
+  private var _childIds: MutableList<Id>? = null
+  public val childIds: MutableList<Id>
+    get() = _childIds ?: mutableListOf<Id>().also { _childIds = it }
 
-  public fun updateLayoutModifier(value: JsonArray)
+  public abstract val widget: Widget<W>
+
+  public abstract fun apply(diff: PropertyDiff, eventSink: EventSink)
+
+  public abstract fun updateLayoutModifier(value: JsonArray)
 
   /**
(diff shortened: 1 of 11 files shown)


redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/DiffConsumingNode.kt
```diff
@@ -16,17 +16,29 @@
...
 import kotlinx.serialization.json.JsonArray
 
 /**
  * A [Widget] which consumes protocol diffs and applies them to a platform-specific representation.
+ *
+ * @suppress
  */
-public interface DiffConsumingWidget<W : Any> : Widget<W> {
-  public fun apply(diff: PropertyDiff, eventSink: EventSink)
+public abstract class DiffConsumingNode<W : Any>(
```
JakeWharton, line 29: This type will eventually be annotated with `@RedwoodCodegenApi` but that currently also affects the nested `Factory` which is public API. These two types will split in a follow-up at which point this class will receive the annotation.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
```diff
@@ -75,37 +75,26 @@ public class ProtocolBridge<W : Any>(
     for (layoutModifier in diff.layoutModifiers) {
       val node = node(layoutModifier.id)
       val childIndex = node(node.parentId).childIds.indexOf(layoutModifier.id)
-      node.widget.updateLayoutModifier(layoutModifier.elements)
+      node.updateLayoutModifier(layoutModifier.elements)
       node.parentChildren.onLayoutModifierUpdated(childIndex)
```
JakeWharton, line 79: This logic can be dramatically cleaned up if we stop delivering `LayoutModifiers` to the `Widget` and _only_ deliver them to the parent. The protocol entry for layout modifier changes can reference the parent ID, the children tag, and the index inside that set of children on which the new value applies. That will eliminate nearly 100% of the `parentId`/`parentChildren` wiring present in this change.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/DiffConsumingNode.kt
```diff
@@ -16,17 +16,29 @@
...
 import app.cash.redwood.protocol.EventSink
+import app.cash.redwood.protocol.Id
 import app.cash.redwood.protocol.PropertyDiff
 import app.cash.redwood.widget.Widget
 import kotlinx.serialization.json.JsonArray
 
 /**
  * A [Widget] which consumes protocol diffs and applies them to a platform-specific representation.
+ *
+ * @suppress
```
swankjesse, line 27: What does this do?

swankjesse (approved):

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/DiffConsumingNode.kt
JakeWharton: Omit from Dokka

PR #551, opened 2022-11-24
Ensure git index is clean after builds

### LINKED ISSUE #546: Add GitHub Action to ensure git isn't dirty after build
This protects against forgetting to update `kotlin-js-store/` in PRs.

JakeWharton:
Closes #546.

First commit has two failing shards to demonstrate behavior: one with a modified tracked file and one with an untracked file. Second commit removes those from being generated and will pass.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -31,7 +31,9 @@ jobs:
           path: build/dokka/htmlMultiModule/
           if-no-files-found: error
 
-      - run: echo "// test" >> build.gradle
+      # Run the npm install on the full project. Without this, the partial build above updates the
+      # lock file with a subset of the full project's needs (thus making the git index dirty).
+      - run: ./gradlew kotlinNpmInstall
 
       - name: Ensure git index is clean
         run: test -z "$(git status --porcelain | tee /dev/fd/2)"
@@ -50,8 +52,6 @@ jobs:
       - uses: gradle/gradle-build-action@v2
 
       - run: ./gradlew verifyPaparazziDebug
-
-      - run: echo "test" > sup.txt
 
       - name: Ensure git index is clean
         run: test -z "$(git status --porcelain | tee /dev/fd/2)"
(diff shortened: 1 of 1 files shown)


JakeWharton:
Ahaha it already caught a problem

JakeWharton:
Oh no. This is because we exclude the samples in this job. Hmmmmmmmmm

.github/workflows/build.yaml
```diff
@@ -25,6 +25,9 @@ jobs:
 
       - run: ./gradlew build dokkaHtmlMultiModule --parallel -PredwoodNoSamples
 
+      - name: Ensure git index is clean
+        run: test -z "$(git status --porcelain | tee /dev/fd/2)"
```
swankjesse, line 39: TIL https://www.youtube.com/watch?v=IJWlBfo5Oj0

colinrtwhite: Maybe we should only check this as part of the other jobs? The check will still work since those other jobs will fail  if `kotlin-js-store` isn't up-to-date.

Could also create a new job that only runs `assemble` or something minimal then checks the git index.

colinrtwhite (approved):

.github/workflows/build.yaml
JakeWharton: I guess it depends on when it's written. Is it during configuration? Because I'd be happy to run a no op like `tasks` on the whole project to ensure the entire configuration runs.

colinrtwhite: Looks like it's generated by [`kotlinNpmInstall`](https://kotlinlang.org/docs/js-project-setup.html#version-locking-via-kotlin-js-store), which is pretty quick. I tried deleting the file and `kotlinNpmInstall` recreated it.

JakeWharton: Yeah I'm just worried because that fixes only this specific case. I'll add it and see if we run into problems like this in the future. Frankly, this is the only file I'm truly worried about anyway.

JakeWharton:
Gah why did it merge on red?

JakeWharton:
Screen Shot 2022-11-29 at 9 35 56 PM

JakeWharton:
Surely skipped does not mean it is treated as success...

PR #554, opened 2022-11-27
Update Gradle to 7.6. Update CI to Java 19.

colinrtwhite:


JakeWharton (approved):

colinrtwhite:
Interesting `RedwoodLayoutWidgetFactory` resolves fine, but `RedwoodTreehouseLazyLayoutWidgetFactory` isn't resolved.

EDIT: Whichever is first in the `dependencies` block gets resolved. The other doesn't 🤔 .

colinrtwhite:
If modules have the same artifact name (in this case `widget`), only the first is resolved. Seems like a bug with Gradle 7.6. I'll see if I can create a quick repro project and file an issue.

EDIT: Maybe related to https://github.com/gradle/gradle/issues/847? Not sure why this would be affecting us now.

JakeWharton:
The samples are no longer a standalone project. They should not have a wrapper of their own and it can be deleted.

colinrtwhite:
Created https://github.com/gradle/gradle/issues/22867. The only effective work-around I've found so far is to rename the modules to different names.

colinrtwhite:
Whoops, looks related to https://github.com/cashapp/redwood/pull/553? Not sure how it made it into `trunk` since CI didn't pass. Reverting it fixes the issue, though I don't know why that commit would be the root cause.

I think we need to enable "Require status checks to pass before merging" in settings.

JakeWharton:
It's enabled but the check name changed. So his branch was probably based on an old merge base.

colinrtwhite:
This is strange... `build / sample-emoji (pull_request)` fails, but `build / sample-emoji (push)` doesn't. I updated to not include the latest commit on `trunk`.

EDIT: The `pull_request` job and `push` job are checking out two different commits.

EDIT 2: `pull_request` automatically merges `trunk` into the branch. `push` does not. Interesting!

colinrtwhite:
This breaks the `serveDevelopmentZipline` task:

```
BUILD SUCCESSFUL in 4s
164 actionable tasks: 10 executed, 154 up-to-date
Watched directory hierarchies: []

Exiting continuous build as Gradle does not watch any file system locations.
Stopped Server@17f51ce1{STOPPING}[11.0.11,sto=5000]
Shutdown Server@17f51ce1{STOPPING}[11.0.11,sto=5000]
Stopped ServerConnector@737c8a33{HTTP/1.1, (http/1.1)}{0.0.0.0:8080}
Stopped o.e.j.s.ServletContextHandler@1859f943{/,null,STOPPED}
```

I think this PR should fix: https://github.com/cashapp/zipline/pull/753

PR #559, opened 2022-11-27
Revert "Make treehouse-lazylayout-schema publishable (#553)"

colinrtwhite:
Reverts 54720fea2853c478b705e9a2dccde96e2cd931e1 to fix `trunk`. `build / sample-emoji (pull_request)` automatically merges `trunk` when building so that job will fail for all PRs.

### DIFF
--- a/redwood-treehouse-lazylayout-schema/build.gradle
+++ b/redwood-treehouse-lazylayout-schema/build.gradle
@@ -1,6 +1,4 @@
 apply plugin: 'org.jetbrains.kotlin.jvm'
-apply plugin: 'com.vanniktech.maven.publish'
-apply plugin: 'org.jetbrains.dokka' // Must be applied here for publish plugin.
 
 dependencies {
   api projects.redwoodSchemaAnnotations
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):

JakeWharton:
Are we missing an archivesBaseName? I think the flexbox one has it.

colinrtwhite:
Looks like `archivesBaseName` is only declared for `compose`/`widget`/`protocol` modules for the emoji and counter samples.

JakeWharton:
Yeah only for when we have a collision on the projects name like it seems these would.

colinrtwhite:
Hmm I tried building locally with custom `archivesBaseName`s for each `widget` module but ran into the same compilation issue. [CI build.](https://github.com/cashapp/redwood/tree/colin/lazylayout_publish)

JakeWharton:
I haven't been at a computer since Wednesday so just speculation from afar.

colinrtwhite:
Ah gotcha! Well thanks for all the weekend reviews then. Worst case we can rename the modules to something like `layout-widget` and `lazylayout-widget`, but still publish them with the same coordinates to work around the issue.

PR #561, opened 2022-11-28
Allow reset to clear out-of-band views

JakeWharton:
Child views such as progress indicators may be used with Treehouse and are not tracked by the normal children mechanism.

This isn't the most satisfying solution, but I think some forthcoming refactors will actually make this better. We cannot simply throw away the `Widget.Children` instance because it's captured before the call to `reset()`.

### DIFF
--- a/redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseWidgetView.kt
+++ b/redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseWidgetView.kt
@@ -45,7 +45,8 @@ public class TreehouseWidgetView<A : Any>(
       }
     }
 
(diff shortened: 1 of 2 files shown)


redwood-treehouse/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseWidgetView.kt
```diff
@@ -54,7 +55,10 @@ public class TreehouseWidgetView<A : Any>(
     get() = mutableHostConfiguration
 
   override fun reset() {
-    children.remove(0, childCount)
+    _children.remove(0, _children.widgets.size)
+
+    // Ensure any out-of-band views are also removed.
```
swankjesse, line 60: I like it.

swankjesse (approved):

Issue #541, opened 2022-11-23
Split `TreehouseView`


JakeWharton:
Currently it mixes two concepts which makes the API less-than-ideal:
- Functions for the app to control our implementations for Android views, Compose UI, and UIKit
   - `boundContent`
   - `hostConfiguration`
   - `children`
   - `reset()`
- Configuration and callbacks for the specific rendering
   - `widgetSystem`
   - `codeListener`

You can also tell that it's slightly weird because each view subtype should be able to force a specific type parameter for the widget factory but currently cannot.

I think `widgetSystem` and `codeListener` can be combined, and in the process some of the listener callbacks eliminated. Not sure what to call this type.

The others can stay together and exist only to bridge the app to the rendering layer but otherwise their APIs should never be seen by callers.

JakeWharton:
Okay I started pulling on this thread and it's promising. But first it actually looks like there's some code-gen changes that would also help simplify Treehouse setup. Leaving until next week.

swankjesse:
One quick note on combining `widgetSystem` and `codeListener` . . . in our current adoption there’s some TreehouseWidgetViews that use a loading spinner, and some that don’t. Both of these share the same widgets.

So wherever we land it shouldn’t be much work to configure the factory & loading spinner independently.

JakeWharton:
Expanding on this more: the current design makes it nearly impossible to unit test `TreehouseView` subtypes. They're currently too smart, whereas we probably want them to be instantiable without a `TreehouseApp` and generally know nothing about the `A` type parameter. They're just focused on `W`.

I think what we want is a design where the `TreehouseView` subtype knows about the widget factory, has callbacks for lifecycle events, and exposes the `HostConfiguration` as a `StateFlow`. It's nothing but a very, VERY fancy `Button`. Or a much simpler `RecyclerView`. Just a UI node with no smarts.

Then, separately, you bind a `TreehouseView` to a `TreehouseApp` explicitly (rather than implicitly via constructor). This knows nothing of the `W` type parameter but only knows about `A`. It is also where you provide the `Content`. If you want to change the `Content` of an existing `TreehouseView` you re-`bind` it. If you want to bind an existing `TreehouseView` to new content you re-`bind` it.

This will make testing the `TreehouseView` subtypes simple. It will also make use in a recycled view situation easier. It should also mean the death of `WidgetSystem` and the re-injection of the `TreehouseApp` as a parameter. Recursive adapters for lazy column should be easier to create as a result.

[...]

PR #562, opened 2022-11-28
Reduce listener callbacks to one

JakeWharton:
This listener is was used for loading indicators and displaying animations on reloads. For loading indicators, display them upon initial creation of the TreehouseView subtype or when one is reused (such as in a data binding list view). For animating, the provided boolean differentiates first load from subsequent loads and it runs at the same time as before.

The view subtype parameter is not needed since you can capture the actual TreehouseView subtype when creating the lambda.

Start of #541

### DIFF
--- a/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
+++ b/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
@@ -28,7 +28,6 @@ import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
 import kotlinx.serialization.modules.EmptySerializersModule
 import kotlinx.serialization.modules.SerializersModule
 
@@ -98,7 +97,7 @@ public class TreehouseApp<A : Any> internal constructor(
   }
 
   /** This function may only be invoked on [TreehouseDispatchers.zipline]. */
-  private suspend fun bind(view: TreehouseView<A>) {
+  private fun bind(view: TreehouseView<A>) {
     dispatchers.checkZipline()
 
     // Make sure we're tracking this view, so we can update it when the code changes.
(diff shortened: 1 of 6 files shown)


redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -111,15 +110,7 @@ public class TreehouseApp<A : Any> internal constructor(
       viewToBoundContent.remove(view)
     }
 
-    val ziplineSession = this.ziplineSession
-    if (ziplineSession != null) {
-      ziplineSession.bind(view, content)
-    } else {
-      // If we're waiting for code to load, show a loading indicator until it's ready.
```
swankjesse, line 118: I don’t love the slight regression here . . . the caller doesn’t know if code is already ready, so they potentially waste work creating spinners & placeholders that aren’t necessary

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseView.kt
```diff
@@ -45,14 +45,11 @@ public interface TreehouseView<A : Any> {
...
-    public open fun beforeInitialCode(view: TreehouseView<*>) {}
-
-    /** Clear the previous UI and show a quick animation for subsequent code updates. */
-    public open fun beforeUpdatedCode(view: TreehouseView<*>) {}
+  public fun interface CodeListener {
+    /**
+     * Invoked each time new code is loaded. This is called after the view's old children have
+     * been cleared but before the children of the new code have been added.
+     */
+    public fun onCodeLoaded(initial: Boolean)
```
swankjesse, line 53: 👍🏻

swankjesse (approved):

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
JakeWharton: Ah, that makes sense. We can keep it for that.

JakeWharton: #566

PR #563, opened 2022-11-29
Use codegen to de-duplicate Compsoe helpers

JakeWharton:
This isn't great, but it's also not terrible. These never change so not really a big deal. The biggest downside is that the IDE does not run the task on sync and so you need to do a build before they resolve in the IDE. They will always link during compilation.


JakeWharton:
What do you think about this?

A potential future optimization is to actually put these into a library module (which isn't published) and have the task consume the sources from that library module, rewrite the package name, and copy them into the destination projects. That may be needed if we want to take the same approach with the flexbox helpers.

JakeWharton:
Could also put it into `src/main/resources/` so that it's at least a `.kt` file, but it won't be linked against the libraries it depends on.

In that form, if you needed to change it (with IDE help) you would perform the changes in one of the generated files and then copy/paste that into the resources version. I'm going to do that. Seems like a nice balance of simple and functional. I will also migrate the flexbox stuff to it.

colinrtwhite:
@JakeWharton The `src/main/resources/` solution sounds good to me! I think it's fine that it takes one build to resolve the imports.

build-support/src/main/resources/app/cash/redwood/buildsupport/flexboxLayout.kt
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example
+
+import app.cash.redwood.flexcontainer.AlignItems
```
colinrtwhite: I'd probably rename this file to `flexContainer` or something similar to match the package names of the other modules. I migrated all flexbox references to flex container a while back.

colinrtwhite (approved):

build-support/src/main/resources/app/cash/redwood/buildsupport/flexboxLayout.kt
JakeWharton: Yes. I will change.

Although... I've been biding my time for when to bring up that my original proposal was that the package name and project name should still be 'flexbox' and its only the type name that I thought should be `FlexContainer`. But that's a separate thing.

colinrtwhite: Changing the package back to `flexbox` sounds fine to me!

PR #564, opened 2022-11-29
Revert "Update Gradle to 7.6. Update CI to Java 19. (#554)"

colinrtwhite:
Reverting this as it breaks `serveDevelopmentZipline`. We can update once there's a new Zipline release that includes https://github.com/cashapp/zipline/commit/ff76a3254ff1628514443d9548a839b419cd35cb.

This reverts commit d1388da2e0c07f6e2f15decef78f7552d1a04d4a.


fryn (approved):
Thanks! 👍🏽

redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodLintTask.kt
```diff
@@ -40,8 +37,7 @@ internal abstract class RedwoodLintTask @Inject constructor(
   @get:Classpath
   abstract val toolClasspath: ConfigurableFileCollection
 
-  @get:InputDirectory
-  @get:PathSensitive(PathSensitivity.ABSOLUTE)
+  @get:Input
```
JakeWharton: This doesn't need reverted, right? It was just incorrect always.

samples/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-bin.zip
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
```
JakeWharton: These two files don't need re-added. They shouldn't have been there in the first place.

JakeWharton (approved):
Oh shit my git check actually worked!

redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodLintTask.kt
colinrtwhite: Yep! I also removed `@get:PathSensitive(PathSensitivity.ABSOLUTE)` since `ABSOLUTE` is the default.

PR #565, opened 2022-11-29
Delete screenshot failure diff files.

colinrtwhite:
These screenshot diff files are unused and shouldn't be committed - added their directories to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ xcuserdata
 
 # Cocoapods
 Pods
+
+# Paparazzi
+redwood-layout-composeui/out/failures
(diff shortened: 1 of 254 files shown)


JakeWharton (approved):
Hmm why are these being generated in user folders instead of build?

colinrtwhite:
Seems to be [working as intended](https://github.com/cashapp/paparazzi/blob/c7c19125fa05f044c5d865f42f970776f3171fd8/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt#L56). Agreed it should probably be updated to `project.buildDir`!

Issue #567, opened 2022-11-30
Figure out why automerge allowed red code


JakeWharton:
See #551

JakeWharton:
For the time being I have explicitly listed our four jobs as being required.

JakeWharton:
According to the docs the fact that the job was skipped means it gets reported as successful. This is different to a job which does not run due to a conditional which would still be considered pending and thus not be successful. So, this "trick" of having a single fan-in job on which we can gate auto merge does not actually work, and we must maintain a list of statuses in branch protection that match the jobs in the action yaml. Super gross.

PR #568, opened 2022-11-30
Fix the build after partial revert

JakeWharton:
This annotation is apparently required with `@InputDirectory`.


colinrtwhite:
Ah sorry! I wonder why [Zipline doesn't require it](https://github.com/cashapp/zipline/pull/753).

JakeWharton:
Yep. Weird.

Anyway not your fault. Somehow the branch protection failed #567

PR #579, opened 2022-12-01
Break the `TreehouseView` dependency on `TreehouseApp`

JakeWharton:
This makes no attempt to clean up the `TreehouseView` API but instead contains the minimal changes to facilitate `TreehouseView`s being testable in isolation.

Refs #541


JakeWharton:
I kept failing at refactors which did everything at once so I'm going to do things much more incrementally.

swankjesse (approved):

swankjesse:
Ugh, I shoulda landed this first. That’s my fault.

JakeWharton:
Nah it's easy to rebase. If I didn't inline `onContentChanged` it probably wouldn't have conflicted.

PR #581, opened 2022-12-02
Redo widget binding lifecycles

swankjesse:
Goals of this change:

 * Do more stuff on the UI thread, and own more state on the UI thread. We used to own binding state on the Zipline thread, but that was clumsy 'cause widgets change more frequently than Zipline instances.
 * Stop holding references to widget views as quickly as possible

### DIFF
--- a/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
+++ b/redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
@@ -28,7 +28,6 @@ import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.yield
 import kotlinx.serialization.modules.EmptySerializersModule
 import kotlinx.serialization.modules.SerializersModule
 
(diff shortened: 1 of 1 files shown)


redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -16,18 +16,19 @@
...
 import app.cash.redwood.widget.Widget
 import app.cash.zipline.Zipline
-import app.cash.zipline.ZiplineService
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.yield
```
JakeWharton:
```suggestion
```

JakeWharton (approved):

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
swankjesse: Oooh what a great shame, being caught yielding! Because you demand to know what I was doing with that naughty device...

There’s a crash when doing a code swap in EmojiSearch. I think it might be just the 60hz frame pulse being enqueued and executing too late, though I haven’t really locked that down. I wanted to try yield to see if it was just a race in how different coroutines were sequenced during tear down. That wasn‘t it.

The real fix is for me to do the [shutdown thing](https://github.com/cashapp/zipline/issues/694) I’ve been sitting on for forever in Zipline. My first PR works but I’m not happy with how it works specifically around enqueued jobs. Time to do some hard design on what the requirements there should be.

JakeWharton: Hah I immediately did a search for its usage and was relieved when it came up empty.

PR #586, opened 2022-12-04
Attempt to use cinterop to override UIScrollView.

### LINKED ISSUE #489: Remove viewFactory from UIViewRedwoodLayoutWidgetFactory.
`UIViewRedwoodLayoutWidgetFactory` currently accepts a `viewFactory: RedwoodUIScrollViewFactory` param that creates `UIScrollView` instances. This is a work-around as it's not (easily) possible to extend `UIScrollView` from Kotlin multiplatform code so we delegate the creation to Swift code.

This is a pretty inconvenient work-around and we should figure out a way to remove the `viewFactory` param and create `UIScrollView`s without an external factory.

colinrtwhite:
Attempts to resolve: https://github.com/cashapp/redwood/issues/489

I think I made some progress with the `cinterop` solution and wanted to get @kpgalligan's eyes on it. The Kotlin compile tasks succeed, but then `linkDebugFrameworkIosArm64` fails:

```kotlin
> Task :redwood-layout-uiview:linkDebugFrameworkIosArm64 FAILED
e: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld invocation reported errors

The /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld command returned non-zero exit code: 1.
output:
Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_UIScrollViewWithInterop", referenced from:
      objc-class-ref in result.o
ld: symbol(s) not found for architecture arm64
```

Is it possible fix this or is it a dead end? This solution seems a bit cleaner than the `cklib` solution so I'm hoping to use `cinterop` if possible.


kpgalligan:
Posted a different draft PR (https://github.com/cashapp/redwood/pull/599).

About
```
Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_UIScrollViewWithInterop", referenced from:
      objc-class-ref in result.o
```

`_UIScrollViewWithInterop` is defined in the def file, but there's no implementation so the linker fails. I tried adding the implementation in the def file as well, but it didn't like it. There might be something else to do there, but I wound up going with cklib as that works. I tried some other tricks, but nothing worked out so far.

PR #588, opened 2022-12-05
Refactor FlexContainer to position its children as part of measure.

colinrtwhite:
Removes `FlexContainer.layout` and folds its responsibilities of updating the child's coordinates into `FlexContainer.measure`. This ensures that a child's position is always up-to-date with the parent's constraints.

This means you only need to call `container.measure(widthSpec, heightSpec)`, which returns the measured container size and updates the `left/right/top/bottom` of each child.


redwood-flex-container/src/commonMain/kotlin/app/cash/redwood/flexbox/FlexContainer.kt
```diff
@@ -81,13 +80,90 @@ public class FlexContainer {
    */
   public val items: MutableList<FlexItem> = mutableListOf()
 
+  //region measure
+
+  /**
+   * Measures the flexbox's items according to the parent's [widthSpec] and [heightSpec] and returns
+   * the expected size of the flexbox.
```
JakeWharton: Minor but I think we should stick to using "container" here instead of "flexbox"

JakeWharton (approved):

PR #589, opened 2022-12-05
Rollback action upgrade

JakeWharton:
The 3.7.0 release was deleted, because that's a thing you can do.

💀 https://github.com/actions/setup-java/issues/422


colinrtwhite (approved):
Wow! That's going to cause some issues.

PR #592, opened 2022-12-05
Migrate property tag to value class with signed storage

JakeWharton:
Split out these types from the actual serialized protocol from the sink interfaces while we're here.

Refs #418


redwood-cli/src/main/kotlin/app/cash/redwood/generator/types.kt
```diff
@@ -27,10 +27,11 @@ import com.squareup.kotlinpoet.UNIT
 
 internal object Protocol {
   val Event = ClassName("app.cash.redwood.protocol", "Event")
-  val EventTag = Event.nestedClass("Tag")
+  val EventTag = ClassName("app.cash.redwood.protocol", "EventTag")
```
JakeWharton: I may go back and make this change on #591 since I'll have to rebase it anyway once #590 squash merges.

colinrtwhite (approved):

JakeWharton:
I canceled the workflow because it was hung and it auto-merged. WTF.

PR #595, opened 2022-12-06
Move the FrameClock into the host code

swankjesse:
This introduces a new type, AppService, that acts as the base type 'A' for the TreehouseApp type parameter. It is the bridge type that spans the host and downloaded code.

It also introduces an interface, FrameClockService, that receives frames in downloaded code.

### DIFF
--- a/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppService.kt
+++ b/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppService.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Square, Inc.
+ * Copyright (C) 2022 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,12 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package app.cash.zipline.samples.emojisearch
+package app.cash.redwood.treehouse
 
-import app.cash.redwood.treehouse.AppService
-import app.cash.redwood.treehouse.ZiplineTreehouseUi
 import app.cash.zipline.ZiplineService
 
-interface EmojiSearchPresenter : AppService, ZiplineService {
-  fun launch(): ZiplineTreehouseUi
+/**
+ * Base interface for Treehouse applications. Each application should extend this interface to
+ * declare APIs that are declared by downloaded code and called from host code.
+ *
+ * Note that due to a Zipline limitation it's necessary for implementing classes to declare a direct
+ * dependency on [ZiplineService]. https://github.com/cashapp/zipline/issues/765
+ */
+public interface AppService : ZiplineService {
+  public val frameClockService: FrameClockService
+    get() = StandardFrameClockService
 }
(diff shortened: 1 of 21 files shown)


redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -87,6 +88,22 @@ public class TreehouseApp<A : Any> internal constructor(
     // This job lets us cancel the Android Treehouse job without canceling its sibling jobs.
     val supervisorJob = SupervisorJob(scope.coroutineContext.job)
     val cancelableScope = CoroutineScope(supervisorJob + dispatchers.zipline)
+
+    cancelableScope.launch(dispatchers.zipline) {
```
swankjesse, line 92: For follow-up: connect to a real vsync source

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppService.kt
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Base interface for Treehouse applications. Each application should extend this interface to
+ * declare APIs that are declared by downloaded code and called from host code.
+ *
+ * Note that due to a Zipline limitation it's necessary for implementing classes to declare a direct
+ * dependency on [ZiplineService]. https://github.com/cashapp/zipline/issues/765
+ */
+public interface AppService : ZiplineService {
+  public val frameClockService: FrameClockService
+    get() = StandardFrameClockService
```
swankjesse, line 29: I’m unsure what other - if any - services we’ll want to bridge automatically.

redwood-treehouse/src/jvmMain/kotlin/app/cash/redwood/treehouse/FrameClockServiceJvm.kt
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+// TODO(jwilson): deduplicate with identical symbol in hostMain
```
JakeWharton, line 18: Is something preventing this from working as-is?

JakeWharton (approved):

redwood-treehouse/src/jvmMain/kotlin/app/cash/redwood/treehouse/FrameClockServiceJvm.kt
swankjesse: It’s just about creating a `jniMain` source set.

JakeWharton: That's a zipline source set for sharing between Android and JVM. We don't need that. Seems like JVM is missing from the host source set. I didn't even realize we had JVM enabled for Treehouse. I guess it comes by accident.

swankjesse: Might be useful for testing? I’m not using it.

PR #599, opened 2022-12-06
Custom UIScrollView with cinterop

### LINKED ISSUE #489: Remove viewFactory from UIViewRedwoodLayoutWidgetFactory.

kpgalligan:
There's probably some renaming or moving to do here, but this PR includes a UIScrollView in Objc with a delegate, to avoid copy/pasting code.

Closes #489. Closes #586.


redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/RedwoodUIScrollViewFactory.kt
```diff
@@ -15,30 +15,17 @@
...
-  public fun setNeedsLayout()
-  public fun layoutSubviews()
-}
-
 /**
  * A factory that creates [UIScrollView] subclasses that delegate to a [RedwoodUIScrollViewDelegate].
  */
-public interface RedwoodUIScrollViewFactory {
-  public fun create(delegate: RedwoodUIScrollViewDelegate): UIScrollView
+public class RedwoodUIScrollViewFactory {
```
colinrtwhite: We can remove this too and allocate the `HostView` directly inside `UIViewFlexContainer`.

samples/emoji-search/ios/shared/src/commonMain/kotlin/app/cash/zipline/samples/emojisearch/exposed.kt
```diff
@@ -38,6 +39,7 @@ fun exposedTypes(
   treehouseUIKitView: TreehouseUIKitView<*>,
   uiViewRedwoodLayoutWidgetFactory: UIViewRedwoodLayoutWidgetFactory,
   uiViewRedwoodTreehouseLazyLayoutWidgetFactory: UIViewRedwoodTreehouseLazyLayoutWidgetFactory<*>,
+  redwoodUIScrollViewFactory: RedwoodUIScrollViewFactory,
```
colinrtwhite: I think we can remove this and inline the `HostView` creation in `UIViewFlexContainer`.

colinrtwhite (approved):
💯 Thanks for fixing this! The iOS folks are going to be stoked to remove the copy/pasted Swift code.

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/RedwoodUIScrollViewFactory.kt
kpgalligan: Updated

samples/emoji-search/ios/shared/src/commonMain/kotlin/app/cash/zipline/samples/emojisearch/exposed.kt
kpgalligan: Updated

colinrtwhite:
@kpgalligan Pushed a fix for the compilation issue [here](https://github.com/cashapp/redwood/commit/a381cd98fc57a5f0bdc4dd119e38a7a35f56373a), but when I run the emoji sample app it doesn't render correctly. It seems like `RedwoodScrollViewDelegateProtocol` produces `intrinsicContentSize` and `sizeThatFits` methods with a different signature that use a custom `app.cash.redwood.layout.uiview.cinterop.CGSize`.

kpgalligan:
I was having an issue running the sample locally anyway, so was flying blind a bit. The app ran, but wasn't getting local data properly. Will try to fix that and adjust.

kpgalligan:
Took a bit to get back to this @colinrtwhite. I don't think you need `app.cash.redwood.layout.uiview.cinterop.CGSize`. I got your branch, and replaced it with just `platform.CoreGraphics.CGSize` and it compiles for me.

I updated my branch with your changes to `EmojiSearch.kt`. Here's a screenshot with the output (and some vertical padding on the Column, because the search box was up by the notch).

Simulator Screen Shot - iPhone 13 - 2022-12-12 at 20 10 14

I just tried `import app.cash.redwood.layout.uiview.cinterop.CGSize` in my branch and it fails to compile.

Will need to dig into more tomorrow I guess.

kpgalligan:
> Pushed a fix for the compilation issue [here](https://github.com/cashapp/redwood/commit/a381cd98fc57a5f0bdc4dd119e38a7a35f56373a), but when I run the emoji sample app it doesn't render correctly. It seems like `RedwoodScrollViewDelegateProtocol` produces `intrinsicContentSize` and `sizeThatFits` methods with a different signature that use a custom `app.cash.redwood.layout.uiview.cinterop.CGSize`.

@colinrtwhite I tried your branch again and its the same result as mine. IDE and CLI build don't resolve `import app.cash.redwood.layout.uiview.cinterop.CGSize`. Not sure why you're seeing that, but I can't repro that in either branch.
Screen Shot 2022-12-15 at 4 37 37 PM

I don't want to just dismiss what you're seeing, though, so if you could, please try a clean build or whatever to see if we can figure that out.

PR #602, opened 2022-12-07
Create value class for layout modifier tag

### LINKED ISSUE #418: Make protocol values all value types

JakeWharton:
Closes #418

### DIFF
--- a/redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
+++ b/redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
@@ -15,11 +15,18 @@
  */
 package app.cash.redwood.protocol
 
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.KSerializer
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.builtins.serializer
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonNull
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.serializer
 
 @Serializable
 public data class Event(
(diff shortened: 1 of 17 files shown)


redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
```diff
@@ -50,13 +57,37 @@ public data class PropertyDiff(
...
-  /**
-   * Array of layout modifiers. Each element of this array is itself a two-element array of the
-   * layout modifier tag and then serialized value.
-   */
-  val elements: JsonArray,
+  val elements: List<LayoutModifierElement> = emptyList(),
 )
 
+@Serializable(with = LayoutModifierElementSerializer::class)
+public data class LayoutModifierElement(
```
swankjesse, line 64: Yay, better structure.

redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
```diff
@@ -50,13 +57,37 @@ public data class PropertyDiff(
...
+
+private object LayoutModifierElementSerializer : KSerializer<LayoutModifierElement> {
+  private val delegate = serializer<Array<JsonElement>>()
+
+  @OptIn(ExperimentalSerializationApi::class)
+  override val descriptor = SerialDescriptor("LayoutModifierElement", delegate.descriptor)
+
+  override fun deserialize(decoder: Decoder): LayoutModifierElement {
+    val decoded = decoder.decodeSerializableValue(delegate)
+    check(decoded.size in 1..2) { "TODO" }
```
swankjesse: lenient?

```
check(decoded.size >= 2)
```

swankjesse (approved):
I like pushing the compact encoding work to the serializer

redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
JakeWharton: 0 elements are also invalid. The tag is required.

But I'll add the message I forgot and maybe write a test for this logic, too.

JakeWharton:
Well I'm glad I wrote tests. This is trickier than it seems!

PR #604, opened 2022-12-07
Make `MeasureSpec` a value type

JakeWharton:
This value is forced to be non-negative which does give us one free bit to manipulate, but there are no other easy bits to steal from a floating-point representation. Instead, use the two least-significant bits from the fractional part. This will slightly alter values, but at precisions which are imperceptible to users.

You can play with the algorithm at https://pl.kotl.in/MjyOdx8-A


colinrtwhite (approved):
TIL about `toRawBits`. Awesome!

JakeWharton:
There's also a regular `toBits()`, but it includes logic to normalize `NaN` which we don't really need.

PR #611, opened 2022-12-09
Introduce tooling package/artifacts, move schema parser

JakeWharton:
Rename schema-annotations to just schema to match its package.


JakeWharton:
Going to pull the code generator and the lint support into 'tooling' artifacts.

swankjesse (approved):

PR #628, opened 2022-12-13
Add POC for swiftUI version of emoji-search

skorulis-ap:
This creates a version of emoji-search which is built on top of SwiftUI. It's currently as bare bones as possible to get something to render. The following have not been touched:

1. Handling layoutModifiers
2. Responding to WidgetRow and WidgetColumn parameters (waiting on enums to come through)

The code was based off the UIKit app with the rendering and binding parts replaced. Because SwiftUI isn't representable in Kotlin the implementations of common layout structures need to be done in Swift. This creates the biggest hurdle which is how to share this code between projects.

**How this works** 

Each Binding is implemented as a class, which has a number of `@Published` properties. When one of these changes SwiftUI will compute which parts of the view tree need to be rendered. Each binding also has a `view` parameter which defines how to render the binding. Classes get an automatic `id` property which allows swiftUI to understand the difference between an object being modified or added/removed.


samples/emoji-search/ios/swiftUI/EmojiSearchApp/Widget/ColumnBinding.swift
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import shared
+import SwiftUI
+
+final class ColumnBinding: BaseWidget, WidgetColumn, SwiftUIViewBinding {
+    func height(height: Int32) {
```
colinrtwhite, line 22: Oof yeah Kotlin value classes don't interop well with Swift unfortunately. On the JVM (Android) these let us define 0 overhead enums.

samples/emoji-search/ios/swiftUI/EmojiSearchApp/Widget/ColumnBinding.swift
```diff
@@ -0,0 +1,64 @@
...
+        
+    }
+
+    func horizontalAlignment(horizontalAlignment: Int32) {
```
colinrtwhite, line 26: @skorulis-ap The way we've been implementing these params to far is to pass them to a [layout engine](https://github.com/cashapp/redwood/blob/trunk/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt#L42) written in common Kotlin code. Basically it takes size constraints + views and returns positions for each of the children.

Is it possible to position children manually in a `VStack` or `HStack`? Else, we'll probably have to do something custom for Swift UI.

skorulis-ap: Manual positioning is possible, but kind of fights the system. That I think is less of a worry than getting the sizes of each child. SwiftUI calculates the sizes during the layout pass so you would need to have some sort of side rendering happening to get these values. 
iOS 16 adds a [Layout](https://developer.apple.com/documentation/swiftui/layout) object but that's not an option for a few years.

My plan was to use the SwiftUI layout system and respect the parameters that are passed in. wrapping likely being the hardest one to implement.

samples/emoji-search/ios/swiftUI/EmojiSearchApp/Widget/ColumnBinding.swift
underscoretang: I think we might want to start considering using primitives and language features that do have better inter-opt, since the framework is meant to target both platforms. are there alternatives that could be considered for some of these?

colinrtwhite: We could convert these properties to enums, though that has a small performance impact. Initially the plan was these interfaces would only be implemented in Kotlin.

JakeWharton:
Haven't forgot about this! Going to try to get it (and a version for the Counter sample) landed first thing next year.

Issue #621, opened 2022-12-12
Emoji Search Swift UI frontend


JakeWharton:
We have (finally) confirmed that Swift UI works and we should land a frontend for emoji search into this repo using Swift UI.

skorulis-ap:
https://github.com/cashapp/redwood/pull/628

Issue #620, opened 2022-12-12
Counter Swift UI frontend


JakeWharton:
We have (finally) confirmed that Swift UI works and we should land a frontend for counter into this repo using Swift UI.

PR #632, opened 2022-12-15
Prepare for multiple iOS frontends

JakeWharton:
Move the shared module and frontend to both be top level inside the sample folder and rename the frontend to have a 'uikit' suffix.

Refs #620. Refs #621.


samples/emoji-search/ios/README.md
```diff
@@ -1,3 +0,0 @@
-# iOS Emoji Search
-
-In `app/`, run `pod install` and then `open EmojiSearchApp.xcworkspace/`.
```
JakeWharton, line 3: This content is already in the sample's root README

colinrtwhite (approved):

PR #644, opened 2022-12-16
Expose only a minimal bridge type for diff-producing codegen

JakeWharton:
This reduces the generated API surface to something that hides 100% of the factory and widget types. Instead, a `ProtocolBridge` subtype is exposed with its companion acting as a factory.

### DIFF
--- a/redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolBridge.kt
+++ b/redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolBridge.kt
@@ -15,84 +15,44 @@
  */
 package app.cash.redwood.protocol.compose
 
-import app.cash.redwood.protocol.ChildrenDiff
 import app.cash.redwood.protocol.ChildrenTag
 import app.cash.redwood.protocol.Diff
 import app.cash.redwood.protocol.Event
 import app.cash.redwood.protocol.EventSink
 import app.cash.redwood.protocol.Id
-import app.cash.redwood.protocol.LayoutModifiers
-import app.cash.redwood.protocol.PropertyDiff
 import app.cash.redwood.widget.Widget
+import kotlinx.serialization.json.Json
 
-public class ProtocolBridge : EventSink {
-  private var nextValue = Id.Root.value + 1L
-  private val nodes = mutableMapOf<Id, DiffProducingWidget>()
-
-  private var childrenDiffs = mutableListOf<ChildrenDiff>()
-  private var layoutModifiers = mutableListOf<LayoutModifiers>()
-  private var propertyDiffs = mutableListOf<PropertyDiff>()
-
-  public fun nextId(): Id {
-    val value = nextValue
-    nextValue = value + 1L
-    return Id(value)
-  }
-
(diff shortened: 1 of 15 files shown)


JakeWharton:
This side is mostly boring but was easier. There's a symmetric change to the consuming side which is much more interesting as it should allow divorcing the creation of the diff-consuming wrapper (one per code load) from creation of the native widget factory (effectively a singleton) meaning we can eliminate the widget system concept / awkwardness.

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolBridge.kt
```diff
@@ -15,84 +15,44 @@
...
 import app.cash.redwood.protocol.Diff
 import app.cash.redwood.protocol.Event
 import app.cash.redwood.protocol.EventSink
 import app.cash.redwood.protocol.Id
-import app.cash.redwood.protocol.LayoutModifiers
-import app.cash.redwood.protocol.PropertyDiff
 import app.cash.redwood.widget.Widget
+import kotlinx.serialization.json.Json
 
-public class ProtocolBridge : EventSink {
```
JakeWharton, line 28: This class was renamed without changes to `ProtocolState`. Git can't track rename + new file with same name in a single commit.

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolBridge.kt
```diff
@@ -15,84 +15,44 @@
...
-  public fun append(propertyDiff: PropertyDiff) {
-    propertyDiffs += propertyDiff
-  }
-
+/**
+ * Exposes a [Widget.Children] and [Widget.Provider] whose changes can be captured as a [Diff] to
+ * send to a remote frontend. Incoming [Event]s can also be sent to this instance and will be routed
+ * to the appropriate handler.
+ */
+public interface ProtocolBridge : EventSink {
```
JakeWharton, line 31: This is an entirely-new type unrelated to the old one aside from sharing the `EventSink` + `createDiffOrNull` APIs.

samples/emoji-search/presenters/src/jsMain/kotlin/app/cash/zipline/samples/emojisearch/RealEmojiSearchPresenter.kt
```diff
@@ -43,15 +42,15 @@ class RealEmojiSearchPresenter(
 
   override fun launch(): ZiplineTreehouseUi {
     val events = MutableSharedFlow<EmojiSearchEvent>(extraBufferCapacity = Int.MAX_VALUE)
-    val factories = EmojiSearchDiffProducingWidgetFactories(ProtocolBridge(), json)
+    val bridge = EmojiSearchProtocolBridge.create(json)
```
JakeWharton, line 45: This also makes "diff-producing" a thing that is no longer in public APIs which is a nice win and that I'm eager to see for "diff-consuming".

colinrtwhite (approved):

PR #645, opened 2022-12-16
Make hosting a Redwood composition inside Compose UI simpler

JakeWharton:


JakeWharton:
Helpful for the Desktop Compose UI samples I'm toying with...

colinrtwhite (approved):

Issue #221, opened 2022-03-02
Remove the need to annotate schema if you are not using protocol


JakeWharton:
Things are never serialized so you shouldn't need to specify all this nonsense.

JakeWharton:
Related to #14

JakeWharton:
I think this means we need a separate `@Tag` annotation because we still need the signal from `@Widget`/`@LayoutModifier`/`@Property`/`@Children`.

PR #650, opened 2022-12-19
Split the parse API into protocol and non-protocol

JakeWharton:
Refs #221


JakeWharton:
After this we can start splitting the parsing codepaths so that the protocol gets mixed in rather than is required.

colinrtwhite (approved):

PR #653, opened 2022-12-21
Fix folder casing to match declaration in build.gradle

JakeWharton:


JakeWharton:
`trunk` is broken on case-sensitive file systems.

colinrtwhite (approved):

PR #679, opened 2022-12-22
Compile with and target API 33

JakeWharton:
Needed for some dependency bumps.


JakeWharton:
Needs paparazzi release :(

PR #688, opened 2022-12-23
Update release process

JakeWharton:
- Do not do a full build. Just publish and trust things are good.
- Run in parallel. Latest publish plugin uses a build service to prevent parallelism on upload tasks and also creates an explicit staging repo to prevent Sonatype repo split.
- Fix Dokka task, and also run that in parallel.
- Match RAM provisioning of regular build.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,7 +6,7 @@ on:
       - '**'
 
 env:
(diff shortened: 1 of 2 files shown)


JakeWharton:
Probably should run Dokka before creating the GitHub release

colinrtwhite (approved):

PR #689, opened 2022-12-23
Fix site source folder

JakeWharton:
The source folder is 'site/' for 'trunk' builds because we run Dokka on one shard and then download the HTML into 'site/'. When deploying from the build shard we need to use original folder.


JakeWharton:
YOLOing this in because I want to start the release

PR #707, opened 2023-01-03
Unify the View Row/Column implementations

JakeWharton:
They're 99% the same. That also means we can fold in the helper.


colinrtwhite (approved):
Nice!

PR #711, opened 2023-01-04
Eliminate 'DiffProducing' prefix

JakeWharton:
We now only have 'Protocol' which exists as Widgets (this change) or Nodes (future change to eliminate 'Diff Consuming').

Also cleaned up some leftover 'Sunspot' references and incorrect local names.

### DIFF
--- a/redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/ComposeProtocolGenerationTest.kt
+++ b/redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/ComposeProtocolGenerationTest.kt
@@ -22,7 +22,7 @@ import app.cash.redwood.tooling.schema.parseProtocolSchema
 import com.google.common.truth.Truth.assertThat
 import org.junit.Test
 
-class DiffProducingGenerationTest {
+class ComposeProtocolGenerationTest {
   @Schema(
     [
       IdPropertyNameCollisionNode::class,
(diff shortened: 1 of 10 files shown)


colinrtwhite (approved):

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/ComposeProtocolGenerationTest.kt
```diff
@@ -39,21 +39,19 @@ class DiffProducingGenerationTest {
   @Test fun `id property does not collide`() {
```
JakeWharton, line 39: Really should be a functional test...

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/ComposeProtocolGenerationTest.kt
```diff
@@ -39,21 +39,19 @@ class DiffProducingGenerationTest {
...
       |  public override fun id(id: String): Unit {
-      |    bridge.append(PropertyDiff(this.id, PropertyTag(2), json.encodeToJsonElement(serializer_0, id)))
-      |  }
-      |
+      |    this.state.append(PropertyDiff(this.id,
       """.trimMargin(),
     )
   }
 
   @Test fun `dependency layout modifiers are included in serialization`() {
```
JakeWharton, line 51: Could also be functional test...

PR #713, opened 2023-01-05
Generate less serialization code

JakeWharton:
No more empty object to hold a function which just returns a `LayoutModifierElement` with an integer. That is inlined to the callsite.

No more secondary constructor to unwrap the real `LayoutModifier` into its component values which is only called from a companion function. That unpacking is done directly in the companion function now.


JakeWharton:
Use [`?w=1`](https://github.com/cashapp/redwood/pull/713/files?w=1) to see the actual changes without indent change noise.

JakeWharton:
In Emoji Search, for the protocol serialization code only:

Before: 63450 bytes
After: 62040 bytes

Savings of 2.2%.

colinrtwhite (approved):

PR #718, opened 2023-01-09
Create a CoroutineScope for each bound TreehouseView

swankjesse:


redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -87,11 +87,10 @@ public class TreehouseApp<A : AppService> internal constructor(
     dispatchers.checkZipline()
     check(!closed)
 
-    // This job lets us cancel the Android Treehouse job without canceling its sibling jobs.
-    val supervisorJob = SupervisorJob(scope.coroutineContext.job)
-    val cancelableScope = CoroutineScope(supervisorJob + dispatchers.zipline)
-
-    cancelableScope.launch(dispatchers.zipline) {
+    val sessionScope = CoroutineScope(
+      SupervisorJob(scope.coroutineContext.job) + dispatchers.zipline
```
JakeWharton: If you put the dispatcher into the context then you don't need to specify it below anywhere that it matches.

JakeWharton (approved):

redwood-treehouse/src/hostMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
swankjesse: Yeah, I’d considered that. I also really like how obvious it is what’s running on the Zipline dispatcher vs. the UI dispatcher as-is, and I want to keep that.

One thing I considered was removing the dispatcher as a parameter from the scope constructor. Any consequences of that to be worried about?

JakeWharton: Launches without dispatchers would run on the default dispatcher, but the runtime checks should catch it.

swankjesse: Good call. Done.

PR #720, opened 2023-01-10
Compile with and target API 33

JakeWharton:
Needed for some dependency bumps.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -129,14 +129,14 @@ allprojects {
 
   plugins.withType(com.android.build.gradle.BasePlugin).configureEach {
     project.android {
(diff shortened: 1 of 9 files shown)


JakeWharton:
~Going to rebase onto `trunk` manually so I can retain the separate commit for future revert of the plugin switch (if Paparazzi ever releases).~ Not needed now that it's only a dependency coordinate change.

JakeWharton:
~Waiting on an alpha04 which allows retaining the use of the `app.cash.paparazzi` plugin ID.~ Done.

veyndan (approved):

PR #715, opened 2023-01-06
Generate one-way serializers for Compose protocol

JakeWharton:
This:
- Eliminates allocation of surrogate instance
- Dramatically reduces generated code by only including serialization half

For the Emoji Search JS module dedicated to Compose protocol serialization:

    62040  before.js
    44221  after.js

A 29% reduction in size.

Closes #714, but more size savings to come!

PR #728, opened 2023-01-12
Ensure we fall back to regular serializers if available

JakeWharton:
For `@Serializable` types we can fall back to their generated serializer if none is registered with the module.

This mimics the old behavior of the generated surrogate serializers.


veyndan (approved):
Awesome! Double checked and https://github.com/cashapp/redwood/pull/715#issuecomment-1380514574 is fixed with this change

PR #746, opened 2023-01-23
De-nest RecyclerView from ScrollView

veyndan:
When running Emoji Search with the Android View target, the performance is awful.

`onCreateViewHolder` was being invoked for **every** item in the list, regardless of whether it was in the viewport or not. This is because the `RecyclerView` height contained all the items, and was not bounded by the viewport. I narrowed this down to the fact the `RecyclerView` was being nested in a `ScrollView`.

### DIFF
--- a/redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
+++ b/redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
@@ -17,9 +17,9 @@ package app.cash.redwood.layout.view
 
 import android.annotation.SuppressLint
 import android.content.Context
-import android.view.MotionEvent
 import android.view.View
 import android.view.ViewGroup
+import android.widget.FrameLayout
 import android.widget.HorizontalScrollView
 import androidx.core.widget.NestedScrollView
 import app.cash.redwood.LayoutModifier
@@ -49,7 +49,9 @@ internal class ViewFlexContainer(
   private val density = DensityMultiplier * context.resources.displayMetrics.density
 
   private val hostView = HostView(context)
-  override val value = newScrollView()
+
+  private val _value = Container()
+  override val value: View = _value
 
   override val children = ViewGroupChildren(hostView)
 
@@ -73,8 +75,9 @@ internal class ViewFlexContainer(
   }
 
   override fun overflow(overflow: Overflow) {
+    val oldScrollEnabled = scrollEnabled
     scrollEnabled = overflow == Overflow.Scroll
-    invalidate()
+    invalidate(overflowChanged = oldScrollEnabled != scrollEnabled)
   }
 
   override fun horizontalAlignment(horizontalAlignment: MainAxisAlignment) {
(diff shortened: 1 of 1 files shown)


JakeWharton (approved):
Should we be using `NestedScrollView` for the scrolling case, too?

veyndan:
I'm not sure I understand. We [already are using](https://github.com/cashapp/redwood/blob/fcbb979bab9813892321ec7f0c80e73073c034a9/redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt#L127) `NestedScrollView` when `scrollEnabled` is `true`.

swankjesse (approved):

JakeWharton:
Ah, nice. I assumed the lack of performance was from not using such a type.

colinrtwhite:
I don't think we can change the `value` depending on scroll state unfortunately. If the `Overflow` property changes after the initial composition it won't re-call `value get()` to create a new view.

JakeWharton:
Yeah so the simple way to do this would be to return a FrameLayout whose sole child switches between a HostView or a NestedScrollView+HostView.

A more complicated version is to always return a HostView and in the case of no scrolling add children directly. But in the case of scrolling add a NSV at full size and another HostView inside of that. This basically eliminates one level of indirection in the (probably common) non-scrolling case.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
```diff
@@ -139,6 +141,11 @@ internal class ViewFlexContainer(
     }
   }
 
+  private fun newLinearLayout() = LinearLayout(context).apply {
+    orientation = if (direction.isHorizontal) LinearLayout.HORIZONTAL else LinearLayout.VERTICAL
```
JakeWharton: We only need orientation switching for scrolling. This could be a FrameLayout for the non-scrolling case as the HostView and FlexContainer take care of the orientation otherwise.

JakeWharton (commented):
Yes. We could always use a FrameLayout and conditionally use a NSV as its sole child, though.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
veyndan: I tried both replacing the `LinearLayout` with a `FrameLayout` and de-nesting completely to just return a `hostView`, but when doing that the views overlap each other.

Here's what it looks like when replacing [these lines](https://github.com/cashapp/redwood/blob/967b6f4ec2a531c5310b904e997cd8e12e1c9e84/samples/emoji-search/presenter/src/commonMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearch.kt#L105-L116) with `repeat(100) { Text("Hello") }` using either a `FrameLayout` or de-nesting completely:

veyndan:
> Yeah so the simple way to do this would be to return a FrameLayout whose sole child switches between a HostView or a NestedScrollView+HostView.
> 
> A more complicated version is to always return a HostView and in the case of no scrolling add children directly. But in the case of scrolling add a NSV at full size and another HostView inside of that. This basically eliminates one level of indirection in the (probably common) non-scrolling case.

@JakeWharton @colinrtwhite I just updated the PR using the simple way (for the time being). With these changes, I'm now getting a UI bug with Emoji Search, where the Search Bar is being cut off by the list. I don't know enough about how flex calculates the layout to say where this bug lies.

colinrtwhite (approved):

colinrtwhite:
@veyndan The view's size is determined by the `Size` returned by `container.measure(widthSpec, heightSpec)`. I'd check if that's outputting half size. I can take a look soon as well.

veyndan:
@colinrtwhite Thanks for the pointer. I tried debugging it, and I'm coming across some obscure bugs that don't exist on `trunk`.

For example, replacing:

https://github.com/cashapp/redwood/blob/10e82f08a83a654265acb914ce183b65af6f1222/samples/emoji-search/presenter/src/commonMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearch.kt#L105

with

```kotlin
    filteredEmojis.take(numberOfItems).forEach { image ->
```

1. If you set `numberOfItems` to a low number, where the number of items will all fit on the screen, then it all looks good.
2. If you set `numberOfItems` to a higher number, where the number of items won't all fit on the screen, then the width of **every** item is set to `48` (I'm assuming the content width is `0`, and `48` comes from twice the padding, i.e., `24 x 2`).

veyndan:
Filed the above bugs in https://github.com/cashapp/redwood/issues/770 as I think it's outside the scope of this PR.

PR #754, opened 2023-01-27
Use platform-specific collections inside Compose protocol

JakeWharton:
These collections are implementation details and are used at some of the lowest layers of the app. They should be as efficient as possible.

Before JS:
```js
function ProtocolState() {
  this.a3h_1 = _Id___get_value__impl__b6l2oq(Companion_getInstance().r3d_1) + 1 | 0;
  this.b3h_1 = LinkedHashMap_init_$Create$();
  this.c3h_1 = ArrayList_init_$Create$();
  this.d3h_1 = ArrayList_init_$Create$();
  this.e3h_1 = ArrayList_init_$Create$();
  this.f3h_1 = false;
  this.g3h_1 = 8;
}

ProtocolState.prototype.i3h = function (childrenDiff) {
  this.c3h_1.a(childrenDiff);
  this.f3h_1 = true;
};

ProtocolState.prototype.u3g = function () {
  if (!this.f3h_1)
    return null;
  var diff = new Diff(this.c3h_1, this.d3h_1, this.e3h_1);
  this.c3h_1 = ArrayList_init_$Create$();
  this.d3h_1 = ArrayList_init_$Create$();
  this.e3h_1 = ArrayList_init_$Create$();
  return diff;
};

ProtocolState.prototype.o3h = function (id) {
  return this.b3h_1.t2(new Id(id));
};
```

After JS:
```js
function ProtocolState() {
  this.a3h_1 = _Id___get_value__impl__b6l2oq(Companion_getInstance().r3d_1) + 1 | 0;
  this.b3h_1 = new Map();
  this.c3h_1 = new Array();
  this.d3h_1 = new Array();
  this.e3h_1 = new Array();
  this.f3h_1 = false;
  this.g3h_1 = 8;
}

ProtocolState.prototype.i3h = function (childrenDiff) {
  this.c3h_1.push(childrenDiff);
  this.f3h_1 = true;
};

ProtocolState.prototype.u3g = function () {
  if (!this.f3h_1)
    return null;
  var tmp = new JsArrayList(this.c3h_1);
  var tmp_0 = new JsArrayList(this.d3h_1);
  var tmp$ret$2 = new JsArrayList(this.e3h_1);
  var diff = new Diff(tmp, tmp_0, tmp$ret$2);
  this.c3h_1 = new Array();
  this.d3h_1 = new Array();
  this.e3h_1 = new Array();
  return diff;
};

ProtocolState.prototype.o3h = function (id) {
  return this.b3h_1.get(_Id___get_value__impl__b6l2oq(id));
};
```

In terms of code size it's a lateral move at best, but by using the raw underlying collections we are not subject to layers of wrappers and indirection for emulating the JVM collection semantics that Kotlin's collection have unfortunately inherited. Additionally, we explicitly unbox `Id` before it goes into our map to avoid an actual boxing operation that is not needed in JS. On the JVM the Int will still box, but that's fine since it always boxed `Id` before. When Babel ultimately runs over this code functions like `_Id___get_value__impl__b6l2oq` will disappear completely as it is an identity function.

### DIFF
--- a/redwood-protocol-compose/build.gradle
+++ b/redwood-protocol-compose/build.gradle
@@ -25,6 +25,18 @@ kotlin {
         implementation projects.testSchema.compose.protocol
       }
     }
+    nonJsMain {
+      dependsOn(commonMain)
+    }
+    androidMain {
+      dependsOn(nonJsMain)
+    }
+    jvmMain {
+      dependsOn(nonJsMain)
+    }
+    nativeMain {
+      dependsOn(nonJsMain)
+    }
   }
 }
 
--- a/redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/PlatformList.kt
+++ b/redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/PlatformList.kt
@@ -1,0 +1,22 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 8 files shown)


veyndan (approved):
What's the reasoning in not just depending on https://github.com/JakeWharton/platform-collections?

JakeWharton:
It has no released versions (yet), the number of APIs we need is extremely small, and we really only care about specializing JS. Someday it might be worth switching over, such as if we start needing these across multiple modules.

Issue #335, opened 2022-08-30
Restore the value-based testing tree that was removed


JakeWharton:
For presenter-like testing of the UI. Previously we tried to re-use the schema types as the value types but this proved much too difficult. Instead, we're going to generate a widget factory whose product is a value type that we can assert against easily.

This probably also entails generating tree-like navigation helpers to assert on only parts of the tree. Plus these will aid in avoiding asserting instance equality of lambdas which are held for events.

PR #767, opened 2023-01-30
Initial implementation of values-based testing tree

swankjesse:
https://github.com/cashapp/redwood/issues/335


swankjesse:
@JakeWharton I’d like you to take a look at this before I invest effort doing real code gen. . . .

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/MutableWidget.kt
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package com.example.redwood.emojisearch.presenter
+
+import app.cash.redwood.RedwoodCodegenApi
+import app.cash.redwood.widget.Widget
+
+@RedwoodCodegenApi
+interface MutableWidget : Widget<MutableWidget> {
```
swankjesse: I ‘generated’ two classes for each widget, like `MutableTextInput` and `TextInputValue`. The mutable one is implementation-detail and hidden from the public API with `@RedwoodCodegenApi`. The value one is almost a data class, but excluding lambas from equals/hashCode/toString. This whole dual hierarchy is necessary 'cause Kotlin doesn’t have lenses.

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearchWidgets.kt
```diff
@@ -0,0 +1,186 @@
...
+
+@RedwoodCodegenApi
+private class MutableTextInput : TextInput<MutableWidget>, MutableWidget {
+  override val value: MutableWidget
+    get() = this
+
+  override var layoutModifiers: LayoutModifier = LayoutModifier
+
+  var state: TextFieldState? = null
+  var hint: String? = null
```
swankjesse, line 67: I didn’t differentiate between never-set and null, for properties that can be null. I think that’s fine.

samples/emoji-search/presenter/build.gradle
```diff
@@ -5,12 +5,19 @@ kotlin {
...
+  jvm {
+  }
 
   sourceSets {
     commonMain {
       dependencies {
         api projects.samples.emojiSearch.schema.compose
       }
     }
+    commonTest {
```
swankjesse, line 17: Everything’s in one module for now. I think I should explode this into a few modules:

 * `redwood-testing`: RedwoodTester, MutableWidget, WidgetValue, ManualCoroutineDispatcher
 * `samples/emoji-search/widget/values`: the generated code currently in `EmojiSearchWidgets.kt`. This will depend on `redwood-testing`.

We could further split testing stuff from values stuff. I’d like a use case for values-without-testing first.

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearchTest.kt
```diff
@@ -0,0 +1,89 @@
...
+      listOf(
+        TextInputValue(
+          state = TextFieldState(text = ""),
+          hint = "Search",
+        ),
+      ),
+      snapshot0,
+    )
+
+    (snapshot0.single() as TextInputValue).onChange!!.invoke(TextFieldState(text = "tree"))
```
swankjesse: I’d like something to get a flattened hierarchy. This would be useful with functions like `filterIsInstance()`:

```
val widgets = tester.snapshot()
widgets.flattened().filterIsInstance<TextInputValue>().single().onChange!!.invoke(...)
```

(`WidgetValue` doesn’t offer a function to get its children. I’ll need to add that to implement `flattened()`.)

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/ManualCoroutineDispatcher.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.example.redwood.emojisearch.presenter
+
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Runnable
+
+/** Extremely basic dispatcher that requires [executeQueuedJobs] to be invoked manually. */
+internal class ManualCoroutineDispatcher : CoroutineDispatcher() {
```
swankjesse: The `CoroutineDispatcher` I’m using here is super naive and likely to cause problems. Is there something built-in that works similarly?

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearchWidgets.kt
```diff
@@ -0,0 +1,186 @@
...
+
+  override fun hint(hint: String) {
+    this.hint = hint
+  }
+
+  override fun onChange(onChange: ((TextFieldState) -> Unit)?) {
+    this.onChange = onChange
+  }
+
+  override fun snapshot() = TextInputValue(layoutModifiers, state, hint, onChange)
```
swankjesse, line 82: I could cache snapshot objects. This might need a way for children to know their parent widgets so they could invalidate 'em. Unclear whether this juice is worth the squeeze.

samples/emoji-search/presenter/build.gradle
JakeWharton: I would put it at `redwood-compose-testing` since it's specifically a widget backend for the compose stuff (and the compose modules already depend on the widget modules).

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearchWidgets.kt
JakeWharton: Because the source of these properties is function arguments from the composable, there's no such thing as never-set properties.

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearchWidgets.kt
JakeWharton: Yeah I'm not sure that offers anything we need. This is exactly what the old code did. Plus it's all test-specific.

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/ManualCoroutineDispatcher.kt
JakeWharton: Yeah you don't want to be in the dispatcher business. This one has non-determinism with anyone that uses libraries which run on other dispatchers and it infinite loops on self-updating effects. I would delete this and use a normal one.

My intuition is that upon a snapshot, the best thing to do is tick the clock until the applier signals that it has new changes. And then when you get that snapshot, to ensure it actually differs from the previous one (and if not keep sending frames). Do this until a timeout after, say, 2 seconds, or 10.

This ensures that each call to snapshot produces a new, distinct snapshot which reflects some change. Beyond that, though, you still likely will have some non-determinism unless you rigorously control all of the dispatchers used by libraries.

I think a better user API is to completely ignore the concept of frames, and instead focus on the desired conditions and performing actions. This is in line with how a user would behave on the real display. 

For example:
- Wait for "email" input field to appear
- Enter email
- Enter password
- Click submit button
- Wait for payment pad to appear

Behind the scenes this will feature the same automatic frame clock ticking and waiting for the applier to signal changes have occurred, but then there's an additional step of ensuring that it's a snapshot which meets the requirements of the caller. And if we can't find one within a reasonable timeout the test fails with the latest snapshot and the missing condition.

This is the pattern we are encouraging with regular UI tests and it's why we are deprecating RxIdler and okhttp-idling-resource. You simply can't get reliable determinism by trying to monitor running work. And while you're testing Compose logic and value types, fundamentally is _is_ the same as a UI test since we normally project those values directly into platform-specific UI.

swankjesse: One nice advantage we have here is that there’s no concurrency unless we introduce it.

But I take your point that we don’t want to be in the dispatcher business, not at all. I think I’ll just punt that to the user, and make them inject whatever dispatcher they want.

samples/emoji-search/presenter/build.gradle
swankjesse: Can do.

swankjesse: Done.

swankjesse: I still need to move out the generated code...

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/ManualCoroutineDispatcher.kt
swankjesse: Alright, I’ve added code that lets callers await a distinct snapshot.

I decided to introduce a `ChangeTracker` thing to signal when changes were applied.

The only thing I’m worried about is tearing, where a snapshot is observed mid-frame.  I don‘t think this can occur in practice, since that coroutine would need to suspend mid-frame. Instead I suspect the `Semaphore` I’m using won’t resume until the composition is idle.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/RedwoodTester.kt
```diff
@@ -0,0 +1,108 @@
...
+  private val clock = BroadcastFrameClock()
+  private var timeNanos = 0L
+  private val frameDelay = 1.seconds / 60
+
+  /** Top-level children of the composition. */
+  private val mutableChildren = mutableListOf<Widget<MutableWidget>>()
+
+  private val composition = RedwoodComposition(
+    scope = scope + clock,
+    container = MutableListChildren(mutableChildren),
```
JakeWharton:
```suggestion
  private val mutableChildren = MutableListChildren<Widget<MutableWidget>>()

  private val composition = RedwoodComposition(
    scope = scope + clock,
    container = mutableChildren,
```

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/ManualCoroutineDispatcher.kt
JakeWharton: > I decided to introduce a `ChangeTracker` thing to signal when changes were applied.
>
> The only thing I’m worried about is tearing, where a snapshot is observed mid-frame. I don‘t think this can occur in practice, since that coroutine would need to suspend mid-frame. Instead I suspect the `Semaphore` I’m using won’t resume until the composition is idle.

The `WidgetApplier` already has a callback to tell you when changes have been applied.

Here's how the protocol-based widgets work:
```kotlin
val applier = WidgetApplier(bridge.provider, bridge.root) {
  bridge.createDiffOrNull()?.let(diffSink::sendDiff)
}
val composition = RedwoodComposition(scope, applier)
```

> One nice advantage we have here is that there’s no concurrency unless we introduce it.

We tried this with Molecule and it failed miserably. We can try it again if you want, but I think once you move away from testing samples to testing real-world code it becomes much more challenging to accomplish this.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/RedwoodTester.kt
```diff
@@ -0,0 +1,108 @@
...
+  public fun setContent(content: @Composable () -> Unit) {
+    composition.setContent(content)
+  }
+
+  /**
+   * Returns a snapshot, waiting if necessary for changes to occur since the previous snapshot.
+   *
+   * @throws TimeoutCancellationException if no new snapshot is produced before [timeoutMillis].
+   */
+  public suspend fun awaitSnapshot(timeoutMillis: Long = 1_000): List<WidgetValue> {
```
JakeWharton:
```suggestion
  public suspend fun awaitSnapshot(timeout: Duration = 1.seconds): List<WidgetValue> {
```

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/RedwoodTester.kt
```diff
@@ -0,0 +1,108 @@
...
+      changeTracker.changes.acquire()
+      sendFramesJob.cancel()
+    }
+
+    val snapshot = mutableChildren.map { it.value.snapshot() }
+
+    // Consume any extra changes on the returned snapshot.
+    changeTracker.changes.acquireAll()
+
+    return snapshot
```
JakeWharton: With the `WidgetApplier` callback and a `Channel` this should become pretty simple
```suggestion
    withTimeout(timeout) {
      val sendFramesJob = sendFrames()
      snapshot = snapshots.receive()
      sendFramesJob.cancel()
      return snapshot
    }
```

samples/emoji-search/presenter/src/commonTest/kotlin/com/example/redwood/emojisearch/presenter/ManualCoroutineDispatcher.kt
swankjesse: Ooooh I was looking for the change signalling API and couldn’t find it. This is much better, especially 'cause I get one batch of changes.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/RedwoodTester.kt
swankjesse: Yes, that’s exactly what I need. Now the mutable widgets are exclusively accessed by the composition.

swankjesse:
This is ready for review; I’ll do the codegen in a follow-up.

JakeWharton (approved):

Issue #605, opened 2022-12-08
`LazyColumn` usage outside of Treehouse


JakeWharton:
Can we break this dependency somehow such that it can be used with or without Treehouse?

PR #772, opened 2023-01-31
Add placeholder support for LazyColumn

### LINKED ISSUE #771: Add placeholder support to `LazyColumn`
For the host to receive an item for a given index, it must [request the item from the guest](https://github.com/cashapp/redwood/blob/174657a67d91945472576a4464f55beca91a7f4b/redwood-treehouse-lazylayout-api/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/api/properties.kt#L29-L31). While the host is waiting for a response from the guest, the host [**must** display something](https://developer.android.com/jetpack/compose/lists#avoid-0-size-items).

We should have an additional argument to `LazyColumn` that accepts a `Composable` for display during the interim. It's usage would look something like:

```kotlin
LazyColumn(
  placeholder = { ProgressIndicator() },
) {
  items(…) { … }
}
```

veyndan:
Fixes #771.

- [x] Compose UI support
- [ ] Android View support (BLOCKED)
- [ ] UIKit support

### DIFF
--- a/redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyColumn.kt
+++ b/redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyColumn.kt
@@ -59,14 +59,12 @@ internal class ComposeUiLazyColumn<A : AppService>(
           if (isPlaceholderVisible) {
             placeholder.render()
           }
-          val codeListener by remember {
-            mutableStateOf(
-              object : TreehouseView.CodeListener() {
-                override fun onCodeLoaded(initial: Boolean) {
-                  isPlaceholderVisible = false
-                }
-              },
-            )
+          val codeListener = remember {
+            object : TreehouseView.CodeListener() {
+              override fun onCodeLoaded(initial: Boolean) {
+                isPlaceholderVisible = false
+              }
+            }
           }
           TreehouseContent(treehouseApp, widgetSystem, codeListener) { interval.itemProvider.get(index) }
         }
--- a/redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyColumn.kt
+++ b/redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyColumn.kt
@@ -15,17 +15,15 @@
  */
 package app.cash.redwood.treehouse.lazylayout.composeui
 
-import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.unit.dp
 import app.cash.redwood.LayoutModifier
 import app.cash.redwood.treehouse.AppService
 import app.cash.redwood.treehouse.TreehouseApp
@@ -33,6 +31,7 @@ import app.cash.redwood.treehouse.TreehouseView
 import app.cash.redwood.treehouse.composeui.TreehouseContent
 import app.cash.redwood.treehouse.lazylayout.api.LazyListIntervalContent
 import app.cash.redwood.treehouse.lazylayout.widget.LazyColumn
+import app.cash.redwood.widget.compose.ComposeWidgetChildren
 
 internal class ComposeUiLazyColumn<A : AppService>(
   treehouseApp: TreehouseApp<A>,
@@ -41,6 +40,8 @@ internal class ComposeUiLazyColumn<A : AppService>(
   private var intervals by mutableStateOf<List<LazyListIntervalContent>>(emptyList())
 
   override var layoutModifiers: LayoutModifier = LayoutModifier
+
+  override val placeholder = ComposeWidgetChildren()
 
   override fun intervals(intervals: List<LazyListIntervalContent>) {
     this.intervals = intervals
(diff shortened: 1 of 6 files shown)


redwood-treehouse-lazylayout-view/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/view/ViewLazyColumn.kt
```diff
@@ -77,12 +76,12 @@ internal class ViewLazyColumn<A : AppService>(
...
     private val widgetSystem: TreehouseView.WidgetSystem<A>,
-    private val contentHeight: Int,
+    private val placeholder: List<Widget<View>>,
   ) : ListAdapter<LazyContentItem, ViewHolder<A>>(LazyContentItemDiffCallback) {
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder<A> {
-      val container = FrameLayout(parent.context).apply {
-        layoutParams = ViewGroup.LayoutParams(MATCH_PARENT, contentHeight)
-      }
+      val container = FrameLayout(parent.context)
+      // TODO Clone placeholder.
```
veyndan, line 83: Cloning View's isn't possible. Unsure how to proceed.

redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyColumn.kt
```diff
@@ -54,9 +55,20 @@ internal class ComposeUiLazyColumn<A : AppService>(
...
+          }
+          val codeListener by remember {
+            mutableStateOf(
+              object : TreehouseView.CodeListener() {
+                override fun onCodeLoaded(initial: Boolean) {
+                  isPlaceholderVisible = false
+                }
+              },
+            )
           }
```
hoc081098: I think mutablestate is unnecessary 
```suggestion
          val codeListener = remember {
           
              object : TreehouseView.CodeListener() {
                override fun onCodeLoaded(initial: Boolean) {
                  isPlaceholderVisible = false
                }
              },
            
           }
```

redwood-treehouse-lazylayout-view/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/view/ViewLazyColumn.kt
JakeWharton: Yeah this is an interesting use case. You're basically looking to have a single subset of a composition drive multiple identical trees on the display-side. It won't work for any traditional UI toolkit since even if you could clone them you are disconnecting them from the composition so they wouldn't see updates. That's okay in this specific case since your use is static, but in general we can't assume that.

I need to think about this.

JakeWharton: I don't see any easy path forward to achieving this, and I think that is partly due to the current design's forced usage of Treehouse for individual rows.

If LazyColumn were solely a Redwood-based widget, then the usage of Treehouse for individual rows would mean we could use normal state-handling mechanisms for this.
```kotlin
TreehouseLazyColumn(
  placeholder: @Composable () -> Unit,
  ..
) {
  LazyColumn(..) {
    items(..) {
      var showPlaceholder by remember { mutableStateOf(false) }
      val callback = remember { object : Callback {
        override fun onCodeLoaded(..) { showPlaceholder = false }
      }}
      if (showPlaceholder) {
        placeholder()
      } else {
        TreehouseContent(..)
      }
    }
  }
}
```
This gives you the composable-lambda-is-a-template functionality but its templating still happens within the Redwood composition–not on the display.

On the display side we would use a `RecyclerView` of `FrameLayout`s and the children of each item would render at each position the same it does today if that conditional was your root view. And equivalent things for other platforms/toolkits.

Non-Treehouse `LazyColumn` is tracked by #605 as it was blocking Emoji Search from working on web. We sorta hacked around it with some inversion of control for now. But it feels like a lazy row/column are fundamentals in the same vein as normal row/column that we aught to provide at the Redwood layer.

redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyColumn.kt
veyndan: Good catch!

redwood-treehouse-lazylayout-view/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/view/ViewLazyColumn.kt
veyndan: It'd definitely be preferable if `LazyColumn` were a Redwood-based widget as you said. I don't think I'm fully grasping the implementation though:

1. Would the above snippet be within `TreehouseLazyColumn.kt`? Because if so, idk how we'd access `TreehouseContent` considering it exists within the host.
2. In your proposal, if we have `items(count = 1000) { block(it) }`, would `block(it)` be invoked 1000 times as soon as the `LazyColum` is displayed on screen?

JakeWharton: `TreehouseContent` is a hypothetical Treehouse-equivalent of `RedwoodContent` which is just something that simplifies hosting Treehouse inside of an existing composition. It's basically what is inside of today's `LazyColumn`. We also might not even need this, which is where your next question is going.

If you did `items(count = 1000)` it would not create all 1000, but instead have a policy which would attempt to start filling the viewport + some amount off-screen (maybe a second viewport's worth). As the widget scrolls, it communicates with the composition to fill the upcoming items. So you could imagine for activity we load something like 30 items eagerly and then the others are deferred until you get within a viewport's height of them.

While I'm confident it could be done, it's certainly complicated and I don't think I have the time to really think about it in depth before leave. I'll try to come up with some workaround for the existing design to accomplish this.

veyndan: @JakeWharton Reigniting this discussion as much has changed!

As you know, `LazyList` no longer depends on Treehouse, and the widget children are windowed within a couple of viewports of the current viewport.

Up until recently, we didn't have a strong need for user-provided placeholders, as the number of widget children available would dictate the `itemCount` of the backing `RecyclerView`. This was possible because we weren't popping off widget children from the top of the list — instead only adding and removing widget children from the bottom of the list. Since https://github.com/cashapp/redwood/pull/1094, this is no longer the case, hence the need for placeholders.

As a result, I'm currently trying to lift the construction of placeholders from host to guest.

Ideally, we'd have an implementation similar to [what you proposed above](https://github.com/cashapp/redwood/pull/772#discussion_r1092301456), but the issue I'm seeing is that we'd have to serialize many placeholder widgets and instantiate the same number of placeholder `View`s.

Presumably view recycling at the Redwood layer would solve the latter, but I'm not too sure about the former (i.e., having to serialize many placeholder widgets). Any thoughts?

JakeWharton:
@veyndan safe to close this now?

PR #786, opened 2023-02-02
Allow duplicate widgets across dependencies

JakeWharton:
This is weird, but there isn't any technical reason to prevent this. In the future we won't be able to prevent it, so drop it now.


colinrtwhite (approved):

JakeWharton:
On second thought, this isn't true. We are going to need the set of fully-qualified widget types from dependencies for protocol generation, so that means we can always perform this check. I guess I'll leave it for now to err on the side of being strict.

PR #787, opened 2023-02-02
Generate testing APIs

swankjesse:


redwood-compose-testing/build.gradle
```diff
@@ -17,6 +17,11 @@ kotlin {
         api projects.redwoodRuntime
       }
     }
+    commonTest {
+      dependencies {
+        implementation libs.kotlin.test
+      }
+    }
```
JakeWharton, line 23: This is already in `commonMain` should should be available in tests. Was it not working?

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -0,0 +1,353 @@
...
+      }
+      .addCode("⇤),\n")
+      .addCode("⇤)\n", RedwoodTesting.RedwoodTester)
+      .build())
+    .build()
+}
+
+/*
+@RedwoodCodegenApi
+public class MutableSunspotWidgetFactory : SunspotWidgetFactory<MutableWidget> {
```
JakeWharton, line 78: This type should be `internal`.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -0,0 +1,353 @@
...
+          }
+        }
+        .build(),
+    )
+    .build()
+}
+
+/*
+@RedwoodCodegenApi
+public class MutableSunspotButton : SunspotButton<MutableWidget>, MutableWidget {
```
JakeWharton: This type should be `internal`

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -0,0 +1,353 @@
...
+    .add("layoutModifiers,\n")
+  val toStringBuilder = StringBuilder()
+    .append("${widgetValueType.simpleName}(layoutModifiers=${'$'}layoutModifiers")
+
+  for (trait in widget.traits) {
+    val type = when (trait) {
+      is Property -> trait.type.asTypeName()
+      is Event -> trait.lambdaType
+      is Children -> Stdlib.List.parameterizedBy(RedwoodTesting.WidgetValue)
+      else -> throw AssertionError()
```
JakeWharton, line 286: Isn't this sealed?

redwood-layout-testing/build.gradle
```diff
@@ -0,0 +1,22 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+apply plugin: 'app.cash.redwood.generator.testing'
```
JakeWharton, line 2: I'm on the fence about making every schema generate this.

Currently, we only make you generate the composables and the widget interfaces (and layout modifiers, but only if needed).

For specialized implementations of the widgets such as the protocol, we currently generate all of the needed widget subtypes within that schema's package. We could do that here.

One difference is that the protocol only has a single public type whereas this testing API has one value type for each widget. That would mean things like `ColumnValue` would show up as `com.example.myschema.testing.redwoodlayout.ColumnValue` instead of `app.cash.redwood.layout.testing.ColumnValue`. I suspect that might be okay.

I would feel a lot more comfortable if I got to play around with automatically propagating these dependencies. We can leave this as-is for now and I"ll play around that sometime later (and file an issue).

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/WidgetValue.kt
```diff
@@ -15,10 +15,66 @@
...
+ * Text,
+ * Button
+ * ```
+ */
+public fun List<WidgetValue>.flatten(): Sequence<WidgetValue> {
+  return sequence {
+    for (widget in this@flatten) {
+      flattenRecursive(widget)
+    }
+  }
```
JakeWharton, line 70: Not that it _really_ matters, but you could eliminate the separate loop with a `drop(1)`. Feel free to merge as-is though.
```suggestion
  return sequence { flattenRecursive(this) }.drop(1) // Skip self.
```

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/TestingGenerationTest.kt
```diff
@@ -0,0 +1,140 @@
...
+    val trait: String,
+    @Property(2)
+    @Default("{ error(\"test\") }")
+    val onEvent: () -> Unit,
+    @Children(1)
+    @Default("{}")
+    val block: () -> Unit,
+  )
+
+  @Test fun `tester happy path`() {
```
JakeWharton, line 48: I think we've almost entirely eliminated our tests which look like this. I would much prefer black-box tests, but I think your current design prohibits it with a bit _too_ much encapsulation.

I have an idea, but it's worth exploring after you merge as-is.

JakeWharton (approved):
Despite all the comments I think there's only like one actual change needed (the `internal` modifier).

redwood-compose-testing/build.gradle
swankjesse: Good eye. It was a mistake to be in commonMain, so I removed it from there.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
swankjesse: That’ll break using this in schema dependencies?

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
swankjesse: Done.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
swankjesse: It wants me to handle `ProtocolTrait`.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
JakeWharton: Ah this was before I realized you were having every schema do their own code generation for testing. See my comment below.

redwood-layout-testing/build.gradle
swankjesse: I could see us wanting to share test utilities for layouts & lazy layouts. I’m not yet sure what that’ll be, but it’s only practical if they’re shared.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/WidgetValue.kt
swankjesse: I don’t think it works 'cause `this` isn’t a WidgetValue; it’s a List.

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/TestingGenerationTest.kt
swankjesse: Yeah, the test isn’t great 'cause it’s whitespace-sensitive. But it was also very convenient to have while writing the thing.

I suspect the most value it could offer going forward is if we introduced new cross-cutting features (say, changing modifiers), we’d get an example of how that changes the output.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/WidgetValue.kt
JakeWharton: Oh, I missed the receiver. Wouldn't you have a single `WidgetValue` instance in the test to represent the content root?

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/TestingGenerationTest.kt
JakeWharton: You can put types like the widget above into `test-schema/` and add a codegen module for its test code. That also lets us pull in the generated code into tests for the Redwood modules.

Your change of an implementation-detail type from `public` to `internal` just broke this test. I'm pretty confident that with a small tweak to the generated code we can drive it without the tester to validate it behaves as expected.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/WidgetValue.kt
swankjesse: Totally unsure.

```
class RootValue  : WidgetValue {
  val children: List<WidgetValue>
}
```

vs.

```
val rootViews: List<WidgetValue>
```

We’d be boxing the list. If there’s something neat that lets us do, let’s do it.

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/TestingGenerationTest.kt
swankjesse: Can do. Though I claim that visibility isn’t just an implementation detail, and having tests for it is nice.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/WidgetValue.kt
JakeWharton: Yeah I guess the root is hidden and `snapshot()` is returning its single-children contents as a list which is almost always itself a single element, but someone _might_ do `setContent { Text(); Text() }` 🫤.

:shipit: as-is!

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/TestingGenerationTest.kt
JakeWharton: I have a branch to add API tracking of generated code: https://github.com/cashapp/redwood/compare/trunk...jw/metalava/2022-10-20. Presumably we'll also enable it for all the libraries, too. I haven't landed it because there's been too much churn, but we're nearly at a relatively stable state.

swankjesse: BTW, being able to use the Gradle plugin from within the other submodules is SO GOOD

swankjesse: ALSO - test schema caught lots of bugs. Good call.

JakeWharton: ~You don't have to change anything in this PR, though. We can always follow up.~ I didn't refresh. Guess it worked out positively.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -0,0 +1,378 @@
...
+                  is Event -> trait.lambdaType
+                  else -> throw AssertionError()
+                }
+                addProperty(
+                  PropertySpec.builder(trait.name, type)
+                    .addModifiers(PRIVATE)
+                    .apply {
+                      if (trait.defaultExpression != null) {
+                        initializer(trait.defaultExpression!!)
+                      } else if (type.isNullable) {
```
swankjesse, line 173: This new case is needed to make the test schema build!

PR #796, opened 2023-02-06
Split redwood-treehouse into shared, host, and guest modules

### LINKED ISSUE #763: Split redwood-treehouse into host and guest artifacts
Will still need a shared artifact, but this should allow separating Zipline from Compose since they don't play well together.

swankjesse:
Previously we'd hardcoded host=Android/JVM/Native and guest=JS. With this change it's possible to run guest code on other platforms as desired.

Closes #763


build-support/src/main/kotlin/app/cash/redwood/buildsupport/KmpTargets.kt
```diff
@@ -0,0 +1,88 @@
...
+ * limitations under the License.
+ */
+@file:JvmName("KmpTargets")
+package app.cash.redwood.buildsupport
+
+import org.gradle.api.Project
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+
+@JvmOverloads
+fun addAllTargets(project: Project, skipJs: Boolean = false) {
```
swankjesse, line 23: Moved this from Groovy to Kotlin so I could add a parameter, `skipJs`. That might have been possible with the include, but I prefer this!

redwood-treehouse-host/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+apply plugin: 'org.jetbrains.kotlin.plugin.serialization'
+apply plugin: 'com.android.library'
+apply plugin: 'com.vanniktech.maven.publish'
+apply plugin: 'org.jetbrains.dokka' // Must be applied here for publish plugin.
+apply plugin: 'app.cash.zipline'
+
+kotlin {
+  KmpTargets.addAllTargets(project, true /* skipJs */)
```
swankjesse, line 11: We can’t host Treehouse on JS because we don’t have a JS implementation of ZiplineLoader.

swankjesse (commented):
Some potential follow-ups:

 * Change the module name 'presenter' to 'guest' in the samples
 * Fix the two built-in widget sets to offer non-JS guest builds

build-support/src/main/kotlin/app/cash/redwood/buildsupport/KmpTargets.kt
JakeWharton: Yeah custom build plugins are the full realization of what you're wanting here. I have a branch with it started from a few months ago, but our build is surprisingly complex with lots of little special cases. This is a great start.

redwood-treehouse/build.gradle
```diff
@@ -1,17 +1,14 @@
...
 apply plugin: 'org.jetbrains.kotlin.plugin.serialization'
 apply plugin: 'com.android.library'
 apply plugin: 'com.vanniktech.maven.publish'
 apply plugin: 'org.jetbrains.dokka' // Must be applied here for publish plugin.
 apply plugin: 'app.cash.zipline'
-apply plugin: 'app.cash.redwood.build.compose'
-
-redwoodBuildCompose {
-  targetCompilations = ['jsMain']
-}
```
JakeWharton, line 11: Yay we can nuke this "feature"

JakeWharton (approved):

Issue #19, opened 2021-04-07
Parse schema using FIR instead of reflection


JakeWharton:
Switch to using FIR so that we can:
- Parse default expressions on properties

    ```kotlin
    @Widget(1)
    data class Button(
      @Property(1) val bgColor: String = "red",
    )
    ```
    will change the generated code:
    ```diff
     @Composable
     fun Button(
    -  bgColor: String,
    +  bgColor: String = "red",
     ) { .. }
    ```

- Retain comments on the type and property.

    ```kotlin
    /** Hey I'm a button */
    @Widget(1)
    data class Button(
      /** Background color, will be darkened 10% for touch state. */
      val bgColor: String,
    )
    ```
    will change the generated code:
    ```diff
    +/** Hey I'm a button */
     interface Button {
    +  /** Background color, will be darkened 10% for touch state. */
       fun bgColor(bgColor: String)
     }
    ```
    and
    ```diff
    +/** Hey I'm a button */
     @Composable
     fun Button(
    +  /** Background color, will be darkened 10% for touch state. */
       bgColor: String,
     ) { ... }
    ```

- Capture parameter names on event lambdas.

    ```kotlin
    @Widget(1)
    data class Thing(
      @Property(1) val someEvent: (x: Int, y: Int) -> Unit,
    )
    ```
    will change the generated code:
    ```diff
     @Composable
     fun Thing(
    -  someEvent: (Int, Int) -> Unit,
    +  someEvent: (x: Int, y: Int) -> Unit,
     ) { .. }
    ```

JakeWharton:
UAST would also give us default expressions without requiring an annotation

JakeWharton:
We could also theoretically use the new FIR of the compiler.

JakeWharton:
UAST bootstrap can now be copied from the lint support.

JakeWharton:
I have parsing somewhat working using PSI.

The biggest problem I foresee is that we currently parse the schema in multiple places with reflection and we do that by consuming a jar of the compiled Kotlin on the classpath. With PSI we can only access things like default expressions and comments in source. This means that we do not get them from dependencies.

There's two problems with this:
-  We currently re-parse the schema in each codegen project. When you point a widget and compose generator project at your schema they each parse it with reflection and then do codegen.

[...]

PR #801, opened 2023-02-07
Use `FqType` in `Schema`

JakeWharton:
No more `KClass` or `KType`. This also fixes a bunch of places where we didn't support nested types.

Refs #19

### DIFF
--- a/redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/FqType.kt
+++ b/redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/FqType.kt
@@ -1,0 +1,165 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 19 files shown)


redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/FqType.kt
```diff
@@ -0,0 +1,159 @@
...
+    "kotlin.Boolean.Companion" -> listOf("kotlin", "Boolean", "Companion")
+    "kotlin.Byte.Companion" -> listOf("kotlin", "Byte", "Companion")
+    "kotlin.Char.Companion" -> listOf("kotlin", "Char", "Companion")
+    "kotlin.Double.Companion" -> listOf("kotlin", "Double", "Companion")
+    "kotlin.Enum.Companion" -> listOf("kotlin", "Enum", "Companion")
+    "kotlin.Float.Companion" -> listOf("kotlin", "Float", "Companion")
+    "kotlin.Int.Companion" -> listOf("kotlin", "Int", "Companion")
+    "kotlin.Long.Companion" -> listOf("kotlin", "Long", "Companion")
+    "kotlin.Short.Companion" -> listOf("kotlin", "Short", "Companion")
+    "kotlin.String.Companion" -> listOf("kotlin", "String", "Companion")
```
JakeWharton, line 122: These are a list of Kotlin types whose enclosing class name is a type that is mapped to a JVM class. Thus, the class backing the nested Kotlin type does not have an enclosing class (i.e., a parent) and the below algorithm will fail.

In other words, `kotlin.Boolean.Companion`'s `.java` class is `kotlin.internal.jvm.BooleanCompanionObject` or something, and its `enclosingClass` is null.

JakeWharton: Put as a comment in the source.

colinrtwhite (approved):
:100:

PR #759, opened 2023-01-30
Update to Kotlin 1.8.

colinrtwhite:
Updates Redwood to Kotlin 1.8. Resolves: https://github.com/cashapp/redwood/issues/701

This works around https://github.com/Kotlin/kotlinx.serialization/issues/1454 by only running either the redwood or zipline plugin in each module (not both).

PR #803, opened 2023-02-08
Do not manually advance the clock for testing

JakeWharton:
Advance until the composition delivers a diff.


colinrtwhite (approved):
Nice! Should we also add a configurable timeout in case we're expecting a diff that never arrives?

EDIT: Applying this to https://github.com/cashapp/redwood/pull/759 stalls on `protocolSkipsLambdaChangeOfSamePresence`.

JakeWharton:
Sure. `runTest` is switching to automatically timeout after 10 seconds which will be nice since we won't have to remember to add our own timeouts everywhere. But for now I'll throw one in.

JakeWharton:
Okay I'll look tomorrow. At the very least, this behavior is much more like a real-app than a bespoke clock/emission ballet.

PR #809, opened 2023-02-08
Include generated code sources in API reference docs

### LINKED ISSUE #775: Dokka not generating docs for generated code
The following modules do not show up in docs:
- `redwood-layout-compose`
- `redwood-layout-widget`
- `redwood-treehouse-lazylayout-widget`

These are generated-code only modules, but their contents are public and should be documented.

The `redwood-treehouse-lazylayout-compose` is included, but only because it has a few regular sources. Its generated contents are still omitted from docs.

IgnatBeresnev:
Fixes #775

See Dokka's upstream issue for more details https://github.com/Kotlin/dokka/issues/2850

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,6 +108,14 @@ allprojects {
     }
   }
 
+  // DokkaTaskPartial configures subprojects for multimodule docs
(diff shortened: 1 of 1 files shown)


JakeWharton (approved):

JakeWharton:
Thank you!

PR #810, opened 2023-02-08
Split impls from serialization code

JakeWharton:
This is a no-op change in behavior but will simplify a forthcoming change to the schema model object.


JakeWharton:
A future change will probably split each impl into its own file so that we can generate a custom serializer that only handles deserialization and save on some binary size.

colinrtwhite (approved):

PR #824, opened 2023-02-15
Use default values for primitives rather than lateinit

tso:


JakeWharton:
This logic is (already) pretty complex and this only adds to it. We don't need the default expression since widgets see writes for all properties when created. I missed that first time around. Other than that I'd be tempted to make everything nullable and double-bang any non-null properties when creating the value instance. I believe this is what the old tester did. Compose ensures all required values will be set by the time you can request a value instance. There's no real need for default expressions or "primitive" defaults here.

tso:
> This logic is (already) pretty complex and this only adds to it. We don't need the default expression since widgets see writes for all properties when created. I missed that first time around. Other than that I'd be tempted to make everything nullable and double-bang any non-null properties when creating the value instance. I believe this is what the old tester did. Compose ensures all required values will be set by the time you can request a value instance. There's no real need for default expressions or "primitive" defaults here.

Is something like https://github.com/cashapp/redwood/pull/825 what you were thinking?

JakeWharton:
Yes!

PR #842, opened 2023-03-01
Add support for Spacer to redwood-layout.

colinrtwhite:
Adds support for a basic `Spacer` class to `redwood-layout`. The motivation for this is to support cases when we want to add space in between two items in a list, but want to avoid using the `padding` modifier as hiding the item it's attached to will also hide the padding. The `Spacer` uses the same units -> DPs conversion as the `padding` modifier.

API:
```
Spacer(height = 20)
```

[More context.](https://github.com/squareup/cash-treehouse/pull/1566)


redwood-layout-dom/src/main/kotlin/app/cash/redwood/layout/dom/HTMLElementRedwoodLayoutWidgetFactory.kt
```diff
@@ -46,6 +47,8 @@ class HTMLElementRedwoodLayoutWidgetFactory(
       direction = "row",
       overflowSetter = { style.overflowX = it },
     )
+
+  override fun Spacer(): Spacer<HTMLElement> = TODO("Not yet implemented")
```
JakeWharton: I think this should just be a pretty basic `<div>` with an explicit width and height.

I'm not sure if the TODO GitHub action will pick up on TODO function calls or not.

JakeWharton (approved):

swankjesse (approved):
I love how the tests work here

redwood-layout-dom/src/main/kotlin/app/cash/redwood/layout/dom/HTMLElementRedwoodLayoutWidgetFactory.kt
colinrtwhite: I'll add it in!

veyndan (approved):

PR #850, opened 2023-03-10
Kotlin 1.8.10 and friends

JakeWharton:


JakeWharton:
Zipline is compatible. Molecule 0.8.0 is out with Kotlin 1.8.10 support. I'll leave this to you to decide when to merge and upgrade all the internal things.

colinrtwhite:
I think we were thinking of waiting until `1.8.20` to update, though I'm not opposed to updating to `1.8.10` as part of the next internal Treehouse release. `1.8.10` does seems like a minor jump from `1.8.0`. @swankjesse what do you think - should we just wait for `1.8.20`?

JakeWharton:
Yeah there's little reason to wait. It'll take weeks for a 1.8.20-compatible JB Compose just as it did for this one. By then there will probably be a 1.9.0 beta and the cycle will continue.

colinrtwhite:
Agreed - waiting for the `1.8.20` JB Compose compiler could take a while.

colinrtwhite (approved):

PR #857, opened 2023-03-17
Rename TreehouseView.Content to TreehouseContentSource

swankjesse:


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseContentSource.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+public fun interface TreehouseContentSource<A : AppService> {
```
veyndan, line 18: I get we didn't have any kdoc for this before, but would it be possible to add some? Whenever I dig through Redwood source code, any and all kdoc is useful to wrap my head around why stuff is done a certain way.

veyndan (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseContentSource.kt
swankjesse: Can do. But in follow-up. I’m in the process of writing lots of docs on the more interesting type this sits beside.

PR #860, opened 2023-03-18
Move WidgetSystem from TreehouseView to Content

swankjesse:
It's another type that transitively depends on TreehouseApp and Zipline, that we don't need in TreehouseView.

### DIFF
--- a/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/WidgetSystem.kt
+++ b/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/WidgetSystem.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+import app.cash.redwood.protocol.widget.DiffConsumingNode
+import app.cash.redwood.protocol.widget.ProtocolMismatchHandler
+import kotlinx.serialization.json.Json
+
+public fun interface WidgetSystem {
+  /** Returns a widget factory for encoding and decoding changes of a [Content]. */
+  public fun widgetFactory(
+    app: TreehouseApp<*>,
+    json: Json,
+    protocolMismatchHandler: ProtocolMismatchHandler,
+  ): DiffConsumingNode.Factory<*>
+}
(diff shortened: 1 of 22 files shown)


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/WidgetSystem.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import app.cash.redwood.protocol.widget.DiffConsumingNode
+import app.cash.redwood.protocol.widget.ProtocolMismatchHandler
+import kotlinx.serialization.json.Json
+
+public fun interface WidgetSystem {
+  /** Returns a widget factory for encoding and decoding changes of a [Content]. */
+  public fun widgetFactory(
+    app: TreehouseApp<*>,
+    json: Json,
```
swankjesse, line 26: It wasn’t good for TreehouseView to depend on JSON and protocol stuff. I promoted this to a top-level class to later support moving TreehouseView into a different module; perhaps with a different name.

JakeWharton:
> It's another type that transitively depends on TreehouseApp and Zipline, that we don't need in TreehouseView.

Hmm I do not believe this statement to be true.

The widget system is tied to the `TreehouseView` subtype implementation UI toolkit. I actually think a `TreehouseView` should gain a `V` type variable now that its `A` is gone, and that should be shared by `WidgetSystem`. A `View`-based `TreehouseView` should only accept a `View`-based `WidgetSystem`.

The need for the widget system to vary what it creates based on the `A` should be something that is handled at the caller site and facilitated by setting a new instance of the system to a view (such as in the case of `LazyColumn`).

Thought of another way, why would the app content for rendering in Compose need knowledge of the rendering UI toolkit? It can't do anything with the widget system directly nor gain references to anything created by it. It helps if you think of this separately from `LazyColumn` usage which I think we can still solve as I said above.

swankjesse:
I’ve been attempting to split `TreehouseView` from Zipline things, and as a side-effect of that I’ve also been breaking its dependency on protocol things.

`WidgetSystem` is mostly a protocol class: it knows about `Json` and `DiffConsumingNode` and `ProtocolMismatchHandler`. Its ultimate customer is always a `ProtocolBridge`. Perhaps its could be renamed to `ProtocolWidgetSystem` or something.

If we can break `TreehouseView` from both protocol and Zipline, it becomes a pretty basic widget that can produce a host configuration and receive children. What’s neat here is if `TreehouseView` receives already-constructed children, it doesn’t even need to care about what protocol or schema they were created by.

JakeWharton:
The way `LazyColumn` works today is a replica of how a whole application is setup. In a "normal" application you would be able to set up the root view and widget system without knowledge of the app and composition.

It's only because `LazyColumn` needs to create a bunch of mini app setups do we want to have the enclosing widget system have knowledge of the app. I would prefer that breaking of this clean layering to be something that's specific to the client setup of their widget system that includes `LazyColumn` and not something subjected to all widget systems.

Also, if we change `LazyColumn` as proposed elsewhere to communicate with the main composition through API rather than mini-compositons the need for the layering violation goes away (or rather changes to be a shared service interface specific to lazy columns).

JakeWharton:
My previous comment is from before your last reply. Typing with babby is slow.

The widget side of the protocol was never refactored like the Compose side to have a better API for setup (mostly because it also needed some of this work). I think with a change to its API we still are able to have the widget system expose itself to the app from the view so it can create widget factory instances. Basically, a widget factory factory, which is what we have today on the Compose side, but with slightly clever naming to hide it.

We can defer this work if needed, but it was my original goal.

swankjesse:
Lemme move TreehouseApp out of WidgetSystem instead of moving WidgetSystem out of TreehouseView!

PR #863, opened 2023-03-18
Internal cleanup of TreehouseContent

swankjesse:
Rename TreehouseContent to TreehouseAppContent. The name TreehouseContent is already taken by Compose.

Rename RealBinding to ViewContentCodeBinding.

Move ViewContentCodeBinding into the same file as TreehouseAppContent.

Remove the unused type parameter on Content.


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
```diff
@@ -84,7 +149,7 @@ internal class RealBinding<A : AppService>(
   private var treehouseUiOrNull: ZiplineTreehouseUi? = null
 
   /** Only accessed on [TreehouseDispatchers.ui]. */
-  private var firstDiff = true
+  private var diffCount = 0
```
JakeWharton, line 152: Why this change?

JakeWharton (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
swankjesse: Yeah, this is off topic from the rest of the PR. I added `codeLoadCount` to replace tracking the previous binding, and I thought these two should be symmetric.

I’m contemplating about changing `CodeListener` to something less opinionated.

```
fun onCodeLoading(content: Content, view: TreehouseView, count: Int)
fun beforeDiffApplied(content: Content, view: TreehouseView, count: Int)
```

We don’t need this flexibility yet.

PR #865, opened 2023-03-20
Pass the target view to the CodeListener

swankjesse:


JakeWharton (commented):
What's the goal with passing it?

The argument was previously removed because it was only able to be used on Android with an unsafe cast and only with view-based trees. It is a useless argument otherwise with Compose UI-based trees, or if you're on any other platform.

Instead, the calling code listener is supposed to capture the concrete subtype so that it doesn't need to perform an unsafe cast which also means non-Android platforms and Compose UI on Android can use the same patterns to show loading states.

swankjesse:
CodeListener sits in on the boundary between the Zipline world (it cares about hot reloading) and the widget world (it will update UIs when it is signaled.)

It used to be a member of TreehouseView, but I don’t like polluting TreehouseView with the knowledge of hot reloading. In particular I’d like to potentially rename TreehouseView to RedwoodView and break all dependencies on dynamic code.

Which now means it lives with the Zipline / code loading stuff. And in this situation the content doesn’t necessarily have a reference to the view instance it’ll be bound to.

I’m open to suggestions here!

swankjesse:
(And I definitely don’t feel offended by the downcast; this thing’s job is to do widget system-specific things.)

swankjesse:
... and one small nit: it’s a safe cast now, as TreehouseWidgetView doesn’t have type parameters.

JakeWharton:
The unsafety of the cast stems from the part which knows about code loading being unaware of which UI toolkit is actually in use by the UI layer. If I change the UI toolkit from Views to Compose UI nothing about the Zipline layer should change. However, the cast would now fail at runtime, but even worse the pattern of casting this argument no longer affords me the ability to manipulate the UI in any way with Compose UI.

If you just want to get unblocked I can approve it, but the something in the structure of how we've set up this integration is flawed and will have to change. Not so much for the desire for clean layering, but simply because we're exploiting a property of a UI toolkit that is going away soon.

JakeWharton (approved):

swankjesse:
I think there might be a way to use this without any downcasts...

Use the TreehouseView’s WidgetSystem to create a Spinner, then add it to the TreehouseView’s children. It’d be cosplaying as Redwood, but it aught to work on arbitrary UI systems.

JakeWharton:
That's certainly a clever trick that would work well, but it really feels like we're doing some gymnastics to solve this use case where I would hope we could find a reasonable API solution.

PR #866, opened 2023-03-21
Annotate host public API with `@ObjCName`.

### LINKED ISSUE #520: Use @ObjCName in Kotlin 1.8 to improve ObjectiveC/Swift interop
Kotlin 1.8 comes with a new [`@ObjCName`](https://github.com/JetBrains/kotlin/blob/84650d08af34b1c23ce3ae5a758b8d98b3a06a46/libraries/stdlib/src/kotlin/annotations/NativeAnnotations.kt#L68) annotation that we can use to improve the names that are exposed to ObjectiveC/Swift code.

This should let us expose better names than `Redwood_layout_uiviewUIViewRedwoodLayoutWidgetFactory` and `Redwood_treehouseTreehouseApp`.

colinrtwhite:
Resolves: https://github.com/cashapp/redwood/issues/520


redwood-runtime/src/commonMain/kotlin/app/cash/redwood/LayoutModifier.kt
```diff
@@ -94,6 +96,7 @@ public interface LayoutModifier {
    */
   // The companion object implements `LayoutModifier` so that it may be used as the start of a
   // modifier extension factory expression.
+  @ObjCName("Companion")
   public companion object : LayoutModifier {
```
colinrtwhite, line 100: Named this as `LayoutModifier.Companion` for consistency, but might be better as `LayoutModifier.empty` or something else?

JakeWharton:
Huh I expected to see a bunch of our iOS code also change such as in the samples. I guess this doesn't affect the mangled names which get exposed to Swift?

colinrtwhite:
@JakeWharton Ah yep, we need to add `exact = true` to get the behaviour we're after. Updated the PR.

After using this annotation I feel like it's a good candidate for a compiler flag (or compiler plugin) to automate the naming. Though, I guess that could cause issues due to Swift not supporting packages.

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/LayoutModifier.kt
colinrtwhite: Renamed to `EmptyLayoutModifier` as the interop always generates top level classes.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/FlowWithInitialValue.kt
```diff
@@ -15,12 +15,14 @@
...
 
+import kotlin.native.ObjCName
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.serialization.Contextual
 import kotlinx.serialization.Serializable
 
 @Serializable
+@ObjCName("FlowWithInitialValue", exact = true)
 public data class FlowWithInitialValue<T>(
```
JakeWharton, line 26: Can't wait to kill this

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeListener.kt
```diff
@@ -15,6 +15,9 @@
  */
 package app.cash.redwood.treehouse
 
+import kotlin.native.ObjCName
+
+@ObjCName("CodeListener", exact = true)
```
JakeWharton, line 20: These ones we might have to eventually go back and prefix with Treehouse or something. There's a few really generic names from this module. I guess we'll see over time.

JakeWharton (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeListener.kt
colinrtwhite: Yep, same with `Widget`. Worst case I believe they can still be referenced using the module name (e.g. `RedwoodKt.Widget`).

Issue #855, opened 2023-03-16
Width and Height configuration for children of Rows and Columns


swankjesse:
### Problem Statement

I’ve got a column where I’d like to configure the minimum size of a child element. Perhaps something like this:

```
  @Composable
  Column {
    ...
    Button(
      layoutModifier = LayoutModifier
        .horizontalAlignment(CrossAxisAlignment.Stretch)
        .minHeight(40),
      text = theButtonLabel,
      ...
    )
  }
```

For this occurrence I think I want to specify a _minimum_ height, because if the row’s contents exceed that height (perhaps due to a long label text-wrapping), I’m happy with the element being taller than that size.

In another part of my layout I’d like to specify an element’s exact width=0 with `grow(1.0)` so that I can achieve 50/50 balance for adjacent elements:

```
  @Composable
  Row {
    Button(
      layoutModifier = LayoutModifier
        .horizontalAlignment(CrossAxisAlignment.Stretch)
        .width(0)
        .grow(1.0),
      text = "LEFT",
    )
    Button(
      layoutModifier = LayoutModifier
        .horizontalAlignment(CrossAxisAlignment.Stretch)
        .width(0)
        .grow(1.0),
      text = "RIGHT",
    )
  }
```

Without locking the width in at 0, the side-by-side elements only balance if their contents are the exact same width.

### Syntax

I’d like to express these layout goals in the API. The above syntax is illustrative; I’m not at all attached to it. Borrowing terms from HTML or iOS or Android would be ideal.

### Margin, Padding, Size

We have to decide if width and height includes padding or not. My preference is we follow [the standard CSS Box Model](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/The_box_model#the_standard_css_box_model) and not include padding in width or height.

I’d also like for us to rename `Padding` to `Margin`, if that’s closer to its behavior. In particular, if an element doesn’t paint its background in the padding area we should probably rename it to margin.

colinrtwhite:
This is relatively straightforward to implement - we could override a child's minimum width/height by overriding the return value of `Measurable.minWidth`/`Measurable.minHeight`. Likewise, we can override an item's width/height by overriding the return value of `Measurable.measure`. That said, do we want to provide this level of control in Redwood? I think I'm OK with it as we can always deprecate/remove these functions later if they cause issues. Also I think your proposed syntax is good.

[...]

PR #867, opened 2023-03-21
Rename Padding -> Margin.

colinrtwhite:
Also renames `start`/`end` -> `left`/`right`. Left/right fits better with the flexbox spec and we don't support right-to-left layouts.

Follow up from the conversation here: https://github.com/cashapp/redwood/issues/855


swankjesse (approved):

veyndan:
> Also renames `start`/`end` -> `left`/`right`. Left/right fits better with the flexbox spec and we don't support right-to-left layouts.

Have you got a link to the spec? I'm kinda surprised because CSS logical properties are promoted pretty heavily (e.g., `flex-start`).

colinrtwhite:
@veyndan Margin isn't really part of the flexbox spec so it uses the default CSS `left`/`right` names for margin: https://www.w3schools.com/css/css_margin.asp

veyndan:
True, `flex-start` was a bad/wrong example. Margin's logical property counterparts to `margin-left`/`margin-right` are [`margin-inline-start`/`margin-inline-end`](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline). I'm just a bit concerned going with `left`/`right` as it's exclusionary of RTL (and vertical) writing modes.

JakeWharton:
It does seem very weird to not use start/end in anything newly built.

colinrtwhite:
@veyndan That's a good point. I figured that using `left`/`right` was more in-line with web, but that seems incorrect. I'll revert back to `start`/`end`.

PR #870, opened 2023-03-22
Allow TreehouseAppContent code to be preloaded

### LINKED ISSUE #861: Content.preload()
For LazyColumn it’ll be a performance win to run asynchronous guest code eagerly, before the target `TreehouseView` is attached. One possible way to implement this is a `preload()` function on `Content`:

```
public interface Content {
  /**
   * Prepare the content for an upcoming call to [bind]. This may yield
   * lower latency for lazy layouts.
   * 
   * To avoid resource leaks every call to [preload] must be followed by
   * a call to [unbind], whether or not there is a call to [bind].
   */
  public fun preload(view: TreehouseView)

  ...
}
```

The `preload` function doesn’t _really_ need the `TreehouseView` yet, but accepting it keeps the `hostConfiguration` quite simple. Otherwise we’d need to accept a `HostConfiguration` and fuse it to the `StateFlow<HostConfiguration>` that follows.

veyndan:
Closes #861.

### DIFF
--- a/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/SequentialStateFlow.kt
+++ b/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/SequentialStateFlow.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.collectLatest
+
+internal class SequentialStateFlow<T>(
+  first: StateFlow<T>,
+) : StateFlow<T> {
+  public val stateFlowValue: MutableStateFlow<StateFlow<T>> = MutableStateFlow(first)
+
+  override val replayCache: List<T> = listOf()
+
+  override suspend fun collect(collector: FlowCollector<T>): Nothing {
+    stateFlowValue.collectLatest { stateFlow ->
+      stateFlow.collect {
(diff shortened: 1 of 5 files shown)


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
```diff
@@ -163,31 +306,52 @@ private class ViewContentCodeBinding<A : AppService>(
...
+    if (diffCount++ == 0) {
+      view.reset()
+      codeListener.onCodeLoaded(view, isInitialLaunch)
 
-      bridge.sendDiff(diff)
+      // Update the TreehouseAppContent's ViewState on the first diff.
+      val currentState = stateFlow.value
+      if (currentState.codeState is CodeState.Running && currentState.codeState.viewContentCodeBinding == this) {
+        stateFlow.value = State(currentState.viewState, CodeState.Running(this, hasDiffs = true))
+      }
```
veyndan: @swankjesse In the spirit of just having one way of doing things, is there any material difference between `awaitContent` and `codeListener#onCodeLoaded` (apart from one being a suspend function and the other being a callback)?

I know I said it'd be nicer to have a suspend function for the `ViewLazyColumn` implementation, but making a callback suspending isn't that difficult.

veyndan: Actually, I think this code is in the wrong place. It's only executed if we have a `view` that's non-null, which isn't what we want. Gonna move it up.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/SequentialStateFlow.kt
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.collectLatest
+
+public class SequentialStateFlow<T>(
```
JakeWharton: internal

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
swankjesse: ooooh good catch.

swankjesse (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
veyndan: https://github.com/cashapp/redwood/pull/870/commits/3ee06eba42d766bcd4354bb593ff041014599f10

PR #879, opened 2023-03-23
Hide system UI in ComposeUiSpacerTest

veyndan:
We can actually see what the desired rendered effect is.


JakeWharton (approved):

colinrtwhite:
Thanks! Not sure what happened to the previous screenshots

PR #890, opened 2023-03-30
Limit `redwood-treehouse-lazylayout-compose` targets to JS only

veyndan:
When `redwood-trehouse-lazylayout-compose` was used on non-JS targets, it causes exceptions like:

```
> Task :compileDevelopmentExecutableKotlinJs FAILED
e: Module "com.example:treehouse" has a reference to symbol app.cash.redwood.treehouse.lazylayout.paging/itemsIndexed$composable|-2783174316103845623[0]. Neither the module itself nor its dependencies contain such declaration.

This could happen if the required dependency is missing in the project. Or if there is a dependency of "com.example:treehouse" that has a different version in the project than the version that "com.example:treehouse" was initially compiled with. Please check that the project configuration is correct and has consistent versions of all required dependencies.
```

~All the transitive dependencies allow all KMP targets, so I don't see any advantage of restricting this to being JS-only.~


JakeWharton (approved):

JakeWharton:
Hmm I'm not exactly sure this makes sense to move entirely, though.

"Guest" means JS (or WASM in the future). I suspect it is only a MPP project so that we can share the interfaces across to the host-side. The composables using LazyColumn will always be guest code running in JS only.

veyndan:
Yeah that makes sense. I just updated the PR so the module _only_ targets JS.

PR #894, opened 2023-04-04
Fix using grow/shrink inside a nested column/row.

### LINKED ISSUE #852: LayoutModifier.grow() on nested column changes width?
Here’s a layout. Notice the nested columns.

```kotlin
    Column(
      width = Constraint.Fill,
      height = Constraint.Fill,
      padding = Padding(vertical = 80, horizontal = 10)
    ) {
      Text(
        text = "Application Toolbar",
      )
      Column(
        padding = Padding(all = 5),
        width = Constraint.Fill,
        layoutModifier = LayoutModifier
          .horizontalAlignment(CrossAxisAlignment.Stretch),
      ) {
        Text(
          layoutModifier = LayoutModifier
            .horizontalAlignment(CrossAxisAlignment.Center),
          text = "Centered label".plain(),
        )
      }
    }
```

layout_no_grow

I’d like to vertically center the contents of the nested column, so I grow that column to fill the height of the screen.

```diff
     Column(
       width = Constraint.Fill,
       height = Constraint.Fill,
       padding = Padding(vertical = 80, horizontal = 10)
     ) {
       Text(
         text = "Application Toolbar",
       )
       Column(
         padding = Padding(all = 5),
         width = Constraint.Fill,
         verticalAlignment = MainAxisAlignment.Center,
         layoutModifier = LayoutModifier
-          .horizontalAlignment(CrossAxisAlignment.Stretch),
+          .horizontalAlignment(CrossAxisAlignment.Stretch)
+          .grow(1.0),
       ) {
         Text(
           layoutModifier = LayoutModifier
             .horizontalAlignment(CrossAxisAlignment.Center),
           text = "Centered label".plain(),
         )
       }
     }

```

layout_grow

I expected the width of the nested column to still fill the width of the screen, but now it doesn’t.
Note that I couldn’t reproduce this on a single `Column`; I needed to nest columns to trigger this behavior.

colinrtwhite:
Fixes: https://github.com/cashapp/redwood/issues/852

This restores a secondary layout pass when using `grow`/`shrink`, which is fine for Views, but makes Compose UI throw when using those modifiers. I'm going to follow up with a larger refactor PR that defers the `measure` call until after we attempt to grow/shrink items.


JakeWharton:
Does paparazzi just draw a blank screen when the test crashes? I would have expected either the build to fail or to see an ignored test if we're violating some Compose UI invariant.

redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_shortRow.png
swankjesse, line 1: This doesn’t seem right...

colinrtwhite: Yep, need to make some more changes...

colinrtwhite:
@JakeWharton I think Paparazzi displays a blank screen unfortunately.

JakeWharton:
😬 

Can't find any issues on Paparazzi about it. There's some for views from back in 2020. When I find time I'll file one with a reproducer. It's might be something Compose UI-specific.

colinrtwhite:
Closing this out. This doesn't address the problem completely and there's a better fix on the way.

PR #895, opened 2023-04-04
Disable system UI in redwood-layout screenshot tests.

colinrtwhite:


veyndan (approved):

JakeWharton (approved):
No preference here either way. In the other test the layout was appearing behind the system UI. Here it correctly appears within it making the need to disable much lower. But I'm of the opinion system UI with paparazzi should be opt-in so certainly don't mind it being removed.

veyndan:
Should this PR be targeting `trunk`?

JakeWharton:
It was stacked, but it maybe worth rebasing since the base one seems like it needs more work?

colinrtwhite:
Yep, this needs to be retargeted from the base branch.

chrisbanes:
Rebased onto trunk.

JakeWharton:
CI didn't kick before for some reason

PR #908, opened 2023-04-12
Connect to Android's Choreographer for a frame pulse

swankjesse:


redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/TickFrameClockAndroid.kt
```diff
@@ -0,0 +1,58 @@
...
+    choreographer.postFrameCallback(callback)
+  }
+
+  // Collect frames as they arrive.
+  launch(dispatchers.zipline) {
+    latestFrameTimeNanos.consumeEach { frameTimeNanos ->
+      clockService.sendFrame(frameTimeNanos)
+    }
+  }
+}
```
swankjesse, line 58: In this implementation I didn’t build on the native-Android Compose frame clock stuff.

I’m open to changing to do that, but I felt like this was the minimum amount of abstraction.

swankjesse:
There’s a very different solution we could attempt here. . . .

1. Build a ZiplineService that works like `MonotonicFrameClock`, specifically organized around callbacks. This is different from our current approach, where we require frames to be broadcast whether or not anyone is listening.
2. Bridge that from the guest to the host.

I’m open to that! We may also optionally batch up calls so if many guest-side callers request a frame, we only request one frame on the host site.

swankjesse:
I built this here & now 'cause I saw some slowdowns in EmojiSearch and this seemed like a potential source.

swankjesse:
Ideas above implemented here:
https://github.com/cashapp/redwood/pull/957

(We still have the option to do an Android Choreographer-driven pulse, but it’s much less urgent.)

PR #913, opened 2023-04-16
Expose the host's safe area insets in HostConfiguration.

colinrtwhite:
Adds a new `safeAreaInsets` property to `HostConfiguration`. This lets us modify the UI to account for a device's reserved system UI area if necessary. It can be combined with the other layout components like so:

```kotlin
val configuration = LocalHostConfiguration.current

// Column
Column(
  margin = configuration.safeAreaInsets,
) {
  // Content
}

// Spacer
Spacer(
  height = configuration.safeAreaInsets.top
)
// Content
```

Eventually we can add an argument to automatically inset the `TreehouseView` based on the device's insets, however I opted not to implement that yet as we'll probably need to do it without `redwood-layout` and `Spacer`.


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/SequentialStateFlow.kt
```diff
@@ -23,15 +23,13 @@ import kotlinx.coroutines.flow.collectLatest
 internal class SequentialStateFlow<T>(
   first: StateFlow<T>,
 ) : StateFlow<T> {
-  public val stateFlowValue: MutableStateFlow<StateFlow<T>> = MutableStateFlow(first)
+  val stateFlowValue: MutableStateFlow<StateFlow<T>> = MutableStateFlow(first)
 
-  override val replayCache: List<T> = listOf()
+  override val replayCache: List<T> = listOf(value)
```
JakeWharton: Otherwise you always get the initial value
```suggestion
  override val replayCache: List<T> get() = listOf(value)
```

redwood-treehouse-composeui-insets/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+    androidMain {
+      dependencies {
+        implementation libs.androidx.core
+      }
+    }
+  }
+}
+
+android {
+  namespace 'app.cash.treehouse.composeui'
```
JakeWharton:
```suggestion
  namespace 'app.cash.treehouse.composeui.insets'
```

JakeWharton (approved):
I'm okay with moving forward with this as-is for now. The tests and support for all platforms is great!

However, long term, I think we should change the approach to be more like Compose UI. This would mean two things:
- The insets are available as their own composition local within pure Redwood (not Treehouse).
- The insets are consumable rather than static so that they reflect the composables position in the native UI toolkit view hierarchy.

Maybe we can track that in an issue and figure out the right way to do it over time.

redwood-treehouse-composeui-insets/src/androidMain/kotlin/app/cash/redwood/treehouse/composeui/insets.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.foundation.layout.systemBars
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.LocalLayoutDirection
+import androidx.compose.ui.unit.LayoutDirection
+import app.cash.redwood.layout.api.Margin
+
+@Composable
+public actual fun safeAreaInsets(): Margin {
+  val layoutDirection = LocalLayoutDirection.current
+  return WindowInsets.systemBars.asPaddingValues().toMargin(layoutDirection)
```
chrisbanes: Probably want to use `safeDrawing` insets here ([link](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.foundation.layout.WindowInsets.Companion).safeDrawing())). System bars only contains the nav/status bars, but we also need to handle IME and cutouts.

redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
```diff
@@ -146,3 +146,17 @@ public fun Margin(
 public fun Margin(
   all: Int = 0,
 ): Margin = Margin(all, all, all, all)
+
+// Temporary constructor before migrating everything to doubles.
+@Stable
+public fun Margin(
+  left: Double = 0.0,
```
chrisbanes, line 153: Meta comment: no start/end?

redwood-treehouse-host/src/androidInstrumentedTest/kotlin/app/cash/redwood/treehouse/TreehouseWidgetViewTest.kt
```diff
@@ -113,6 +117,26 @@ class TreehouseWidgetViewTest {
     }
   }
 
+  @Test fun hostConfigurationEmitsSafeAreaInsetsChanges() = runTest {
+    val layout = TreehouseWidgetView(context, throwingWidgetSystem)
+    layout.hostConfiguration.test {
+      assertEquals(HostConfiguration(safeAreaInsets = Margin.Zero), awaitItem())
+      val insets = Insets.of(10, 20, 30, 40)
+      val windowInsets = WindowInsetsCompat.Builder()
+        .setInsets(WindowInsetsCompat.Type.systemBars(), insets)
```
chrisbanes, line 125: Add a test for IME insets (IME visible) too?

chrisbanes:
+1 to Jake's comment about making these consumable in the long term. 

I'm personally not a huge fan of the way that Compose UI exposes this though, but we can work out the mechanics later

colinrtwhite:
Sounds good! I'll file an issue to track better support for this. For the moment this unblocks one of our current screens that needs to know the status bar size to avoid accidentally drawing under it.

> However, long term, I think we should change the approach to be more like Compose UI.

Makes sense, though I'm curious what the API would look like for consuming them. Also I think we'll need to avoid applying them as a `Modifier` given how Redwood modifiers are currently only read by the parent widget.

redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
colinrtwhite: Yep! Will fix this soon. I changed this to `left`/`right` as I'm pretty sure we're not handling LTR properly in some spots.

redwood-treehouse-host/src/androidInstrumentedTest/kotlin/app/cash/redwood/treehouse/TreehouseWidgetViewTest.kt
colinrtwhite: @chrisbanes Hmm doesn't seem like there's an easy way to test IME inset changes (`dispatchApplyWindowInsets` strips those insets), but let me know if I'm missing something!

chrisbanes: Ahhh, this is an instrumented test. Yeah this is going to be difficult as you'll have to manually show the IME (which is a nightmare on emulators). Let's skip 👍

redwood-layout-api/src/commonMain/kotlin/app/cash/redwood/layout/api/properties.kt
veyndan: For context, there was a bit of a discussion about it [here](https://github.com/cashapp/redwood/pull/867#issuecomment-1482573336)

chrisbanes: Ah cool, thanks!

PR #934, opened 2023-04-19
Add Dp and Density classes.

colinrtwhite:
Introduces `Dp` and `Density` classes to Redwood with similar designs to Compose UI.

This cleans up a lot of the guest -> host size value conversion logic that was happening in the `redwood-layout` classes and makes it reusable and type-safe. It also changes our size values to be represented as doubles, which helps us avoid losing precision (by rounding to ints) when sending size info host -> guest (like with `HostConfiguration.safeAreaInsets`).

It also reverts `Margin.left/right` back to `Margin.start/end`. Though, I think there's probably still fixes that need to be done to properly support RTL.


redwood-layout-api-density/src/macosMain/kotlin/app/cash/redwood/layout/api/density.kt
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.layout.api
+
+/** MacOS's coordinate system accounts for density automatically. */
```
colinrtwhite, line 18: I'm assuming this is true based on how iOS handles density, though I haven't actually checked this.

redwood-layout-api-density/src/androidMain/kotlin/app/cash/redwood/layout/api/density.kt
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package app.cash.redwood.layout.api
+
+import android.content.res.Resources
+
+// Android uses 2.75 as a density scale for most recent Pixel devices and iOS
+// uses 3. This aligns the two so the generic values used by Redwood layout are
+// visually similar on both platforms.
+internal actual val DensityMultiplier = 1.1
```
colinrtwhite: There's probably a better way to roughly normalize iOS and Android densities, but this is what we've been doing so far.

chrisbanes: Hmmm, is there a reason why we'd want this? My gut feeling is that we should just abide from the platform choice.

chrisbanes (commented):
Looks good overall. Not sure on the `DensityMultiplier` thing though.

redwood-layout-api-density/src/macosMain/kotlin/app/cash/redwood/layout/api/density.kt
```diff
@@ -0,0 +1,23 @@
...
+public val Density.Companion.Default: Density get() = DefaultDensity
+
+private val DefaultDensity = Density(1.0)
+
+internal actual val DensityMultiplier = 1.0
```
JakeWharton: Can all these multiplier `actual`s be `const`?

redwood-layout-api/build.gradle
```diff
@@ -12,6 +12,7 @@ kotlin {
   sourceSets {
     commonMain {
       dependencies {
+        api projects.redwoodLayoutApiDensity
```
JakeWharton, line 15: What's the motivation for a different API module for these?

JakeWharton: Actually I suspect we'll be moving these all into Redwood's core since we want to expose things like screen information, container size, etc. as a composition local and also through `HostConfiguration` in Treehouse.

JakeWharton (approved):

redwood-layout-api/build.gradle
colinrtwhite: I created a separate module so we can expose `Dp` and `Margin` to the guest code, but don't need to depend on the rest of the layout API (i.e. `MainAxisAlignment`, `CrossAxisAlignment`, etc.)

redwood-layout-api-density/src/androidMain/kotlin/app/cash/redwood/layout/api/density.kt
colinrtwhite: @chrisbanes It's to normalize the density values between iOS and Android (and other future platforms). Without a multiple, `4.dp` looked much smaller on Android than iOS.

chrisbanes: This is tricky because that's mostly an intentional choice for the Android OS. I'm not sure we should be getting into the game of creating 1:1(ish) pixel consistency.

WDYT?

colinrtwhite: @chrisbanes That's a good point and we probably need to chat with product + design about it, though my hunch is that they'll want similar sizes on both platforms.

I vote we keep `DensityMultiplier` as an implementation detail that we can change later if necessary, but keep it as is for the moment since we have existing screens built with this assumption.

chrisbanes: Sounds good to me!

chrisbanes (approved):

PR #939, opened 2023-04-20
Use the test schema for some code generation tests

JakeWharton:
As we move toward parsing with PSI and thus loading schema information via JSON it's much easier to consume a multi-schema tree when it's split across actual Gradle modules. The test schema relies on the layout schema and thus fulfils the same structure.


JakeWharton:
GitHub Actions is having problems seeing Maven Central, it seems.

colinrtwhite (approved):

JakeWharton:
Hmm actually this is broken because the schema types and the layout modifier types can't be on the same classpath as they exist in the same package.

PR #942, opened 2023-04-20
Add support for `LazyRow`

veyndan:
Much of this mimics that of Compose UI as much as possible (e.g., file names, the use of `isVertical`).

I've also renamed our GitHub tab from `redwood/lazycolumn` to `redwood/lazylist`, as there are almost zero differences between `LazyColumn` and `LazyRow`, and any issues with either one probably means the other one also has such an issue.

### DIFF
--- a/redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyList.kt
+++ b/redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyList.kt
@@ -16,9 +16,13 @@
 package app.cash.redwood.treehouse.lazylayout.composeui
 
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.LazyListScope
+import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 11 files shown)


redwood-treehouse-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -36,13 +36,19 @@ import platform.UIKit.row
...
+) : LazyList<UIView> {
   private val dataSource = TableViewDataSource(treehouseApp, widgetSystem)
   private val root = UITableView().apply {
-    this.dataSource = this@UIViewLazyColumn.dataSource
+    this.dataSource = this@UIViewLazyList.dataSource
+  }
+
+  override fun isVertical(isVertical: Boolean) {
+    if (!isVertical) {
+      TODO()
```
veyndan, line 51: @apatronl @dnagler Is it possible to toggle the scrolling direction of a `UITableView` — from vertical to horizontal? Or would we need/it be simpler to just convert over to a `UICollectionView` (which I believe supports it)?

dnagler: You'll definitely want to use a `UICollectionView` for horizontal scrolling. As per [Apple's docs](https://developer.apple.com/documentation/uikit/uitableview#//apple_ref/doc/uid/TP40006943-CH3-SW44), `UITableView`s are meant for vertical scrolling only.

swankjesse (approved):

redwood-treehouse-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/uiview/UIViewLazyList.kt
JakeWharton, line 51: I wonder if the TODO action catches `TODO()`. Maybe put a proper `// TODO` to make sure it gets tracked?

redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyList.kt
```diff
@@ -32,33 +36,51 @@ import app.cash.redwood.treehouse.TreehouseApp
...
 
   override val value = @Composable {
-    LazyColumn(
-      horizontalAlignment = Alignment.CenterHorizontally,
-      modifier = Modifier
-        .fillMaxWidth(),
-    ) {
+    // TODO Remove statically sized containers (https://github.com/cashapp/redwood/pull/854).
+    val itemBoxModifier = if (isVertical) Modifier.height(64.dp) else Modifier.width(64.dp)
+    val content: LazyListScope.() -> Unit = {
```
JakeWharton, line 61: Hmm I assume because this isn't a composable lambda that we aren't able to use `movableContentOf`/`movableContentWithReceiverOf`. Probably fine since no one is changing the orientation at any rate where it would matter.

JakeWharton (approved):

redwood-treehouse-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/uiview/UIViewLazyList.kt
veyndan: Thanks for the info! Will probably do it in a follow-up

PR #947, opened 2023-04-21
Fix implicit task dependency issue

JakeWharton:
Redwood Lint can't use the project directory as a "directory" input as gradle considers all the contents of the directory to be inputs to the task. The "official" recommendation is to use a String input with the absolute path as the input.


JakeWharton:
Yanked from #823. No reason to wait.

Issue #903, opened 2023-04-11
iOS flexbox appears to be broken


JakeWharton:
Views appear to be correctly added to the hierarchy but they have no size or position.

<img width="1512" alt="Screen Shot 2023-04-11 at 1 53 07 PM" src="https://user-images.githubusercontent.com/66577/231247592-b3632f42-459a-4ec8-8341-0ab4064f826c.png">

<img width="1512" alt="Screen Shot 2023-04-11 at 1 59 00 PM" src="https://user-images.githubusercontent.com/66577/231249386-e5aba43e-ce83-4def-a2a6-699c7d5560d9.png">

colinrtwhite:
I'm not 100% sure that this is due to flexbox. If I remove the `Column` from the layout and only keep `Button("+1", onClick = { count++ })` the button shows up, but clicking it throws:

```
2023-04-11 20:39:32.787701-0700 CounterApp[66576:2686290] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[CounterKt_kobjcc0 clicked]: unrecognized selector sent to instance 0x6000027a5fb0'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000018040e7c8 __exceptionPreprocess + 172
	1   libobjc.A.dylib                     0x0000000180051144 objc_exception_throw + 56
	2   CoreFoundation                      0x000000018041d47c +[NSObject(NSObject) instanceMethodSignatureForSelector:] + 0
	3   CoreFoundation                      0x00000001804126c8 ___forwarding___ + 1308
	4   CoreFoundation                      0x0000000180414b4c _CF_forwarding_prep_0 + 92
	5   UIKitCore                           0x000000010c85f6c8 -[UIApplication sendAction:to:from:forEvent:] + 96
	6   UIKitCore                           0x000000010c2154bc -[UIControl sendAction:to:forEvent:] + 108
	7   UIKitCore                           0x000000010c215800 -[UIControl _sendActionsForEvents:withEvent:] + 268
	8   UIKitCore                           0x000000010c21242c -[UIButton _sendActionsForEvents:withEvent:] + 120
	9   UIKitCore                           0x000000010c2144bc -[UIControl touchesEnded:withEvent:] + 396
	10  UIKitCore                           0x000000010c894888 -[UIWindow _sendTouchesForEvent:] + 900
	11  UIKitCore                           0x000000010c895d30 -[UIWindow sendEvent:] + 3184
[...]

PR #951, opened 2023-04-21
Fix layout and button click for iOS Counter sample

DylanSale:
Fix? for #930 and #903 
Screenshot 2023-04-21 at 9 19 31 pm

Specifically there are 2 main issues 
1. It seems that internal classes don't get Obj-c classes generated, so they can't be added to views or used as selectors for messages (ie, clicked). So I made the `FlexContainerHostView` and `IosButton` classes public for those reasons.
2. The UIStackView that contained the FlexContainerHostView has NSLayoutConstraints enabled, so the layoutSubviews needs to set the height constraints (or width depending on the direction).

I haven't done iOS dev in about 4 years so please review with caution. :)

Another alternative to implementing the constraints would have been to change the `CounterViewController`'s `container.alignment` to `.fill` which also disables layout constraints and just sets the height to the UIStackView's height. I guess the best alternative depends on the use case. 

There is another issue where the Counter text get's ellipsized when it's too wide, I'm not sure what's causing that though.

### DIFF
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
@@ -122,7 +122,7 @@ public class UIViewFlexContainer(
       } else if (direction.isVertical) {
         NSLayoutConstraint.activateConstraints(listOf(heightConstraint))
       } else {
-        //Unknown??
+      throw AssertionError()
       }
       super.didMoveToSuperview()
     }
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
@@ -16,11 +16,7 @@
 package app.cash.redwood.layout.uiview
 
 import app.cash.redwood.LayoutModifier
-import app.cash.redwood.flexbox.AlignItems
-import app.cash.redwood.flexbox.FlexContainer
-import app.cash.redwood.flexbox.FlexDirection
-import app.cash.redwood.flexbox.JustifyContent
-import app.cash.redwood.flexbox.Size
+import app.cash.redwood.flexbox.*
 import app.cash.redwood.layout.api.Constraint
 import app.cash.redwood.layout.api.CrossAxisAlignment
 import app.cash.redwood.layout.api.Default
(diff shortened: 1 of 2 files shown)


redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -110,7 +107,26 @@ internal class UIViewFlexContainer(
...
   }
 
-  private inner class FlexContainerHostView : UIScrollView(cValue { CGRectZero }) {
+  public inner class FlexContainerHostView : UIScrollView(cValue { CGRectZero }) {
+
+    private var widthContraint: NSLayoutConstraint? = null
+    private var heightConstraint: NSLayoutConstraint? = null
+
+    override fun didMoveToSuperview() {
+      if (superview == null) return
```
JakeWharton: Is it safe not to call super in this case?

I would prefer if we switched to
```kotlin
if (superview != null) {
  // do stuff
}
super.didMoveToSuperview()
```
just to be safe

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -110,7 +107,26 @@ internal class UIViewFlexContainer(
...
+      translatesAutoresizingMaskIntoConstraints = false
+      widthContraint = widthAnchor.constraintEqualToConstant(0.0)
+      heightConstraint = heightAnchor.constraintEqualToConstant(0.0)
+      if(direction.isHorizontal) {
+        NSLayoutConstraint.activateConstraints(listOf(widthContraint))
+      } else if (direction.isVertical) {
+        NSLayoutConstraint.activateConstraints(listOf(heightConstraint))
+      } else {
+        //Unknown??
```
JakeWharton:
```suggestion
      throw AssertionError()
```

JakeWharton (approved):
Please run `spotApply` task to fix a few formatting things (like wildcard imports).

DylanSale:
Sorry for all the pushes, I keep forgetting not to push the cocopods stuff :)

It passes spotApply now.

DylanSale:
I just asked @TwoLivesLeft to look at this and they pointed out that in this case there's no point in using a `UIScrollView` if you are just setting the scrollview's size to the content size because then it will never actually scroll. It only scrolls if the content size is larger than the scrollview's bounds (my interpretation of their words).

I'm not sure what the expected behaviour here should be in that case but I thought I'd raise it as a potential blocker to being merged.

JakeWharton:
Yes the layout implementation is general-purpose and may be used for scrolling views. Today we do not switch the base view type even if you disable scrolling.

DylanSale:
After looking into it more, I think this needs a rework. Nesting Rows and Columns doesn't work at all with the way this is set up. I'll close this for now and make a new PR if I figure out the best approach. Thanks 😄

PR #954, opened 2023-04-21
Apply the schema plugin to fixture tests

JakeWharton:
In preparation to switch to loading only from JSON.


JakeWharton:
Otherwise they just crash!

chrisbanes (approved):

PR #955, opened 2023-04-21
Load schemas only from JSON for codegen

JakeWharton:
Parsing is now entirely isolated to being used for JSON generation and can be switched to PSI!

Refs #19


chrisbanes (approved):

redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/schemaApi.kt
```diff
@@ -124,13 +119,29 @@ public interface ProtocolSchemaSet : SchemaSet {
...
 
 public interface ProtocolSchema : Schema {
   override val widgets: List<ProtocolWidget>
   override val layoutModifiers: List<ProtocolLayoutModifier>
   override val dependencies: List<FqType> get() = taggedDependencies.values.toList()
   public val taggedDependencies: Map<Int, FqType>
+
+  /**
+   * Convert this schema to JSON which can be embedded inside the schema artifact.
+   * This JSON will be read when the schema is used as a dependency.
```
swankjesse, line 142: Supercool

swankjesse (approved):

PR #956, opened 2023-04-21
Build support needs explicit JDK targeting

JakeWharton:
The nested project had one which it inherited from the enclosing gradle module, but the root build-service lacked anything explicit.

### DIFF
--- a/build-support/build.gradle
+++ b/build-support/build.gradle
@@ -1,3 +1,6 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
+
 buildscript {
   dependencies {
     classpath libs.kotlin.gradlePlugin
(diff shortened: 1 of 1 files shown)


build-support/build.gradle
```diff
@@ -47,3 +50,14 @@ buildConfig {
   buildConfigField("String", "composeCompilerArtifactId", "\"${libs.jetbrains.compose.compiler.get().module.name}\"")
   buildConfigField("String", "composeCompilerVersion", "\"${libs.jetbrains.compose.compiler.get().version}\"")
 }
+
+tasks.withType(JavaCompile).configureEach {
```
chrisbanes, line 54: Could use Gradle's toolchain thing? https://docs.gradle.org/7.6/userguide/toolchains.html. The Kotlin plugin automatically pulls in the JVM toolchain value

chrisbanes (approved):

build-support/build.gradle
JakeWharton: Toolchain forces a JDK installation of the exact version. We don't actually care what version of the JDK you run (ideally the latest), but only want to ensure we get builds that target old version (which is also implicitly a minimum-supported JDK for building).

chrisbanes: Ah very true. :shipit:

PR #957, opened 2023-04-21
Build a new AppLifecycle abstraction to host the frame clock

swankjesse:
Previously we sent a frame 60 times per second, whether one was requested or not.

With this change we establish bidirectional communication between the guest and the host, and use that to make a frame clock that only emits frames when they're needed.

This has significant performance benefits, particularly when the application is idle.


jingwei99 (approved):

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineFrameClock.kt
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+import androidx.compose.runtime.BroadcastFrameClock
+import androidx.compose.runtime.MonotonicFrameClock
+
+/** A guest frame clock that asks the host for frames. */
+internal class ZiplineFrameClock(
```
JakeWharton: I think you can delete this whole class and replace it with
```kotlin
val clock = BroadcastFrameClock { host.requestFrame() }
```
at the call site. The lambda is invoked when the number of awaiters (i.e., people who have called `withFrameNanos`) goes from 0 to 1.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppLifecycle.kt
```diff
@@ -18,7 +18,13 @@ package app.cash.redwood.treehouse
...
-public interface FrameClockService : ZiplineService {
+@ObjCName("AppLifecycle", exact = true)
+public interface AppLifecycle : ZiplineService {
+  public fun start(host: Host)
+
   public fun sendFrame(timeNanos: Long)
+
+  public interface Host : ZiplineService {
+    public fun requestFrame()
+  }
```
JakeWharton, line 29: This API is basically `MonotonicFrameClock`. It feels like next we should allow you to supply one of our instances of those rather than driving at 60fps. The implementation on the host-side would then be as simple as:
```kotlin
override fun requestFrame() {
  clock.withFrameNanos { nanos ->
    app.sendFrame(nanos)
  }
}
```

JakeWharton (approved):

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineFrameClock.kt
swankjesse: Oooooh that’s easy.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppLifecycle.kt
swankjesse: Absolutely. Particularly 'cause that’ll plug into iOS and Android differences quite naturally.

PR #959, opened 2023-04-21
Assign a string key to every single item in a LazyColumn

swankjesse:
This fixes issues with items being recreated repeatedly.

I don't like that we need to create lots of strings with this, but it's a simple fix that's likely to prove useful with other features like keeping scroll position.

I've only honored this for Android views. For follow-up it's necessary to honor it for Compose and iOS.

### DIFF
--- a/redwood-treehouse-lazylayout-api/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/api/LazyListIntervalContent.kt
+++ b/redwood-treehouse-lazylayout-api/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/api/LazyListIntervalContent.kt
@@ -24,9 +24,10 @@ import kotlinx.serialization.Serializable
 @Serializable
 @ObjCName("LazyListInterval", exact = true)
 public class LazyListInterval(
-  @Contextual public val count: Int,
+  public val keys: List<String>,
   @Contextual public val itemProvider: Item,
 ) {
+  public val count: Int = keys.size
 
   @ObjCName("LazyListIntervalItem", exact = true)
   public interface Item : ZiplineService {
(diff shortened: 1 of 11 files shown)


veyndan (approved):
I like this! I think this is a good initial step to solve the problem.

Well out of scope of this PR, but just want to give some clarity on what I'm trying to do in a follow-up PR:

From what I can tell, the complete list of `keys` are being serialized every time we load a new page. This'll require a change in how `intervals` are passed, as currently we're passing a snapshot of the current state of things on each page load. We should send a diff down instead.

swankjesse:
Ooooh that’s a fantastic idea

redwood-treehouse-lazylayout-api/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/api/LazyListIntervalContent.kt
```diff
@@ -24,9 +24,10 @@ import kotlinx.serialization.Serializable
 @Serializable
 @ObjCName("LazyListInterval", exact = true)
 public class LazyListInterval(
-  @Contextual public val count: Int,
+  public val keys: List<String>,
```
chrisbanes, line 27: I like the simplicity of `String` but is it better to leave the key type to the user? Tricky with serialization though. Also, having a null key is probably something we'd want to support? 

Compose UI supports this via an `Any?` type, which coaleces to the list index if null.

swankjesse: Yeah, serialization is the hazard with `Any`. There’s probably a way to get kotlinx.serialization to do polymorphic things on `Any?` references. I do like the index fallback thing. Lemme track this.

swankjesse: https://github.com/cashapp/redwood/issues/972

PR #960, opened 2023-04-22
Add and apply logging mismatch handler

jingwei99:


redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolMismatchHandler.kt
```diff
@@ -48,5 +48,23 @@ public interface ProtocolMismatchHandler {
         )
       }
     }
+
+    /** A [ProtocolMismatchHandler] which prints out info for all callbacks. */
+    @JvmField
+    public val Logging: ProtocolMismatchHandler = object : ProtocolMismatchHandler {
```
JakeWharton, line 54: I'm tempted to say that instead of being a companion property this should be a nested class with a signature like
```kotlin
class Logging(logger: (String) -> Unit) : ProtocolMismatchHandler
```
which would allow the `Throwing` instance to be declared as
```
val Throwing = Logging { throw it }
```
and thus de-duplicate the string rendering code.

JakeWharton, line 54: If we're going to add something to the Compose `ProtocolMismatchHandler` we should probably add it to the Widget `ProtocolMismatchHandler` to keep API symmetry.

JakeWharton (approved):

jingwei99:
talked with @swankjesse, there's another direction (forward events to `EventListener`) that is preferred, gonna close this PR, and open a new PR soon.

PR #965, opened 2023-04-24
Remove manual draw invalidation in ViewFlexContainer

chrisbanes:
There's no need to manually call View.invalidate() when triggering a layout. Layout will automatically invalidate if necessary.


veyndan (approved):

JakeWharton (approved):
Needs spotlessApply. I've been meaning to create an action to run it on PRs and push a commit.

chrisbanes:
> I've been meaning to create an action to run it on PRs and push a commit.

That's a neat idea. I'll play around with that today.

chrisbanes:
In fact, that might be tricky as pushing from Actions requires a Personal Access Token. One of us could store one in secrets I guess

JakeWharton:
We have a robot account that we can puppet. I was considering adding its username and a PAT as organization-wide secrets.

PR #973, opened 2023-04-24
Migrate from Truth to AssertK

JakeWharton:
Truth simply does not hold its weight in Kotlin. kotlin.test sometimes just isn't enough. While AssertJ kinda went off the rails (causing our growing usage of Truth) back around Java 8 times, AssertK is really nice and only a mildly-inspired derivative and not a strict port. I've been throwing my weight behind it more and more, so let's use it!


JakeWharton:
As a follow-up I'll probably find/replace all kotlin.test APIs to AssertK. Just wanted to start with Truth.

swankjesse (approved):

PR #974, opened 2023-04-25
Connect protocol mismatch handler to event publisher via AppLifecycle.Host

jingwei99:


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventPublisher.kt
```diff
@@ -16,133 +16,147 @@
...
     listener.appStart(app)
   }
 
-  fun appCanceled(app: TreehouseApp<*>) {
+  public fun appCanceled(app: TreehouseApp<*>) {
     nameToApplication = nameToApplication - app.spec.name
     listener.appCanceled(app)
   }
 
-  fun ziplineEventListener(app: TreehouseApp<*>) = object : ZiplineEventListener() {
```
jingwei99, line 44: these are all formatting changes, I don't like this kind of changes, it messes up the git history/blame

redwood-treehouse-host/build.gradle
```diff
@@ -19,6 +19,7 @@ kotlin {
         api libs.zipline
         api libs.zipline.loader
         api projects.redwoodProtocol
+        api projects.redwoodProtocolCompose
```
JakeWharton: I don't think this dependency is needed. And it would be very advantageous to omit it since it transitively brings in a lot of libraries that are not needed on the host side.

The `EventPublisher` implements the `ProtocolMismatchHandler` from this dependency, but it's invoked from `RealAppLifecycleHost` callbacks which are overrides of `AppLifecycle.Host` and then the implementation only forwards them to `EventListener`. So we basically trampoline through the `ProtocolMismatchHandler` for no reason and it is the only type for which this dependency was needed. It also allocates an instance which implements `ProtocolMismatchHandler` each one of the callbacks are invoked.

Instead, the Compose mismatch handler callbacks should be normal functions on `EventPublisher` which forward to its `EventListener`. This avoids the dependency and avoids the allocations.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -148,6 +150,16 @@ public abstract class EventListener {
   ) {
   }
 
+  /** Invoked on a request to process an unknown event [tag] for the specified widget [widgetTag]. */
+  public open fun onUnknownEvent(widgetTag: WidgetTag, tag: EventTag) {
+    println("Unknown event tag ${tag.value} for widget tag ${widgetTag.value}")
```
JakeWharton: None of the other callbacks log by default. These should probably be empty implementations to match.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventPublisher.kt
```diff
@@ -16,133 +16,147 @@
...
+import app.cash.redwood.protocol.compose.ProtocolMismatchHandler as ComposeProtocolMismatchHandler
 import app.cash.redwood.protocol.widget.ProtocolMismatchHandler
 import app.cash.zipline.Call
 import app.cash.zipline.CallResult
 import app.cash.zipline.EventListener as ZiplineEventListener
 import app.cash.zipline.Zipline
 import app.cash.zipline.ZiplineService
 
-internal class EventPublisher(
+public class EventPublisher(
```
JakeWharton: Why the visibility switch? This seems to only be used by `internal` and `private` types. As far as I can tell this is an internal glue class that's not useful to consumers.

jingwei99: Took a look, this was intermediate, we don't need it for the shape of the code now, switching back!

redwood-treehouse-host/build.gradle
jingwei99: Tracing back my thought process... it's introduced this way, because I was following the pattern of `api projects.redwoodProtocolWidget`

I haven't formed my opinions (too new to the codebase), so definitely gonna take your suggestion.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
jingwei99: I think these are needed

This work started (at least how I know it, @swankjesse probably has a bigger picture story) because Jesse and I were looking at the emoji picker on Activity (in treehouse), and the app crashes because of one of these two events (from the default [protocol mismatch handler](https://github.com/cashapp/redwood/blob/2bb11847d890d2c89b57dd8fa285c4bb30532c54/redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolMismatchHandler.kt#L38)), with this change the same case won't crash anymore, but I also would like to see a log when it happens (instead of being silent)

JakeWharton: Writing to stdout via `println` or any other means is basically a blackhole in production. We have no visibility into these logs and will never see them anywhere but on the emulator. We _must_ override these functions and send the logs to our own logging wrapper (which will eventually forward to the same place stdout goes) or report through some other means in order to have them be visible in the wild.

Conversely, writing to stdout automatically is lightly frowned upon in production because it adds noise to the global logs. In production our logging wrapper does not write to stdout but only captures logs and sends them to the server.

jingwei99: ya right now things are in dev stage, I think we'll need a different plan for prod (like writing to bugsnag/emitting events etc.)

redwood-treehouse-host/build.gradle
JakeWharton: Yeah in general the host should only care about the widgets and the guest should only care about the composables. It's not _quite_ that clean because the composables actually also care about the widgets, but that's a bit of implementation detail and so the host=widget, guest=compose mental model is usually Good Enough™.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
JakeWharton: Definitely. And the way we'll do that is with some other subclass of this event listener specific to our needs.

I guess what I'm saying is that none of the other callbacks in this listener call `println`. So either we should have them all call `println` and the default behavior is to log, or none of them should and the default behavior is to do nothing. I personally lean very heavily towards doing nothing because we can't assume anything about the environment in which the user is running Treehouse.

jingwei99: Also I just paired with @swankjesse and he showed me how Android repo is handling this. Removed!

JakeWharton: Thanks!

JakeWharton (approved):

swankjesse (commented):
Nice! I like where this landed.

PR #976, opened 2023-04-25
Allow nullable keys in LazyList

### LINKED ISSUE #972: Change LazyColumn keys from String to Any?, handle nullable
@chrisbanes, from the PR that added mandatory keys,

> I like the simplicity of String but is it better to leave the key type to the user? Tricky with serialization though. Also, having a null key is probably something we'd want to support?
>
> Compose UI supports this via an Any? type, which coaleces to the list index if null.

Supporting `Any?` is difficult with serialization, and potentially prone to runtime errors if the supplied type is not serializable. We’d need to do something polymorphic and pass the type name in.

Null is significantly easier.

chrisbanes:
Also added support for keys in Compose UI. I don't see a way to use stable keys in UITableView unfortunately.

Follow on from #959. Closes #972

### DIFF
--- a/redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyList.kt
+++ b/redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyList.kt
@@ -60,7 +60,10 @@ internal class ComposeUiLazyList<A : AppService>(
     val itemBoxModifier = if (isVertical) Modifier.height(64.dp) else Modifier.width(64.dp)
     val content: LazyListScope.() -> Unit = {
       intervals.forEach { interval ->
-        items(interval.keys.size) { index ->
+        items(
+          count = interval.keys.size,
+          key = { index -> interval.keys[index] ?: index },
+        ) { index ->
           Box(itemBoxModifier) {
             TreehouseContent(treehouseApp, widgetSystem) { interval.itemProvider.get(index) }
           }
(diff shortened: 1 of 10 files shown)


redwood-treehouse-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/composeui/ComposeUiLazyList.kt
```diff
@@ -60,7 +60,10 @@ internal class ComposeUiLazyList<A : AppService>(
     val itemBoxModifier = if (isVertical) Modifier.height(64.dp) else Modifier.width(64.dp)
     val content: LazyListScope.() -> Unit = {
       intervals.forEach { interval ->
-        items(interval.keys.size) { index ->
+        items(
+          count = interval.keys.size,
+          key = { index -> interval.keys[index] ?: index },
```
chrisbanes, line 65: This was missing from #959. I assume it wasn't intentional?

redwood-treehouse-lazylayout-view/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/view/ViewLazyList.kt
```diff
@@ -118,7 +118,8 @@ internal class ViewLazyList<A : AppService>(
             source = { interval.itemProvider.get(indexInInterval) },
           )
           val keyAndContent = KeyAndContent(
-            key = interval.keys[indexInInterval],
+            // coalesce the key to the item index in the list
```
JakeWharton: We subscribe to the religion of [comments should be sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html).
```suggestion
            // Coalesce the key to the item index in the list.
```

JakeWharton (approved):

veyndan (approved):
Looks good. Merging this will close #972 though, which mentions supporting `Any?`, not just `String?`. Could we edit the description so it doesn't auto-close #972 so we remember to implement support for `Any`?

PR #977, opened 2023-04-25
Try out macos-latest-xl GHA runners

chrisbanes:
https://github.blog/changelog/2023-04-24-github-actions-faster-macos-runners-are-now-available-in-open-public-beta/

Just trying to compare the difference in time a complete run takes vs the `macos-latest` runners.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,7 +15,7 @@ env:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: macos-latest-xl
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3.11.0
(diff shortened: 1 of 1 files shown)


JakeWharton:
Does this invalidate the cache? Probably need a few runs to get a set of good timings.

We should add Gradle buildscan publishing for CI like SQLDelight did.

chrisbanes:
> Does this invalidate the cache? Probably need a few runs to get a set of good timings.

Yeah, I guess it uses the runner type in the cache-key, which is a bit annoying here as they're compatible.

> We should add Gradle buildscan publishing for CI like SQLDelight did.

#978

JakeWharton:
Well that one shard was fast!

chrisbanes:
20% of the time, it works every time.

chrisbanes:
Doesn't look like we have access to the XL runners for some reason. Going to close for this now.

PR #981, opened 2023-04-26
Add `/yarn` PR command to update lock file

JakeWharton:


.github/workflows/yarn.yaml
```diff
@@ -0,0 +1,30 @@
+name: yarn
+
+on:
+  - pull_request
```
chrisbanes: Should we only do this once a PR has merged?

```suggestion
  - pull_request
      types: [closed]
```

chrisbanes (approved):

.github/workflows/yarn.yaml
chrisbanes: Oh actually I see what this is doing. This will break Renovate as it will no longer rebase the change. Probably fine?

JakeWharton: The PR build may fail without the update. Here's an example from last night where I had to append the commit updating the lock file before it could merge: https://github.com/cashapp/turbine/pull/214

JakeWharton: Actually what I really want is to run this if any status in a Renovate PR build fails

JakeWharton: I could make it manually run based on a command like the Spotless one

chrisbanes: > I could make it manually run based on a command like the Spotless one

Yeah that sounds like a good compromise to me!

chrisbanes: Probably need to start documenting these PR commands too 😉

JakeWharton: Switched

JakeWharton:
This is probably a good candidate for local reusable workflows: https://github.blog/changelog/2022-01-25-github-actions-reusable-workflows-can-be-referenced-locally/

But since there's only two I figure copy/paste is good enough for now rather than the overhead of declaring the separate file and inputs and such.

chrisbanes (approved):

PR #983, opened 2023-04-26
Add missing `open` to event listener function

JakeWharton:
Also switch the type from `abstract` to `open` to eliminate the need for a second class just to handle the `NONE` variable. Replace `NONE`'s usage with simple instantiation since this value will generally be single-use within an application and not a widely needed singleton.


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -159,7 +159,7 @@ public abstract class EventListener {
   }
 
   /** Invoked for an event whose node [id] is unknown. */
-  public fun onUnknownEventNode(
+  public open fun onUnknownEventNode(
```
JakeWharton, line 162: The absence of this was producing a build warning since the compiler thought these parameters were unused.

jingwei99 (approved):
🙏

PR #984, opened 2023-04-26
Migrate `kotlin.test` assertions to AssertK

JakeWharton:
The rest of #973. Done 100% using structural search and replace, with manual formatting cleanups.


chrisbanes (approved):
Nice!

PR #985, opened 2023-04-26
Hook up the platform frame clock

swankjesse:
Currently avoiding having a Compose dependency in the host side.


redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/AndroidChoreographerFrameClock.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import android.annotation.SuppressLint
+import android.view.Choreographer
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+/**
+ * A [FrameClock] that suspends with a fixed delay. This does not match the host app's frame pulse.
+ */
+@SuppressLint("NewApi")
```
JakeWharton: What is this triggering on? We should almost never need to suppress this. Otherwise it means we're lying about our minSdk.

samples/emoji-search/ios-uikit/Podfile.lock
```diff
@@ -13,4 +13,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 62788f73308eccd2148887940a388226a1f4c82b
 
-COCOAPODS: 1.12.0
+COCOAPODS: 1.11.3
```
swankjesse: I don’t know how I did this.

JakeWharton: Did you run pod install? I would check if you're cocoapods needs an update (gem update cocoapods, maybe?).

JakeWharton (approved):

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/AndroidChoreographerFrameClock.kt
swankjesse: Android Studio thinks our minSdk is 1. Lemme dig in and figure out why.

JakeWharton: If you haven't, the canary works great and is the only thing I use anymore. Usually a minSdk of 1 means Gradle sync has failed (thus it couldn't read the build file)

swankjesse: That’s it.

> The project is using an incompatible version (AGP 8.0.0) of the Android Gradle plugin. Latest supported version is AGP 8.0.0-beta01
> See Android Studio & AGP compatibility options.

The drawback of aggressive Renovate is my IDE is often out-of-date!

PR #986, opened 2023-04-26
Add serializable view tree class and appropriate method for each widget to generate a view tree

apatronl:


redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/ViewTree.kt
```diff
@@ -0,0 +1,64 @@
...
+import app.cash.redwood.protocol.ChildrenTag
+import app.cash.redwood.protocol.Diff
+import app.cash.redwood.protocol.Id
+import app.cash.redwood.protocol.PropertyDiff
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+
+@Serializable
+public class ViewTree(
+  public val diff: Diff
```
JakeWharton: Do we need a wrapper type for this, or can the type just be `Diff`? The name of the `Diff` type is going to change here soon (because it's not actually a diff), so that'll help it's usage here make more sense. Also when re-hydrating one of these instances `Diff` is the type the rendering layer directly consumes.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/ViewTree.kt
```diff
@@ -0,0 +1,64 @@
...
+) {
+  public class Builder {
```
JakeWharton, line 31: This type is dangerously close to `ProtocolState`. I suspect we should want to converge.

I actually had a plan to make the tester API able to work with the protocol widgets rather than the widget value widgets. The output of such a setup would actually be a `Diff` instance (or multiple over time, if you wanted).

Maybe we can work towards that in some follow-ups.

redwood-compose-testing/src/commonTest/kotlin/app/cash/redwood/compose/testing/WidgetValueTest.kt
```diff
@@ -64,5 +66,9 @@ class WidgetValueTest {
   class SimpleWidgetValue(
     override val layoutModifiers: LayoutModifier = LayoutModifier,
     override val childrenLists: List<List<WidgetValue>> = listOf(),
-  ) : WidgetValue
+  ) : WidgetValue {
+    override fun addTo(parentId: Id, childrenTag: ChildrenTag, builder: ViewTree.Builder) {
+      // Intentionally empty
```
JakeWharton:
```suggestion
      throw AssertionError()
```

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -281,6 +287,23 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
     .add("layoutModifiers,\n")
   val toStringBuilder = StringBuilder()
     .append("${widgetValueType.simpleName}(layoutModifiers=${'$'}layoutModifiers")
+  val childrenDiff = MemberName("app.cash.redwood.protocol", "ChildrenDiff")
+  val addToBuilder = CodeBlock.builder()
+    .addStatement("val widgetId = Id(builder.nextId++)")
```
JakeWharton:
```suggestion
    .addStatement("val widgetId = %T(builder.nextId++)", Protocol.Id)
```

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -281,6 +287,23 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
...
+    .addStatement(
+      "val widgetTag: WidgetTag = %L",
+      buildCodeBlock {
+        val wt = MemberName("app.cash.redwood.protocol", "WidgetTag")
+        if (widget is ProtocolWidget) {
+          add("%M(${widget.tag})", wt)
+        }
+      },
+    )
```
JakeWharton:
```suggestion
    .addStatement("val widgetTag = %T(%L)", Protocol.WidgetTag, (widget as ProtocolWidget).tag)
```
KotlinPoet will throw an exception if the existing `if` block is not taken so switching to an `as` cast will throw earlier in the process and with an exception that indicates what the problem is.

For now, there's _only_ going to be `ProtocolWidget` instances as you can't do a non-protocol schema yet.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -281,6 +287,23 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
...
+    .addStatement(
+      "val childrenDiff = %M.Insert(parentId, childrenTag, widgetId, widgetTag, 0)",
+      childrenDiff,
```
JakeWharton:
```suggestion
      "val childrenDiff = %T(parentId, childrenTag, widgetId, widgetTag, 0)",
      Protocol.ChildrenDiff.nestedClass("Insert"),
```

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -281,6 +287,23 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
     .add("layoutModifiers,\n")
   val toStringBuilder = StringBuilder()
     .append("${widgetValueType.simpleName}(layoutModifiers=${'$'}layoutModifiers")
+  val childrenDiff = MemberName("app.cash.redwood.protocol", "ChildrenDiff")
```
JakeWharton:
```suggestion
```
We should share this through `types.kt`, which you've already done below.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -337,6 +380,14 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
 
   toStringBuilder.append(")")
 
+  addToBuilder
+    .beginControlFlow("for (childrenList in childrenLists)")
+    .addStatement("val nextChildrenTag = childrenTag.value + 1")
+    .beginControlFlow("for (child in childrenList)")
+    .addStatement("child.addTo(widgetId, ChildrenTag(nextChildrenTag), builder)")
```
JakeWharton:
```suggestion
    .addStatement("child.addTo(widgetId, %T(nextChildrenTag), builder)", Protocol.ChildrenTag)
```

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -309,6 +332,26 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
...
+    if (trait is ProtocolWidget.ProtocolProperty) {
+      addToBuilder.addStatement(
+        "builder.propertyDiffs.add(%M(widgetId, %M(%L), builder.json.%M(this.%N)))",
+        propertyDiff,
+        propertyTag,
+        trait.tag,
+        encodeToJsonElement,
+        trait.name,
+      )
+    }
```
JakeWharton:
```suggestion
    addToBuilder.addStatement(
      "builder.propertyDiffs.add(%T(widgetId, %T(%L), builder.json.%M(this.%N)))",
      Protocol.PropertyDiff,
      Protocol.PropertyTag,
      (trait as ProtocolWidget.ProtocolProperty).tag,
      KotlinxSerialization.encodeToJsonElement,
      trait.name,
    )
```

JakeWharton (approved):
This is a great start. Since this is a testing-only API we can evolve it over time to hopefully share a lot more with `ProtocolState` and use it to help fix all the `Diff` and friends naming.

JakeWharton:
Also for codegen, `%T` is for type references (classes, interfaces, etc.) and `%M` is for member references (functions, properties, etc.), usually extensions. It takes a while to get used to how to do it correctly for KotlinPoet but over time it gets easier.

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/ViewTree.kt
swankjesse: I wanted to use the type system to convey that this isn’t just _any_ diff, it’s a diff that assumes an empty initial UI. I think `Diff` is an implementation detail of an idea, and one we could change later.

(I expect this to be the API that end-users see in their tests)

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/ViewTree.kt
swankjesse, line 31: For follow-up: support LayoutModifiers.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -281,6 +287,23 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
...
+      "val widgetTag: WidgetTag = %L",
+      buildCodeBlock {
+        val wt = MemberName("app.cash.redwood.protocol", "WidgetTag")
+        if (widget is ProtocolWidget) {
+          add("%M(${widget.tag})", wt)
+        }
+      },
+    )
+    .addStatement(
+      "val childrenDiff = %M.Insert(parentId, childrenTag, widgetId, widgetTag, 0)",
```
swankjesse: Gotta replace `0` with `builder.childrenDiffs.size`

swankjesse (approved):

redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/ViewTree.kt
apatronl: Added a TODO to track

apatronl: Just looked at `ProtocolState ` and definitely see the similarities, will chat with Jesse on how we can converge the two

Issue #989, opened 2023-04-27
EmojiSearch iOS app crashing on launch


JakeWharton:
Can't find `AppLifecycle`. I closed Xcode and lost the trace, but it was a zipline mismatch exception or whatever they're called.

swankjesse:
I believe this is caused by a stale code in the Zipline cache.

JakeWharton:
Do we just disable the cache then? Rely only on the dev server for now, and eventually the bundled artifacts needed for UI tests.

Issue #537, opened 2022-11-22
UI test for Emoji sample


JakeWharton:
~Blocked on:~
* [x] ~https://github.com/square/okio/issues/1097~
* [x] ~Okio release~

Samples:
* [x] Android Compose UI
* [x] Android views
* [ ] Browser
* [ ] iOS

Probably want to switch build to including an embedded copy of the zipline files so that we don't need to run the server. And we probably want to use an on-screen indicator that the server is not connected (snack bar?) so that you don't forget when doing manual testing.

JakeWharton:
Android ones are now blocked on an Okio release. iOS is not.

We also need to do the work to display a Snackbar (or whatever) when the server is disconnected (but only in debug builds).

JakeWharton:
Code for dynamically creating assets: https://github.com/android/gradle-recipes/blob/0db8094e5a1a9661f61958db85380320157f5050/Kotlin/addCustomAsset/app/build.gradle.kts

We also need to get the presenter-treehouse module publishing the JS as an artifact that we can depend on from another project.

PR #990, opened 2023-04-27
Hook up CADisplayLink for a frame pulse on iOS

swankjesse:


JakeWharton (approved):
Does this fix #989?

JakeWharton:
Hmm maybe the crash only occurs if the devserver isn't running? I'm (slowly) working on making that experience better as part of #537.

PR #992, opened 2023-04-27
Set up basic FIR infrastructure

JakeWharton:
Trying to pull out self-contained bits of this to start landing. Otherwise the final PR would be like 10,000 changed lines.

Refs #19


redwood-tooling-schema/src/test/kotlin/app/cash/redwood/tooling/schema/SchemaParserTest.kt
```diff
@@ -31,18 +31,27 @@ import assertk.assertions.hasMessage
...
 
   interface NonAnnotationSchema
 
   @Test fun nonAnnotatedSchemaThrows() {
     assertFailsWith<IllegalArgumentException> {
-      parseProtocolSchemaSet(NonAnnotationSchema::class)
+      parser.parse(NonAnnotationSchema::class)
     }.hasMessage(
       "Schema app.cash.redwood.tooling.schema.SchemaParserTest.NonAnnotationSchema missing @Schema annotation",
     )
```
JakeWharton, line 57: This is the only test that's working in FIR (in this PR). It means we can successfully look up the schema type and can notice that it's missing annotations.

JakeWharton:
Well well well. This is interesting. So our CLI depends on the schema parser which gained an embeddable compiler dependency but also depends on AGP lint which has its own embeddable compiler dependency (but an older one). And due to the general instability of the compiler's API, any version deviation breaks the other.

I think for now I'll just match lint's embeddable compiler version. Long term we probably want to eliminate the centralized CLI and instead have each tooling module provide its own CLI. This would have the parser exposing a JSON-producing command. The lint tooling would expose the lint command and the API merge command. And the code generator exposing the generate command.

This is ideal because the Gradle plugin loads these in individual classpaths and forks a JVM to execute each tool. Thus, they would not longer interfere with each other.

I'll do that next since it's pretty easy.

JakeWharton:
Nope. Won't work. Versions are too far apart. Doing the other thing first.

JakeWharton:
Might need to change the Lint worker isolation to prevent the dependency clash. We'll see.

redwood-tooling-schema/build.gradle
```diff
@@ -12,11 +17,26 @@ dependencies {
...
 
   testImplementation projects.testSchema
   testImplementation libs.junit
   testImplementation libs.assertk
+  testImplementation libs.testParameterInjector
+}
+
+// In order to simplify writing test schemas, inject the test sources and
+// test classpath as properties into the test runtime. This allows testing
+// the FIR-based parser on sources written inside the test case. Cool!
```
swankjesse, line 31: COOL!

redwood-tooling-schema/src/test/kotlin/app/cash/redwood/tooling/schema/SchemaParserTest.kt
swankjesse: 👍🏻

swankjesse (approved):

Issue #998, opened 2023-05-01
`LayoutModifier`s do not work outside of protocol usage


JakeWharton:
The only call to `onLayoutModifierUpdated` in the codebase is in `ProtocolBridge`.

Thankfully(?) I've been noodling on redesigning how notification to the parent occurs in order to fix #785 via a protocol change, so this issue will track that change which will be done first.

This may also be the source of some of the bugs in the counter sample.

PR #1003, opened 2023-05-03
Invalidate all children on a modifier change

JakeWharton:
This is the lead up to a change where the child will notify the enclosing children container directly of a modifier change.

Refs #998

I'm not super happy with this entire mechanism. There's clear deficiencies in how the Compose UI and HTML parts work. This change and another that will follow it are solely to unblock some other work first. Then we can circle back on cleaning up the relationship between children and their containers.

### DIFF
--- a/redwood-widget-compose/src/commonMain/kotlin/app/cash/redwood/widget/compose/ComposeWidgetChildren.kt
+++ b/redwood-widget-compose/src/commonMain/kotlin/app/cash/redwood/widget/compose/ComposeWidgetChildren.kt
@@ -16,41 +16,41 @@
 package app.cash.redwood.widget.compose
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateListOf
-import app.cash.redwood.LayoutModifier
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
 import app.cash.redwood.widget.Widget
 
 public class ComposeWidgetChildren : Widget.Children<@Composable () -> Unit> {
-  private val layoutModifiers = mutableStateListOf<LayoutModifier>()
+  private var layoutModifierTick by mutableStateOf(0)
 
   private val _widgets = mutableStateListOf<Widget<@Composable () -> Unit>>()
   public val widgets: List<Widget<@Composable () -> Unit>> get() = _widgets
 
   @Composable
   public fun render() {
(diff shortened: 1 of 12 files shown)


redwood-widget-compose/src/commonMain/kotlin/app/cash/redwood/widget/compose/ComposeWidgetChildren.kt
```diff
@@ -16,41 +16,41 @@
...
   @Composable
   public fun render() {
+    // Observe the layout modifier count so we recompose if it changes.
+    layoutModifierTick
+
     for (index in _widgets.indices) {
-      // Observe the layout modifier so we recompose if it changes.
-      layoutModifiers[index]
       _widgets[index].value()
     }
```
JakeWharton, line 38: Should have the same behavior. We dirty the nodes produced by this loop which forces the parent into which these nodes are placed to relayout. Would love to figure out how to replace this implicit relationship with one that's more explicit.

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/HTMLElementChildren.kt
```diff
@@ -61,9 +61,11 @@ public class HTMLElementChildren(
...
 
-  override fun onLayoutModifierUpdated(index: Int) {
-    val element = parent.children[index] as HTMLElement
+  override fun onLayoutModifierUpdated() {
+    // If this function is being invoked we are guaranteed to have at least one child.
+
+    val element = parent.children[0] as HTMLElement
     parent.removeChild(element)
-    parent.insertBefore(element, parent.children[index])
+    parent.insertBefore(element, parent.children[0])
```
JakeWharton, line 69: Again, same behavior. We dirty the container and force it to relayout. However, unlike other platforms, we don't get a hook into layout and rely on built-in flexbox. I'm pretty sure HTML is just broken as the parent needs to be propagating modifiers into HTML attributes. But if it's broken it stays broken with this change so...

redwood-widget-compose/src/commonMain/kotlin/app/cash/redwood/widget/compose/ComposeWidgetChildren.kt
chrisbanes: Tricky without making a `ComposeableWidget` class which produces snapshot changes.

chrisbanes (approved):

Issue #1005, opened 2023-05-03
Add support for horizontal scrolling in `UIViewLazyList`


veyndan:
It'll require converting the implementation of `UIViewLazyList` from `UITableView` to `UICollectionView`, as `UITableView` doesn't support horizontal scrolling.

PR #1006, opened 2023-05-03
Remove Zipline dependency from LazyList

### LINKED ISSUE #605: `LazyColumn` usage outside of Treehouse

veyndan:
#### Conceptually

A list of "iframes" that are requested when the list view is about to display them.

#### Steps

1. The guest sends down a list of intervals. Each interval has an item count and a function to produce a `ZiplineTreehouseUi` given an index.
2. The host list view buffers a list of `ZiplineTreehouseUi` to be displayed synchronously. The buffering is done via an internal `Pager`, which ensures that the range of indices to preload is sane given the currently displayed items.

#### Issues

1. Has a hard dependency on Zipline.
    1. Unnecessary for web (https://github.com/cashapp/redwood/issues/605)
    2. Proper closure of Zipline services is difficult.
2. The buffering mechanism is complicated, and it being on the host means updating it requires an app release.

#### Conceptually

A window of views that shifts depending on the visible items on screen. The window's size is at least a few viewports worth.

#### Steps

1. The guest sends down the first X items as some `WidgetChildren`.
2. The host pings the guest every time a widget is displayed with the position to be displayed.
3. The guest tells its buffering mechanism (i.e., the `Pager`) that a `position` was accessed. All the widgets within the [`PagingConfig#prefetchDistance`](https://developer.android.com/reference/kotlin/androidx/paging/PagingConfig#prefetchDistance()) of the `position` are buffered.
4. The guest sends down updates to the `WidgetChildren`.

#### Resolved Issues

1. No dependency on Zipline. Fixes https://github.com/cashapp/redwood/issues/605.
2. The buffering mechanism has been hoisted to guest, making it easy to update.

#### Persistent Issues

1. When `itemCount` dynamically updates (e.g., when using `collectAsLazyPagingItems`), the items glitch out in new ways. This shouldn't block merging, as dynamically updating `itemCount` was previously broken too.

## Blocking

- [x] https://github.com/cashapp/multiplatform-paging/issues/10
    - Allows MacOS support to be added back in this PR.
- [x] https://github.com/cashapp/redwood/issues/1005
    - CI should fail to compile on iOS. `UICollectionView` is more flexible than `UITableView`, and I'd rather translate it now while there's not much to change vs having to redo the implementation later.


redwood-treehouse-lazylayout-schema/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/widgets.kt
```diff
@@ -15,12 +15,13 @@
...
 import app.cash.redwood.schema.Property
 import app.cash.redwood.schema.Widget
-import app.cash.redwood.treehouse.lazylayout.api.LazyListInterval
 
 @Widget(1)
 public data class LazyList(
   @Property(1) val isVertical: Boolean,
-  @Property(2) val intervals: List<LazyListInterval>,
+  @Property(2) val onPositionDisplayed: (position: Int) -> Unit,
+  @Children(3) val items: () -> Unit,
```
JakeWharton:
```suggestion
  @Children(1) val items: () -> Unit,
```
Occupies a different numberspace than properties

samples/emoji-search/presenter/src/jsMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearchTreehouseUi.kt
```diff
@@ -16,32 +16,28 @@
...
   @Composable
   override fun Show() {
     EmojiSearch(httpClient, lazyColumnProvider)
   }
 }
 
-private class LazyColumnProvider(
-  private val appLifecycle: StandardAppLifecycle,
-) : ColumnProvider {
+private class LazyColumnProvider : ColumnProvider {
```
JakeWharton, line 34: If the Zipline dependency is fully broken we can eliminate this entire `ColumnProvider` abstraction. Can be a follow-up.

redwood-treehouse-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/compose/LazyList.kt
```diff
@@ -18,16 +18,55 @@
...
+    // TODO Don't hardcode pageSizes
+    // TODO Enable placeholder support
+    // TODO Set a maxSize so we don't keep _too_ many views in memory
+    val pager = Pager(PagingConfig(pageSize = 20, initialLoadSize = 20, enablePlaceholders = false)) {
+      itemPagingSource!!
+    }
+    pager.flow
+  }
+  val lazyPagingItems = pagerFlow.collectAsLazyPagingItems()
+  DisposableEffect(scope) {
```
JakeWharton, line 48: Does `scope` need to be remembered? When I read this I think that the `scope` value will be recreated each recomposition and so this will also be invalidated/recreated.

redwood-treehouse-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/compose/LazyList.kt
```diff
@@ -18,16 +18,55 @@
...
+    itemPagingSource = ItemPagingSource(scope)
+    onDispose {
+      itemPagingSource?.invalidate()
+    }
+  }
+  LazyList(
+    isVertical,
+    onPositionDisplayed = { position ->
+      /** Triggers load at position, loading all items within [PagingConfig.prefetchDistance] of the [position]. */
```
JakeWharton, line 57:
```suggestion
      // Triggers load at position, loading all items within [PagingConfig.prefetchDistance] of the [position].
```
Not quite sure why this is a KDoc-style comment. Maybe it originally lived on a function or something?

redwood-treehouse-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/compose/LazyList.kt
```diff
@@ -18,16 +18,55 @@
...
+      try {
+        lazyPagingItems[position]
+      } catch (ignore: IndexOutOfBoundsException) {
+        // Ignore potential race condition where position exceeds available items.
+        // TODO This clause isn't invoked on IndexOutOfBoundsException. Why?
+      }
+    },
+    items = {
+      for (index in (0 until lazyPagingItems.itemCount)) {
+        /** Only invokes Composable lambdas that are loaded. */
```
JakeWharton:
```suggestion
        // Only invokes Composable lambdas that are loaded.
```

samples/emoji-search/presenter/src/jsMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearchTreehouseUi.kt
JakeWharton: Also for follow-up: drop 'treehouse' from package name / artifact name!

JakeWharton (commented):
Excited for this

josereyes:
re this blocker: https://github.com/cashapp/redwood/issues/1005

Would it help if we punted / delayed this for a later version? Currently we don't need to actually support this in any of the shared surfaces and we're moving away from horizontal scrolling in general.

I still think we need to add support for this but it's not time sensitive.

veyndan:
@josereyes I just added this as a blocker as I want/need to interact with `UICollectionView` instead of `UITableVIew` for some changes that need to be done in this PR. I'm expecting the completion of that issue to take like 10mins, so IMO it's quicker to just complete than have to re-prioritise it.

redwood-treehouse-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/compose/LazyList.kt
veyndan: I wanted to have the ability to link to things, i.e., `[PagingConfig.prefetchDistance]` and `[position]`, and regular comments don't support it.

swankjesse:
FYI I rebased and force pushed!

swankjesse:
simulator_screenshot_1056EF9D-9D0C-4374-93F0-CDED60CCACE9

redwood-treehouse-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -18,92 +18,136 @@ package app.cash.redwood.treehouse.lazylayout.uiview
...
+    }
+
+    override fun move(fromIndex: Int, toIndex: Int, count: Int) {
+      TODO("Not yet implemented")
+    }
+
+    override fun remove(index: Int, count: Int) {
+      for (i in 0 until count) {
+        itemsList.removeAt(index)
+      }
```
JakeWharton: Add the `ComposeHelpers` to your source set in the build file and you'll get efficient implementations of `move` and `remove` which operate on a `MutableList<T>`.

redwood-treehouse-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -18,92 +18,136 @@ package app.cash.redwood.treehouse.lazylayout.uiview
...
+      delegate = tableViewDelegate
+      rowHeight = 64.0 // TODO: size rows by their content.
+      prefetchingEnabled = false
+      registerClass(
+        Cell(UITableViewCellStyleDefault, reuseIdentifier).classForCoder() as ObjCClass?,
+        forCellReuseIdentifier = reuseIdentifier,
+      )
+    }
+
+  private val root = tableView
```
JakeWharton: Inline?

redwood-treehouse-lazylayout-view/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/view/ViewLazyList.kt
```diff
@@ -77,124 +90,31 @@ internal class ViewLazyList<A : AppService>(
...
-  ) {
-    suspend fun awaitContent() {
-      // TODO Create a public API that accepts an Android Context and returns the corresponding HostConfiguration
-      content.preload(HostConfiguration(darkMode = true))
-      content.awaitContent()
-    }
-
-    override fun equals(other: Any?) = other is KeyAndContent && key == other.key
+  private class LazyContentItemListAdapter : RecyclerView.Adapter<ViewHolder>() {
+    var onPositionDisplayed: ((Int) -> Unit)? = null
```
JakeWharton, line 98: lateinit?

redwood-treehouse-lazylayout-uiview/build.gradle
```diff
@@ -20,3 +20,11 @@ kotlin {
     }
   }
 }
+
+// TODO: find a way to get Objective-C conflicting overloads to play nice with Kotlin metadata.
+// https://kotlinlang.slack.com/archives/C9JM6Q2UX/p1683686935270659
+tasks.configureEach { task ->
+  if (task.name == "compileCommonMainKotlinMetadata") {
+    task.onlyIf { false }
```
JakeWharton: We used to have to do this a long, long time ago in this repo (back when we used `includeBuild` to split the library + samples) and it caused tons of problems with publishing. This might be okay because it's a leaf module in terms of publishing (no other published module depends on it).

redwood-treehouse-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -18,92 +18,136 @@ package app.cash.redwood.treehouse.lazylayout.uiview
...
+import platform.darwin.NSObject
 
 internal class UIViewLazyList<A : AppService>(
   treehouseApp: TreehouseApp<A>,
   widgetSystem: WidgetSystem,
 ) : LazyList<UIView> {
-  private val dataSource = TableViewDataSource(treehouseApp, widgetSystem)
-  private val root = UITableView().apply {
-    this.dataSource = this@UIViewLazyList.dataSource
+  private val reuseIdentifier = "cell"
```
JakeWharton:
```suggestion
  private val reuseIdentifier get() = "cell"
```
Or `const` at top-level in file?

JakeWharton (approved):

swankjesse (approved):

redwood-treehouse-lazylayout-uiview/build.gradle
veyndan: https://github.com/cashapp/redwood/pull/1006/commits/1f92af9d605174af886dc9fbdb91c088346beae5

redwood-treehouse-lazylayout-view/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/view/ViewLazyList.kt
veyndan: I was unsure with this one, as technically the `onPositionDisplayed` function could be invoked multiple times.

redwood-treehouse-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -13,97 +13,141 @@
...
+  private val itemsList = mutableListOf<Widget<UIView>>()
+
+  override val items: Widget.Children<UIView> = object : Widget.Children<UIView> {
+    override fun insert(index: Int, widget: Widget<UIView>) {
+      itemsList.add(index, widget)
+      tableView.reloadData()
+    }
+
+    override fun move(fromIndex: Int, toIndex: Int, count: Int) {
+      TODO("Not yet implemented")
```
JakeWharton:
```suggestion
      itemsList.move(fromIndex, toIndex, count)
      tableView.reloadData()
```

redwood-treehouse-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -13,97 +13,141 @@
...
-  }
-}
 
-private class TableViewCell(private val view: UIView) :
-  UITableViewCell(UITableViewCellStyle.UITableViewCellStyleDefault, null) {
-  init {
-    addSubview(view)
+  fun setWidget(widget: UIView) {
+    contentView.addSubview(widget)
+    this.widget = widget
```
JakeWharton: The use of "widget" here makes me think it's a `Widget` but it's a `UIView`. Can we call it "view" or "child" or something?

redwood-treehouse-lazylayout-view/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/view/ViewLazyList.kt
```diff
@@ -15,47 +15,60 @@
...
+
+  override fun insert(index: Int, widget: Widget<View>) {
+    _widgets.insert(index, widget)
+    adapter.notifyItemInserted(index)
+  }
+
+  override fun move(fromIndex: Int, toIndex: Int, count: Int) {
+    _widgets.move(fromIndex, toIndex, count)
+    check(count == 1)
+    adapter.notifyItemMoved(fromIndex, toIndex) // TODO Support arbitrary count.
```
JakeWharton, line 46: Probably have to notify a range change on >1. I _think_ it would be something like:
```kotlin
val lowerBound = minOf(fromIndex, toIndex)
val upperBound = maxOf(fromIndex, toIndex) + count
```

JakeWharton (approved):
🥳

redwood-treehouse-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/treehouse/lazylayout/compose/LazyList.kt
veyndan: We might need to? I'll probably do it in a follow up either way, just so I can move this PR out the door.

redwood-treehouse-lazylayout-view/src/main/kotlin/app/cash/redwood/treehouse/lazylayout/view/ViewLazyList.kt
veyndan: I'm not too sure what the `notify*` method I should be calling. Ideally I'd want something like `adapter.notifyItemRangeMoved(fromPosition, toPosition, count)`, but no such method exists. Frustratingly, stepping through the implementation of `RecyclerView.Adapter#notifyItemMoved` shows that it's invoking `RecyclerView.AdapterDataObserver#onItemRangeMoved(fromPosition, toPosition, itemCount)` with `itemCount` hardcoded to 1.

JakeWharton: Hah! File a feature request / send a PR?

veyndan: [Filed!](https://issuetracker.google.com/issues/281929801)

PR #1009, opened 2023-05-03
Move the widget children tester to its own module

JakeWharton:
This allows us to consume it in the tests for the main implementations but also the Compose one. Someone could also consume it for testing a third-party implementation, too.

### DIFF
--- a/redwood-widget-compose/build.gradle
+++ b/redwood-widget-compose/build.gradle
@@ -17,5 +17,11 @@ kotlin {
         api libs.jetbrains.compose.runtime
       }
     }
+    commonTest {
+      dependencies {
+        implementation libs.kotlinx.coroutines.core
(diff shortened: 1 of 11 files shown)


redwood-widget-compose/build.gradle
```diff
@@ -17,5 +17,11 @@ kotlin {
         api libs.jetbrains.compose.runtime
       }
     }
+    commonTest {
+      dependencies {
+        implementation libs.kotlinx.coroutines.core
```
JakeWharton, line 22: We shouldn't need this. Seems like Compose multiplatform isn't correctly exposing coroutines as an `api` dependency. I may look into it... or I may not. Wasn't worth a "TODO".

chrisbanes (approved):

PR #1015, opened 2023-05-04
Add view tree renderer

apatronl:


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/types.kt
```diff
@@ -70,7 +70,7 @@ internal object Redwood {
 internal object RedwoodTesting {
   val MutableWidget = ClassName("app.cash.redwood.compose.testing", "MutableWidget")
   val RedwoodTester = ClassName("app.cash.redwood.compose.testing", "RedwoodTester")
-  val ViewTree = ClassName("app.cash.redwood.compose.testing", "ViewTree")
+  val ViewTree = ClassName("app.cash.redwood.protocol", "ViewTree")
```
JakeWharton: Can you move this into `RedwoodProtocol` object to maintain the association with the module the type comes from.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/ViewTreeRenderer.kt
```diff
@@ -0,0 +1,55 @@
...
+public class ViewTreeRenderer<W : Any>(
+  private val json: Json,
+) {
+  private val refuseAllEvents = EventSink { event ->
+    throw IllegalStateException("unexpected event: $event")
+  }
+
+  @Suppress("UNCHECKED_CAST")
+  public fun render(
+    view: TreehouseView,
```
JakeWharton, line 41: I think eventually we want this to live in 'redwood-protocol-widget' and run directly on a `Widget.Children`, but landing it in here is a good starting location.

JakeWharton (approved):

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/types.kt
swankjesse: This confused me 'cause I was thinking you were talking about moving the ViewTree class somewhere else, but then I realized you just wanna move this line to follow line 57 in this same file.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/ViewTreeRenderer.kt
swankjesse: (If so, we’ll need a passed-in WidgetSystem. That’s a good trade-off in modularity though.)

swankjesse (approved):

PR #1017, opened 2023-05-04
Change our tree to be custom nodes instead of raw widgets

### LINKED ISSUE #998: `LayoutModifier`s do not work outside of protocol usage

JakeWharton:
We already were faking half of the tree with the synthetic children widgets before. Now we keep children nodes which do the same thing as before but are joined by widget nodes which wrap a widget and provide access to the parent children. This is used to notify the parent children when any child's layout modifiers have changed.

Closes #998

### DIFF
--- a/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
+++ b/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
@@ -17,40 +17,45 @@ package app.cash.redwood.compose
 
 import androidx.compose.runtime.AbstractApplier
 import androidx.compose.runtime.Applier
+import app.cash.redwood.LayoutModifier
 import app.cash.redwood.RedwoodCodegenApi
 import app.cash.redwood.widget.Widget
 
 /**
- * An [Applier] for a tree of [Widget]s.
+ * An [Applier] for Redwood's tree of nodes.
  *
- * This applier has special handling for emulating nodes which contain multiple children. Nodes in
- * the tree are required to alternate between [ChildrenWidget] instances and user [Widget] subtypes
- * starting at the root. This invariant is maintained by virtue of the fact that all of the input
- * `@Composables` should be generated Redwood code.
+ * This applier has special handling for emulating widgets which contain multiple children.
+ * Nodes in the tree are required to alternate between [WidgetNode] and [ChildrenNode] starting
+ * from the root. This invariant is maintained by virtue of the fact that all of the input
+ * `@Composable`s should bottom out in generated Redwood code.
  *
  * For example, a widget tree may look like this:
  * ```
- *                    Children(tag=1)
- *                     /          \
- *                    /            \
- *            ToolbarNode        ListNode
- *             ·     ·                 ·
- *            ·       ·                 ·
- * Children(tag=1)  Children(tag=2)   Children(tag=1)
- *        |              |               /       \
- *        |              |              /         \
- *   ButtonNode     ButtonNode     TextNode     TextNode
+ *                        ChildrenNode(root)
+ *                          /           \
+ *                         /             \
+ *          WidgetNode<Toolbar>       WidgetNode<List>
+ *             /          \                      \
+ *            /            \                      \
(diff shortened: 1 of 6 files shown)


redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
```diff
@@ -59,37 +64,102 @@ internal class WidgetApplier<W : Any>(
...
+/** @suppress For generated code usage only. */
+@RedwoodCodegenApi
+public sealed interface Node<W : Any>
+
+/**
+ * A node which wraps a [Widget] and also holds onto the [Widget.Children] in which the widget
+ * is placed. The generics of this class are a little funky in order to avoid casts of the widget
+ * in the generated composables.
+ *
+ * @suppress For generated code usage only.
```
swankjesse, line 116: Why `@suppress` ?

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
```diff
@@ -59,37 +64,102 @@ internal class WidgetApplier<W : Any>(
...
+  public companion object {
+    public val SetLayoutModifiers: WidgetNode<*, *>.(LayoutModifier) -> Unit = {
+      widget.layoutModifiers = it
+      container?.onLayoutModifierUpdated()
+    }
+  }
+}
+
+/**
+ * A synthetic widget which allows the applier to differentiate between multiple groups of children.
```
swankjesse, line 133: These descriptions are really helpful

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
```diff
@@ -59,37 +64,102 @@ internal class WidgetApplier<W : Any>(
...
+ *
+ * Compose's tree assumes each node only has single list of children. Or, put another way, even if
+ * you apply multiple children Compose treats them as a single list of child nodes. In order to
+ * differentiate between these children lists we introduce synthetic nodes. Every real node which
+ * supports one or more groups of children will have one or more of these synthetic nodes as its
+ * direct descendants. The nodes which are produced by each group of children will then become the
+ * descendants of those synthetic nodes.
+ *
+ * This node has two valid states:
```
swankjesse, line 142: This is a tricky process, but the description makes it nice & clear

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
```diff
@@ -59,37 +64,102 @@ internal class WidgetApplier<W : Any>(
...
+  private var _children: Widget.Children<W>?,
+) : Node<W> {
+  constructor(accessor: (Widget<W>) -> Widget.Children<W>) : this(accessor, null)
+  constructor(children: Widget.Children<W>) : this(null, children)
+
+  val children: Widget.Children<W> get() = _children!!
+
+  fun attachTo(parent: Widget<W>) {
+    _children = checkNotNull(_accessor).invoke(parent)
+    _accessor = null
```
swankjesse: Does accessor need `_` ?

swankjesse (approved):

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
JakeWharton: Hides it from docs. No one should ever touch these things, but they're public because generated composable implementation must.

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
JakeWharton: Nope! Forgot I added the function to attach rather than doing it at the call site.

PR #1020, opened 2023-05-05
Restore removing removed nodes from widget-side of protocol

JakeWharton:
This was lost as part of #1017. Add a test for it.


chrisbanes (approved):

redwood-protocol-widget/src/commonTest/kotlin/app/cash/redwood/protocol/widget/EmptyExampleSchemaWidgetFactory.kt
```diff
@@ -34,7 +34,7 @@ open class EmptyExampleSchemaWidgetFactory : ExampleSchemaWidgetFactory<Nothing>
       get() = TODO()
       set(_) { TODO() }
 
-    override fun text(text: String?) = TODO()
+    override fun text(text: String?) {}
```
JakeWharton:
```suggestion
    override fun text(text: String?) = TODO()
```

Issue #1031, opened 2023-05-08
`ViewTree` API incorrect child index calculation


JakeWharton:
The generated code of the tester assumes the children diffs size is a valid insertion index, but that immediately breaks with two levels of hierarchy or two sets of children.

JakeWharton:
I have a fix and tests for this done, but #1104 makes them better so landing that first.

PR #1032, opened 2023-05-08
Rewrite the protocol to be more precise

### LINKED ISSUE #785: Apply initial properties before adding widget to parent
In `ProtocolBridge` we currently process things in the following order:
1. Create widget
2. Add widget to parent
3. Set layout modifiers
4. Set properties

Instead, we should do:
1. Create widget
2. Set layout modifiers
3. Set properties
4. Add widget to parent

This ensures the parent never sees a widget until all of its properties are set. Currently it's Probably Fine™, because the act of adding a widget to a parent always schedules a layout pass on the next frame for every UI toolkit we support, but it's still possible that the parent could synchronously query things on its child within the same stackframe as when it was added.

This change is possible with the current setup by simply deferring a set of widget-to-parent adds in a list, but perhaps we can do better than that. The current order is a function of how we've modeled the `Diff` type in the protocol. We batch children ops (create, move, delete) separately from property ops (layout modifiers, property diffs) and process them in separate loops.

If the `Diff` was instead modeled as groups of operations based around a single node we could move this logic away from the UI-side `ProtocolBridge` and into the Compose-side `ProtocolState`.

### LINKED ISSUE #804: Switch tester tests to black box
I was worried they would be annoying, and they immediately were. We can test it through the public API.

JakeWharton:
The new protocol has two features not present in the old one:
- Creation of a node is separate from attaching it to a parent node's children.
- Mutations of properties, layout modifiers, or the tree structure are recorded precisely in the order they occur by Compose

This change was primarily motivated by the desire to have nodes on the widget-side be created, have all their initial properties set, and then be attached to their parents.

As part of this change there is also a slight change in the order of widget attachment in multi-level branches of the tree. Prior to this change, given

```kotlin
Row {
  Column {
    Text("Hi")
```

We would see `Column` created and attached to `Row`, and then `Text` created and attached to `Column`. With this change, however, `Text` is attached to `Column` before `Column` is attached to `Row`.

Basically, when a new branch is attached you can assume it is fully formed structurally (as well as having initial properties set).

Closes #785. Closes #804.


redwood-protocol-compose/src/commonTest/kotlin/app/cash/redwood/protocol/compose/ProtocolTest.kt
```diff
@@ -80,38 +80,36 @@ class ProtocolTest {
...
+        PropertyChange(Id(2), PropertyTag(1) /* text */, JsonPrimitive("hey")),
+        ChildrenChange.Add(Id(1), ChildrenTag(1), Id(2), 0),
+        Create(Id(3), WidgetTag(1) /* row */),
+        LayoutModifierChange(Id(3)),
+        Create(Id(4), WidgetTag(3) /* text */),
+        LayoutModifierChange(Id(4)),
+        PropertyChange(Id(4), PropertyTag(1) /* text */, JsonPrimitive("hello")),
+        ChildrenChange.Add(Id(3), ChildrenTag(1), Id(4), 0),
+        ChildrenChange.Add(Id(1), ChildrenTag(1), Id(3), 1),
+        ChildrenChange.Add(Id.Root, ChildrenTag.Root, Id(1), 0),
```
JakeWharton, line 98: This is probably the best view of both the properties being applied before the widget is added to its parent, but also how the tree is attached from the bottom-up now.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -287,18 +287,26 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
...
+    )
+    .addStatement(
+      """
+      |builder.changes += %T(
       |  parentId,
       |  childrenTag,
       |  widgetId,
-      |  widgetTag,
-      |  builder.childrenDiffs.size
+      |  builder.changes.size,
```
JakeWharton, line 304: This previously-broken assumption is now super broken after this PR.

https://github.com/cashapp/redwood/issues/1031

chrisbanes (approved):

Issue #1041, opened 2023-05-10
Revamp the widget-side codegen to better facilitate the protocol and testing


JakeWharton:
This is going to be somewhat of a brain dump of a bunch of things bouncing around in my head. There are or were individual issues for some of the points in here, but this will basically track the whole revamp which will need many PRs to complete.

High-level goals is to make the protocol usage for Treehouse and the tester have better UX as well as correctness. The need for Treehouse to have a `WidgetSystem` type and the current design of the tester API (even after #806 was fixed) are failures of good codegen design.

I think the new shape is going to be
```kotlin
public object EmojiSearchWidgets {
  fun <W : Any> provider(
    EmojiSearch: EmojiSearchWidgetFactory<W>,
    RedwoodLayout: RedwoodLayoutWidgetFactory<W>,
    RedwoodTreehouseLazyLayout: RedwoodTreehouseLazyLayoutWidgetFactory<W>,
  ) : EmojiSearchWidgetFactoryProvider<W> {
    return EmojiSearchWidgetFactories(EmojiSearch, RedwoodLayout, RedwoodTreehouseLazyLayout)
  }
}

private class EmojiSearchWidgetFactories<W : Any>(
  public override val EmojiSearch: EmojiSearchWidgetFactory<W>,
  public override val RedwoodLayout: RedwoodLayoutWidgetFactory<W>,
  public override val RedwoodTreehouseLazyLayout: RedwoodTreehouseLazyLayoutWidgetFactory<W>,
) : EmojiSearchWidgetFactoryProvider<W>
```
which is similar to what we have today, except we have a `object` off which to hang other niceties.

Such as: testing!
```kotlin
suspend fun EmojiSearchWidgets.test(body: TestScope.() -> Unit) {
  val provider = provider(
    EmojiSearch = MutableEmojiSearchWidgetFactory(),
    RedwoodLayout = MutableRedwoodLayoutWidgetFactory(),
    RedwoodTreehouseLazyLayout = MutableRedwoodTreehouseLazyLayoutWidgetFactory(),
  )

  blah blah..
}
```
This should also come with the continued revamp to the tester API. We don't need the `MutableWidget` type as we can use `Widget<WidgetValue>` maybe we rename that to `Widget<Snapshot>` or something.

And probably the protocol becomes an extension on `EmojiSearchWidgetFactoryProvider` like
```kotlin
fun EmojiSearchWidgetFactoryProvider<*>.protcolSomethingSomething(
[...]

PR #1042, opened 2023-05-10
Eliminate `MutableWidget` type

JakeWharton:
We can use the `value` property of a widget to snapshot the value.

Refs #1041


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/sharedHelpers.kt
```diff
@@ -101,7 +101,7 @@ internal fun Schema.getWidgetFactoryType(): ClassName {
 }
 
 internal fun Schema.getMutableWidgetFactoryType(): ClassName {
-  return ClassName(widgetPackage(), "Mutable${type.flatName}WidgetFactory")
+  return ClassName(widgetPackage(), "${type.flatName}MutableWidgetFactory")
```
JakeWharton, line 104: This reads gooder. "Mutable" is operating on "Widget", not the whole factory. Need to make the same change for the protocol factory separately.

chrisbanes (approved):

Issue #806, opened 2023-02-08
Tester API needs some tweaking


JakeWharton:
The type accepts a `CoroutineScope` but then the `test` API is a suspend function that isn't used to drive the composition but will cancel it when it completes. The `test` function needs to live outside the tester so that we can create a coroutine scope from its context and cancel when the function ends.

PR #1051, opened 2023-05-11
Bring the test composition API more in line with regular and protocol composition

### LINKED ISSUE #805: See if tester API can be used with protocol widgets
This is a weird thought, but it would be nice to be able to use it to await diffs for our own tests.

JakeWharton:
This also generalizes it so that we can use it for testing the protocol.

Closes #805. Refs #806.


chrisbanes (approved):
Nice

PR #1053, opened 2023-05-11
Fix JS build by specifying unique module names

### LINKED ISSUE #988: Publishing broken on `trunk`
Gradle 8 failing on broken task dependency between webpack and compilation.

Example: https://github.com/cashapp/redwood/actions/runs/4815788034/jobs/8575162000

JakeWharton:
By default these JS modules inherit the project name which was either 'browser' or 'presenter-treehouse', except those names are used twice because we have multiple samples with the same folder structure. When these modules were installed into a repo-local folder of all modules, the duplication would cause the second task to overwrite the first. As of Gradle 8, this is a build-time error.

Specifying unique module names ensures these write to separate folders and thus are able to be built without stomping on each other.

Closes #988

### DIFF
--- a/samples/counter/browser/build.gradle
+++ b/samples/counter/browser/build.gradle
@@ -1,10 +1,13 @@
 apply plugin: 'org.jetbrains.kotlin.js'
 apply plugin: 'app.cash.redwood'
 
+// The name of the final JS file to be referenced from HTML.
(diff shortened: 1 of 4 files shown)


chrisbanes (approved):

chrisbanes:
Woah! Great detective work on this.

PR #1065, opened 2023-05-12
Parse FIR into annotation mirror types

JakeWharton:
This separates the FIR parsing from the interpretation of the semantics.

Refs #19

### DIFF
--- a/redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/schemaParserFir.kt
+++ b/redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/schemaParserFir.kt
@@ -15,7 +15,12 @@
  */
 package app.cash.redwood.tooling.schema
 
+import app.cash.redwood.tooling.schema.Deprecation.Level
+import app.cash.redwood.tooling.schema.ProtocolWidget.ProtocolChildren
+import app.cash.redwood.tooling.schema.ProtocolWidget.ProtocolProperty
+import app.cash.redwood.tooling.schema.SchemaAnnotation.DependencyAnnotation
 import java.io.File
+import java.net.URLClassLoader
 import org.jetbrains.kotlin.KtVirtualFileSourceFile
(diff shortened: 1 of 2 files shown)


JakeWharton:
If I was smarter, I would have migrated the existing parser away from reflection into an abstraction similar to the data classes added in this PR. Then the migration would be "just" adding a FIR-to-data-class step with the semantic parsing still remaining the same.

I guess there's still time... maybe tomorrow.

redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/schemaParserFir.kt
```diff
@@ -160,10 +166,44 @@ public fun parseProtocolSchema(
...
+  val dependencyClassLoader = URLClassLoader(dependencies.map { it.toURI().toURL() }.toTypedArray())
+  val dependencySchemas = schema.taggedDependencies.entries
+    .associate { (dependencyTag, dependencyType) ->
+      require(dependencyTag != 0) {
+        "Dependency $dependencyType tag must not be non-zero"
+      }
+
+      val dependency = loadProtocolSchema(
+        type = dependencyType,
+        classLoader = dependencyClassLoader,
```
JakeWharton, line 178: All these big chunks are copy/pasta from reflection parser

redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/schemaParserFir.kt
```diff
@@ -374,19 +450,224 @@ private fun FirContext.parseLayoutModifier(
...
   return ParsedProtocolLayoutModifier(
     tag = tag,
-    scopes = scopes,
+    scopes = annotation.scopes,
     type = memberType,
-    deprecation = TODO(),
+    deprecation = deprecation,
     properties = properties,
   )
 }
```
JakeWharton, line 464: Everything below this comment is the new part, although a lot of it was moved from above.

chrisbanes (approved):

PR #1073, opened 2023-05-15
Remove `redwood-treehouse-lazylayout-paging`

### LINKED ISSUE #1046: Remove (or deprecate) `redwood-treehouse-lazylayout-paging`
This artifact adds support for the functions`items` and `itemsIndexed`, both of which are deprecated starting with `paging-compose:1.0.0-alpha19`. They have been replaced with direct calls to the underlying `items` Composable, and helper functions to create the `key` and `contentType` (see [the docs](https://developer.android.com/jetpack/androidx/releases/paging#1.0.0-alpha19) for more info).

Both of these helper functions (`itemKey` and `itemContentType`) exist in `app.cash.paging:paging-compose-common:3.2.0-alpha05-0.2.1`, so the functions within `redwood-treehouse-lazylayout-paging` can be deprecated or removed.

I don't have strong feelings between deprecating or removing them. I'm assuming that no one is currently using this artifact as Lazy Layout is currently under development, so it should be safe to just delete the module all together. I'm happy to deprecate it though instead.

veyndan:
Fixes #1046


JakeWharton (approved):
Need to update settings.gradle

Issue #1043, opened 2023-05-10
Eliminate `data class` from public API


JakeWharton:
They're a nightmare to evolve in a source- and binary-compatible way.

Detekt has a rule for this!

JakeWharton:
As part of this, switch the protocol to using `Array<out Xyz>` rather than `List<Xyz>`, although this means we can't use Poko because of https://github.com/drewhamilton/Poko/issues/1.

PR #1076, opened 2023-05-15
Support multiple arguments in event lambdas

### LINKED ISSUE #1069: Allow more than one event argument
Change the encoding on the wire to be a JSON array of values. The default changes to empty array.

JakeWharton:
Closes #1069

### DIFF
--- a/redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
+++ b/redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
@@ -37,7 +37,7 @@ public data class Event(
   val id: Id,
   /** Identifies which event occurred on the widget with [id]. */
   val tag: EventTag,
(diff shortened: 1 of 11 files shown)


chrisbanes (approved):
🚢

redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
```diff
@@ -37,7 +37,7 @@ public data class Event(
   val id: Id,
   /** Identifies which event occurred on the widget with [id]. */
   val tag: EventTag,
-  val value: JsonElement = JsonNull,
+  public val args: List<JsonElement> = emptyList(),
```
JakeWharton: I was flirting with making this not a `data class`, but I'll save that for #1043 
```suggestion
  val args: List<JsonElement> = emptyList(),
```

veyndan (approved):
Sweeeet

PR #1081, opened 2023-05-15
Parse event lambdas and children scopes in FIR

JakeWharton:
Refs #19. All tests now pass. Need to test on our larger schema and see what breaks before switching over.


redwood-tooling-schema/src/test/kotlin/app/cash/redwood/tooling/schema/SchemaParserTest.kt
```diff
@@ -406,14 +409,14 @@ class SchemaParserTest(
...
-      .hasMessage(
-        "@Children app.cash.redwood.tooling.schema.SchemaParserTest.ScopedChildrenInvalidWidget#children lambda receiver can only be a class. " +
-          "Found: kotlin.collections.List<kotlin.Int>",
-      )
+      .message()
+      .isNotNull()
+      .all {
+        contains("@Children app.cash.redwood.tooling.schema.SchemaParserTest.ScopedChildrenInvalidWidget#children lambda receiver can only be a class.")
+        containsMatch(Regex("""Found: (kotlin\.collections\.)?List<(kotlin\.)?Int>"""))
+      }
```
JakeWharton, line 419: Can't easily get these to render the same. Will switch back to testing only one format once we drop reflection.

chrisbanes (approved):

PR #1082, opened 2023-05-16
Add ‘pull to refresh’ and indicator support to LazyLists

chrisbanes:
This maps to:

- Android views: SwipeRefreshLayout
- Compose UI: Material's Modifier.pullRefresh()
- UIKit: UIRefreshControl

https://github.com/cashapp/redwood/assets/227486/b02e83b4-bc52-48db-8434-2816aa1e9302

https://github.com/cashapp/redwood/assets/227486/616e81bc-dfaf-4054-a2e0-d29c8bd5c162

https://github.com/cashapp/redwood/assets/227486/9dafb10b-6939-4d0d-9c3d-b17bd75d936e

### DIFF
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyDsl.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyDsl.kt
@@ -16,6 +16,7 @@
 package app.cash.redwood.lazylayout.compose
 
 import androidx.compose.runtime.Composable
+import app.cash.redwood.LayoutModifier
 import app.cash.redwood.LayoutScopeMarker
 
 @LayoutScopeMarker
@@ -64,22 +65,63 @@ public inline fun <T> LazyListScope.itemsIndexed(
   itemContent(it, items[it])
 }
 
+@RequiresOptIn("This API is for use by Redwood's generated code only")
+public annotation class ExperimentalRedwoodLazyLayoutApi
+
 @Composable
 public fun LazyRow(
+  layoutModifier: LayoutModifier = LayoutModifier,
   content: LazyListScope.() -> Unit,
 ) {
   LazyList(
     isVertical = false,
+    layoutModifier = layoutModifier,
+    content = content,
+  )
+}
+
+@ExperimentalRedwoodLazyLayoutApi
+@Composable
+public fun LazyRow(
+  refreshing: Boolean,
+  onRefresh: (() -> Unit)?,
+  layoutModifier: LayoutModifier = LayoutModifier,
+  content: LazyListScope.() -> Unit,
+) {
+  RefreshableLazyList(
+    isVertical = false,
+    refreshing = refreshing,
+    onRefresh = onRefresh,
+    layoutModifier = layoutModifier,
     content = content,
   )
 }
 
 @Composable
 public fun LazyColumn(
+  layoutModifier: LayoutModifier = LayoutModifier,
   content: LazyListScope.() -> Unit,
 ) {
   LazyList(
     isVertical = true,
+    layoutModifier = layoutModifier,
+    content = content,
+  )
+}
+
+@ExperimentalRedwoodLazyLayoutApi
+@Composable
+public fun LazyColumn(
+  refreshing: Boolean,
+  onRefresh: (() -> Unit)?,
+  layoutModifier: LayoutModifier = LayoutModifier,
+  content: LazyListScope.() -> Unit,
+) {
+  RefreshableLazyList(
+    isVertical = true,
+    refreshing = refreshing,
+    onRefresh = onRefresh,
+    layoutModifier = layoutModifier,
     content = content,
   )
 }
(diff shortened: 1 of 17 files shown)


chrisbanes:
The UIKit version acts a little strangely, in that it refreshes instantly after pulling down a pixel or so. From what I can tell from searching, the refresh control is supposed to refresh after the user has scrolled 50% of the host view, but that obviously isn't happening. Might be due to the way we're setting up the UITableView. This is probably fine to land as is, and I'll follow up with a fix once we work out the issue.

veyndan (commented):
I'm a bit wary about having the refresh indicator baked into Lazy List, as it mixes (imo) logically two separate things. I'd also assume that one could refresh other things beyond Lazy Lists, e.g., a `Column`, or some arbitrary user-defined Widget (`MyFancyCustomLazyList`).

From what I can tell from an internal Slack thread, you changed up the API design to better match UIKit, as iOS doesn't separate the idea of the refresh indicator and lazy list like Android does. Is that correct, or am I missing something?

chrisbanes:
> From what I can tell from an internal Slack thread, you changed up the API design to better match UIKit, as iOS doesn't separate the idea of the refresh indicator and lazy list like Android does. Is that correct, or am I missing something?

Correct. We're mostly stuck between a rock and a hard place here. We can't create a separate wrapper widget (`PullToRefresh` or whatever) as that breaks all of the layering for UIKit _and_ Compose UI, where you need to decorate the child widget.

This solution is mostly a practical choice for right now. One thing we could do is add a overloads called `RefreshableLazyColumn`, etc, which are marked experimental. That should allow us to change things up in the future.

chrisbanes:
Split out `RefreshableLazyList`, with experimental marked `Lazy*` composable overloads.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyDsl.kt
```diff
@@ -64,22 +65,63 @@ public inline fun <T> LazyListScope.itemsIndexed(
   itemContent(it, items[it])
 }
 
+@RequiresOptIn("This API is for use by Redwood's generated code only")
```
JakeWharton: Bad copy/paste?

JakeWharton (commented):
Missed this today. Too hard to review on mobile. Will grab in the AM.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyDsl.kt
chrisbanes: Woops, fixed.

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -55,35 +60,51 @@ internal class Items<VH : RecyclerView.ViewHolder>(
...
 }
 
-internal class ViewLazyList(
-  override val value: RecyclerView,
+/**
+ * Public function to allow downstream factories to create their own LazyList
+ */
+public fun ViewLazyList(
+  recyclerViewFactory: () -> RecyclerView,
+): LazyList<View> = ViewLazyListImpl(recyclerViewFactory)
```
veyndan, line 68: I'm a bit confused by this bit. Surely if someone didn't want to use our implementation of `ViewLazyList`, then they could just not depend on the module.

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -55,35 +60,51 @@ internal class Items<VH : RecyclerView.ViewHolder>(
...
-  override val items = Items(adapter)
+  override val value: View get() = recyclerView
+
+  final override val items = Items(adapter)
 
   init {
     adapter.items = items
-    value.apply {
+    recyclerView.apply {
+      setHasFixedSize(true)
```
veyndan, line 97: I don't think we can guarantee this

veyndan (approved):
Looks good. I still feel a bit iffy about having the lazy list and refresh controls be inside the same widget. I'm curious if Redwood can support this use case somehow, because going from one Redwood nodes to multiple views is trivial, but is the reverse possible (i.e., can a Redwood subtree map to a single view)?

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
chrisbanes: It's for downstream apps. It allows apps (say Cash) to provide their own factory for `LazyList<View>` and create the 'default' impl and wrap it as necessary.  We could remove this and just rely on the `Factory` interface itself, but this allows apps to inject their won `RecyclerView` or whatever.

chrisbanes:
> I'm curious if Redwood can support this use case somehow, because going from one Redwood nodes to multiple views is trivial, but is the reverse possible?

I raised #1084 to see if we can do something better. We'll need something similar for animations anyway.

@JakeWharton feel free to review and I'll fix anything in a follow-up.

PR #1083, opened 2023-05-16
Use IntervalList data structure to efficiently handle LazyList intervals

veyndan:
This is how Compose UI internally handles intervals within `LazyList`. I opted for a standard `List` before the recent refactoring as the data structure would've had to be serialized and deserialized, which didn't seem worth it. Now that the concept of `intervals` purely exists on the guest side, we don't have to handle serialization of the data structure.

Couple of benefits from the additional data structure besides being what Compose UI internally uses:

1. The total size of the list of intervals is efficiently calculated.
2. Retrieval of an interval is done via a binary search instead of search sequentially.

### DIFF
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/layout/IntervalList.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/layout/IntervalList.kt
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/layout/IntervalList.kt
```diff
@@ -0,0 +1,213 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package app.cash.redwood.lazylayout.compose.layout
+
+import androidx.compose.runtime.collection.MutableVector
+import androidx.compose.runtime.collection.mutableVectorOf
+
+// Copied from https://github.com/androidx/androidx/blob/67066da9297ce407186a08725dd43bc66694e115/compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/lazy/layout/IntervalList.kt
```
veyndan, line 21: The only changes to this file are making the visibility modifiers `internal` instead of `public`, and removing the `@ExperimentalFoundationApi` annotation, as it is unavailable outside of Compose Foundation.

JakeWharton (approved):

PR #1085, opened 2023-05-16
Introduce LazyLayoutIntervalContent for generic interval-based content

veyndan:
Further decreasing the delta between our implementation of `LazyList` and Compose UI's.


JakeWharton (approved):
I'm not sure what this buys us. Seems to be all implementation detail. Is it just alignment with the upstream?

veyndan:
Yep, it's alignment with upstream. Practically, this change would aid the addition of a `LazyGrid` layout, but that's not yet on the horizon. The main motive was aligning with upstream.

PR #1094, opened 2023-05-19
Limit window size of LazyList items instead of growing as one scrolls

veyndan:
The rough idea is here. Obviously, placeholders need to be hoisted to the guest code, but will do that in a follow-up.


redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
```diff
@@ -22,6 +22,8 @@ import app.cash.redwood.schema.Widget
 @Widget(1)
 public data class LazyList(
   @Property(1) val isVertical: Boolean,
-  @Property(2) val onPositionDisplayed: (position: Int) -> Unit,
+  @Property(2) val onViewportChanged: (range: IntArray) -> Unit,
```
JakeWharton: Multiple event arguments are supported now if you want to explicitly declare the two here instead of this more general structure.

JakeWharton (approved):
Nice

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
```diff
@@ -53,16 +56,22 @@ internal fun LazyList(
...
     isVertical,
-    onPositionDisplayed = { position ->
-      /** Triggers load at position, loading all items within [PagingConfig.prefetchDistance] of the [position]. */
-      if (position < lazyPagingItems.itemCount) {
-        lazyPagingItems[position]
-      }
+    itemsBefore = window.value.first,
+    itemsAfter = lazyPagingItems.itemCount - (window.value.last + 1),
+    onViewportChanged = { (start, end) ->
+      val newWindow = maxOf(start - pagingConfig.prefetchDistance, 0) until (end + pagingConfig.prefetchDistance)
```
swankjesse: Nice!

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
swankjesse: Yay!

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -84,22 +94,62 @@ internal class ViewLazyList(
...
+    this.onViewportChanged = onViewportChanged
   }
 
-  private class LazyContentItemListAdapter : RecyclerView.Adapter<ViewHolder>() {
-    var onPositionDisplayed: ((Int) -> Unit)? = null
+  private fun updateViewport() {
+    val newViewport = linearLayoutManager.findFirstVisibleItemPosition()..linearLayoutManager.findLastVisibleItemPosition()
+    if (newViewport != viewport) {
+      this@ViewLazyList.viewport = newViewport
+      onViewportChanged?.invoke(intArrayOf(newViewport.first, newViewport.last + 1))
```
swankjesse: Do we do too many `+ 1` calls on the range end? I think if we have the `+ 1` here we can probably remove it on the assignment of `itemsAfter` in `LazyList.kt`.

swankjesse: Checked the math, and I’m wrong. As-is is correct! (I really wish they hadn’t deprecated `endInclusive`. What a fail.)

swankjesse (approved):

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
veyndan: I added some logs and `linearLayoutManager.findLastVisibleItemPosition() == newViewport.last`, so I'm gonna remove the `+ 1` and propagate that to the callback.

PR #1096, opened 2023-05-19
Use javac release flag on all compilations

JakeWharton:
This fixes a mismatch of target/release usage in the Gradle plugin:

    Execution failed for task ':redwood-gradle-plugin:compileKotlin'.
    > 'compileJava' task (current target is 1.8) and 'compileKotlin' task (current target is 11) jvm target compatibility should be set to the same Java version.

No idea why this wasn't caught on the PR. We absolutely run the Gradle plugin tests on CI.

cc @ZacSweers

### DIFF
--- a/build-support/build.gradle
+++ b/build-support/build.gradle
@@ -55,8 +55,7 @@ buildConfig {
 }
 
 tasks.withType(JavaCompile).configureEach {
-  sourceCompatibility = JavaVersion.VERSION_11.toString()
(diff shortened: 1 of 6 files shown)


chrisbanes (approved):

JakeWharton:
Ugh. I'm just going to revert the original commit for now. I have no idea how it got through CI but causes other PRs to fail. Gradle \*shakes fist\*

PR #1097, opened 2023-05-19
Revert "Use release flag in JavaCompile"

JakeWharton:
Reverts cashapp/redwood#1092


chrisbanes (approved):

JakeWharton:
Probably can bring this back once configuration moves into the build plugin and we have more knowledge about what type of module it is being applied to.

PR #1099, opened 2023-05-19
Add documentation to the schema and codegen

JakeWharton:
Start parsing it with FIR, but currently we retain the syntax which forms the comment and need to parse it away in the future.

Refs #19


colinrtwhite (approved):
Awesome!

PR #1103, opened 2023-05-19
Make value class constructors and properties fully private

JakeWharton:
The bug was fixed in the 1.4 Compose compiler.


colinrtwhite (approved):
Nice!

PR #1111, opened 2023-05-22
Rename LayoutModifier to Modifier

### LINKED ISSUE #1047: Rename `LayoutModifier` to `Modifier`
We don't want to go full Compose UI `Modifier`, but we also think that want to use this as a mechanism to globally attach some properties to all widgets (such as stable IDs, accessibility information, etc.). Additionally, it would let users attach metadata globally.

We talked about this back in February and agreed to do it. This is a tracking issue for following through on it.

chrisbanes:
- This was performed mostly through 'Replace in Files'. Everything has been renamed.
- `Widget.modifiers` has been renamed to `modifier` to follow Compose UI
- For Compose UI widgets/tests, we need to disambiguate the Modifier class.

Fixes #1047


build-support/src/main/resources/app/cash/redwood/buildsupport/flexboxHelpers.kt
```diff
@@ -75,14 +75,14 @@ internal fun Margin.toSpacing(density: Density) = with(density) {
 internal fun newFlexItem(
   direction: FlexDirection,
   density: Density,
-  layoutModifiers: LayoutModifier,
+  modifiers: Modifier,
```
JakeWharton: Does Compose UI use "modifiers" or "modifier" for parameter/property name? Next up we should normalize that because I feel like we mix and match both right now.

redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodGeneratorPlugin.kt
```diff
@@ -54,7 +54,7 @@ public abstract class RedwoodGeneratorPlugin(
   ) {
     Compose("--compose", "redwood-compose"),
     ComposeProtocol("--compose-protocol", "redwood-protocol-compose"),
-    LayoutModifiers("--layout-modifiers", "redwood-runtime"),
+    Modifiers("--layout-modifiers", "redwood-runtime"),
```
JakeWharton: Find/replace this string too. Can be follow-up. Will also be referenced from the Gradle plugin's `GenerateTask`

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/GenerateCommand.kt
```diff
@@ -35,7 +35,7 @@ internal class GenerateCommand : CliktCommand(name = "generate") {
     .switch(
       "--compose" to CodegenType.Compose,
       "--compose-protocol" to ProtocolCodegenType.Compose,
-      "--layout-modifiers" to CodegenType.LayoutModifiers,
+      "--layout-modifiers" to CodegenType.Modifiers,
```
JakeWharton: Another copy of the string form

redwood-layout-schema/src/main/kotlin/app/cash/redwood/layout/modifiers.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Square, Inc.
+ * Copyright (C) 2023 Square, Inc.
```
JakeWharton:
```suggestion
 * Copyright (C) 2022 Square, Inc.
```

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -124,9 +124,9 @@ public open class EventListener {
   /**
    * Invoked on a request to create an unknown layout modifier [tag].
```
JakeWharton:
```suggestion
   * Invoked on a request to create an unknown modifier [tag].
```
Probably worth searching for "layout modifier" string, too.

settings.gradle
```diff
@@ -61,6 +60,7 @@ include ':redwood-lazylayout-testing'
 include ':redwood-lazylayout-uiview'
 include ':redwood-lazylayout-view'
 include ':redwood-lazylayout-widget'
+include ':redwood-modifiers'
```
JakeWharton: I think this should have ended up as 'redwood-layout-modifiers' since it's modifiers that are specific to the 'layout' modules.

JakeWharton (approved):

build-support/src/main/resources/app/cash/redwood/buildsupport/flexboxHelpers.kt
chrisbanes: It uses `modifier: Modifier = Modifier`. Updated

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/GenerateCommand.kt
chrisbanes: ✅

settings.gradle
chrisbanes: ✅

redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodGeneratorPlugin.kt
chrisbanes: ✅

PR #1118, opened 2023-05-22
Use a schema's `ProtocolBridge` to record changes from a `WidgetValue`

### LINKED ISSUE #1033: `ViewTree` API incorrect children tag logic
https://github.com/cashapp/redwood/blob/trunk/redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt#L371-L373

Tags are not guaranteed to be contiguous and they are not offsets of tag into which their own node is inserted

JakeWharton:
This adds a function which given a `WidgetValue` can re-hydrate a full `Widget` given an appropriate `Widget.Provider`. When the protocol-based `Widget.Provider` is used, the result is the same set of changes that normal protocol-based composition produces.

Closes #1033

### DIFF
--- a/redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
+++ b/redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
@@ -15,11 +15,6 @@
  */
 package app.cash.redwood.tooling.codegen
 
-import app.cash.redwood.tooling.codegen.Protocol.ChildrenTag
-import app.cash.redwood.tooling.codegen.Protocol.Id
-import app.cash.redwood.tooling.codegen.Protocol.ViewTreeBuilder
-import app.cash.redwood.tooling.schema.ProtocolWidget
-import app.cash.redwood.tooling.schema.ProtocolWidget.ProtocolProperty
 import app.cash.redwood.tooling.schema.ProtocolWidget.ProtocolTrait
 import app.cash.redwood.tooling.schema.Schema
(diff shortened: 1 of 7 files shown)


JakeWharton:
Another way to think about this is that we no longer duplicate the code which serializes properties to the protocol. Instead, by adding a way for a `WidgetValue` to write its values back into a real `Widget`, we can reuse the existing serialization code of the widget protocol.

chrisbanes (approved):

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -315,23 +310,18 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
...
       }
       is Children -> {
         type = Stdlib.List.parameterizedBy(RedwoodTesting.WidgetValue)
         defaultExpression = CodeBlock.of("%M()", Stdlib.listOf)
         addEqualsHashCodeToString(trait)
 
         childrenLists += CodeBlock.of("%N", trait.name)
+
+        toWidgetChildrenBuilder.beginControlFlow("for ((index, child) in %N.withIndex())", trait.name)
+          .addStatement("instance.%N.insert(index, child.toWidget(provider))", trait.name)
```
JakeWharton, line 323: This is the change that required #1110. We now access the children property in a loop. And while I could pull it into a local or use `apply` or `let` or something, still seems like a good change because it's a needless foot-gun otherwise.

PR #1122, opened 2023-05-22
Rename `ViewTree` to `SnapshotChangeList`

### LINKED ISSUE #953: Tools to serialize, deserialize, and render a view tree
I’d like to be able a serializable snapshot of a view tree. 

We can represent this today as a `Diff`.

```kotlin
@Serializable
class ViewTree(val diff: Diff)
```

Next I’d like an easy API to collect these. `RedwoodTester` is a natural choice.

```kotlin
  fun collectEmojiSearchSamples(): Map<String, ViewTree> {
    val result = mutableMapOf<ViewTree>()

    EmojiSearchTester().test {
      setContent {
        BasicEmojiSearch()
      }

      val snapshot0 = awaitSnapshot()
      result["home"] = snapshot0.asViewTree()

      snapshot0.flatten().filterIsInstance<TextInputValue>().first()
        .onChange!!.invoke(TextFieldState(text = "tree"))

      val snapshot1 = awaitSnapshot()
      result["search"] = snapshot1.asViewTree()
    }

    return result
  }
```

Once I can collect 'em and encode ’em, I can write my own code to keep an inventory:

```
class EmojiSearchSnapshotsStore(
  val fileSystem: FileSystem,
  val json: Json,
) {
  fun storeSnapshots(path: Path, snapshots: Map<String, ViewTree>) { ... }
  fun loadSnapshots(path: Path): Map<String, ViewTree> { ... }
}
```

Next I can create snapshots in our iOS codebase:

```swift
let widgetFactories = EmojiSearchWidgetFactories(...)
let viewTreeUIKitView = ViewTreeUIKitView(widgetFactories)

func screenshotSample(id: String, viewTree: ViewTree) {
  viewTreeUIKitView.setModel(viewTree)
  assertSnapshot(
    matching: viewTreeUIKitView,
    as: .image(size: view.intrinsicContentSize, fileName: id))
}
```

I’ll write similar code in our Android codebase, and then I can check the images for deviations.

JakeWharton:
Closes #953

### DIFF
--- a/redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/toChangeList.kt
+++ b/redwood-compose-testing/src/commonMain/kotlin/app/cash/redwood/compose/testing/toChangeList.kt
@@ -15,24 +15,32 @@
  */
 package app.cash.redwood.compose.testing
 
-import app.cash.redwood.protocol.ViewTree
+import app.cash.redwood.protocol.SnapshotChangeList
 import app.cash.redwood.protocol.compose.ProtocolBridge
 import kotlinx.serialization.json.Json
 
-public fun List<WidgetValue>.toViewTree(
+/**
+ * Encode this snapshot of widget values into a list of changes which can be serialized and
+ * later applied to the UI to recreate the structure and state.
+ */
+public fun List<WidgetValue>.toChangeList(
   factory: ProtocolBridge.Factory,
   json: Json = Json.Default,
-): ViewTree {
+): SnapshotChangeList {
   val bridge = factory.create(json)
   for ((index, child) in withIndex()) {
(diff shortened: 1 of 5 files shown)


JakeWharton:
Open to other prefixes, but "change list" feels like an appropriate name for what these are.

~Was a tossup between "initial" and "snapshot" as the first choice. Went with "initial" since it better describes the contents (containing no move/remove), whereas "snapshot" more reflects what it models. Happy to change it to "snapshot" or something better with a gentle nudge.~

Hi

JakeWharton:
Actually I just convinced myself to flip-flop to "snapshot". It's far more clear when talking about the feature as a whole as well as the type.

colinrtwhite (approved):

PR #1136, opened 2023-05-24
Do not spend more than 5m trying to download caches on CI

JakeWharton:
Otherwise we're just wasting time. The default timeout is 360(!!!) minutes.


JakeWharton:
Hey it works! Check the 'sample-counter' job.

veyndan (approved):

PR #1138, opened 2023-05-24
Group ktlint and Spotless in Renovate

### LINKED ISSUE #1132: Bundle ktlint and Spotless in Renovate
Also need a dummy version catalog entry for ktlint similar to AndroidX Compose so that it gets updated.

JakeWharton:
Closes #1132

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -14,5 +14,13 @@
         "org.jetbrains.kotlin:kotlin",
       ],
     },
+    {
+      // ktlint breaks their API constantly and regularly needs a Spotless update to work.
+      "groupName": "Spotless/ktlint",
+      "matchPackagePrefixes": [
+        "com.diffplug.spotless",
+        "com.pinterest:ktlint",
+      ],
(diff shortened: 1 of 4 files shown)


renovate.json5
```diff
@@ -14,5 +14,13 @@
         "org.jetbrains.kotlin:kotlin",
       ],
     },
+    {
+      // ktlint breaks their API constantly and regularly needs a Spotless update to work.
```
colinrtwhite, line 18: Supposedly `0.49.0` should be the last set of breaking changes (see "API Changes & RuleSet providers & Reporter Providers" [here](https://github.com/pinterest/ktlint/releases/tag/0.49.0)) 🤞

colinrtwhite (approved):

renovate.json5
JakeWharton: Oh nice. If it goes well we can drop this. It doesn't force them together anyway, only merges when there are both.

PR #1154, opened 2023-05-31
Naively extract placeholders from host to guest

### LINKED ISSUE #771: Add placeholder support to `LazyColumn`

veyndan:
Closes #771.

This change allows callers to specify what sort of placeholder they want to display. The purpose of pooling is outlined in the opening paragraph of #771.

The pool of placeholders has been hardcoded to 75. It's important that this number always be at least the number of placeholders on screen, and I feel that 75 is sufficiently large in cases where items are very small. This mechanism should obviously be improved in follow up PRs.

### DIFF
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
@@ -32,6 +32,7 @@ private const val OffscreenItemsBufferCount = 30
 internal fun LazyList(
   isVertical: Boolean,
   modifier: Modifier = Modifier,
+  placeholder: @Composable () -> Unit,
   content: LazyListScope.() -> Unit,
 ) {
   val itemProvider = rememberLazyListItemProvider(content)
@@ -48,6 +49,7 @@ internal fun LazyList(
       lastVisibleItemIndex = localLastVisibleItemIndex
     },
     modifier = modifier,
+    placeholder = { repeat(75) { placeholder() } },
     items = {
       for (index in itemsBefore until itemProvider.itemCount - itemsAfter) {
         key(index) {
@@ -64,6 +66,7 @@ internal fun RefreshableLazyList(
   refreshing: Boolean = false,
   onRefresh: (() -> Unit)? = null,
   modifier: Modifier = Modifier,
+  placeholder: @Composable () -> Unit,
   content: LazyListScope.() -> Unit,
 ) {
   val itemProvider = rememberLazyListItemProvider(content)
@@ -82,6 +85,7 @@ internal fun RefreshableLazyList(
     refreshing = refreshing,
     onRefresh = onRefresh,
     modifier = modifier,
+    placeholder = { repeat(75) { placeholder() } },
     items = {
       for (index in itemsBefore until itemProvider.itemCount - itemsAfter) {
         key(index) {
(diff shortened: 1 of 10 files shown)


redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
```diff
@@ -82,6 +85,7 @@ internal fun RefreshableLazyList(
     refreshing = refreshing,
     onRefresh = onRefresh,
     modifier = modifier,
+    placeholder = { repeat(75) { placeholder() } },
```
swankjesse, line 88: Whoa, neat idea

swankjesse (approved):
This is a nice simplification. I think 75 is a lot! What if it was like 5 or 10 and there was a graceful recovery if the pool was empty?

veyndan:
> This is a nice simplification. I think 75 is a lot! What if it was like 5 or 10 and there was a graceful recovery if the pool was empty?

75 is a lot — and 5 or 10 is much too small! As mentioned in the PR description, "It's important that this number [75] always be at least the number of placeholders on screen" — which wouldn't be the case ~80% of the time if the number of placeholders was 5 or 10.

In terms of graceful recovery, if we exhaust the placeholder pool, we'd have to request more placeholders, and show a placeholder for the placeholder while we're requesting more placeholders. The fact that that sentence was a bit confusing is the point. I don't particularly want to double down on the approach here. The motivation of the PR was purely to allow user-defined placeholders in the least radical way possible.

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -163,33 +184,49 @@ internal open class ViewLazyListImpl(
...
-      },
-    )
+    override fun getItemViewType(position: Int): Int {
+      val index = position - items.itemsBefore
+      return if (index in items.widgets.indices) VIEW_TYPE_ITEM else VIEW_TYPE_PLACEHOLDER
+    }
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
+      return when (viewType) {
+        VIEW_TYPE_PLACEHOLDER -> ViewHolder.Placeholder(placeholders.take().value)
```
JakeWharton, line 201: I think this crashes if we exceed 75 items? Extremely rare, of course, but not technically impossible.

JakeWharton (commented):
Cool solution.

I agree that 75 is a lot, but weirdly it could also sometimes not be enough.

I think we probably want to pick a lower value which targets the 99% case rather than what we have now which seems more like the 99.999% case. By virtue of using `LinearLayoutManager` get can query the number of visible items on screen on the widget-side. That should allow dynamically scaling the placeholder count to be something like 1.5x the number of currently-displayed items. This should handle a full page jump and will handle a full fling. The additional .5x means we should hopefully always stay ahead of item size changes.

Then, in the extraordinarily rare case of not having a placeholder, it's probably easiest to display an empty layout at the height of the item whose view holder is being recycled.

Maybe that's not the exact solution. But I'm definitely wary of a hard-coded magic value, especially one so large.

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
veyndan: Yep, it does. I chose the number 75 based on how many TextView's without any spacing could fit on a vertical Pixel 6. Completely arbitrary though!

PR #1155, opened 2023-05-31
Expose changing of `LazyList` size

### LINKED ISSUE #1114: sizing should be controlled by Modifiers
https://github.com/cashapp/redwood/blob/b22693362e45997ae67afb98e9450097f1975f57/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt#L107

```text

      setHasFixedSize(true)
      layoutManager = linearLayoutManager

      // TODO: sizing should be controlled by Modifiers
      layoutParams = ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT)

      addOnScrollListener(

```

veyndan:
Closes #1114

### DIFF
--- a/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
+++ b/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
@@ -24,10 +24,12 @@ import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
 import android.widget.FrameLayout
 import androidx.core.view.doOnDetach
+import androidx.core.view.updateLayoutParams
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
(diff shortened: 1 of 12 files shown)


redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -123,6 +122,20 @@ internal open class ViewLazyListImpl(
     recyclerView.adapter = adapter
   }
 
+  override fun width(width: Constraint) {
+    recyclerView.updateLayoutParams {
+      this.width = if (width == Constraint.Fill) MATCH_PARENT else WRAP_CONTENT
+    }
+    recyclerView.requestLayout()
```
JakeWharton: Setting layout parameters will automatically request layout. Although it doesn't really hurt to have it here...

JakeWharton (approved):

swankjesse (approved):

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
veyndan: Oh, true! I'll remove it cause I don't want to imply that the code is doing more than it is.

PR #1160, opened 2023-06-01
Upgrade to Zipline 0.9.20.

swankjesse:
This update is neither forwards- nor backwards-compatible with the previous version of Zipline.


JakeWharton (approved):

gradle/libs.versions.toml
```diff
@@ -6,7 +6,7 @@ androidxComposeCompiler = "1.4.5"
 androidx-activity = "1.7.2"
 jbCompose = "1.4.0"
 lint = "31.0.2"
-zipline = "0.9.18"
+zipline = "0.9.20"
```
swankjesse, line 9: (I’m jumping the gun here. This isn’t on Maven Central yet!)

PR #1162, opened 2023-06-01
Remove explicit `public` modifier usage in codegen

JakeWharton:
Harmless for regular types and functions, but in the case of being paired with an `override` modifier this will actually cause a redundant emission in the codegen where one is not required to be explicit.


JakeWharton:
Also updated to use new `FunSpec.builder(MemberName)` helper, where possible.

colinrtwhite (approved):
Does this prevent users from applying explicit API mode to modules that use this plugin?

JakeWharton:
Nope! The library always emits with explicit API compatibility and our test schema has it turned on (as well as warnings as errors) to validate this.

JakeWharton:
The only use of public should be with override when increasing the visibility of a member. Otherwise it's the default everywhere.

PR #1164, opened 2023-06-02
Rollback serialization to 1.5.0

swankjesse:
The 1.5.1 update requires Kotlin 1.8.21.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 kotlin = "1.8.20"
 kotlinx-coroutines = "1.7.1"
-kotlinx-serialization = "1.5.1"
+kotlinx-serialization = "1.5.0"
 androidxComposeCompiler = "1.4.5"
 androidx-activity = "1.7.2"
 jbCompose = "1.4.0"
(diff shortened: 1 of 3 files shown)


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 kotlin = "1.8.20"
 kotlinx-coroutines = "1.7.1"
-kotlinx-serialization = "1.5.1"
+kotlinx-serialization = "1.5.0"
```
swankjesse, line 4: This was causing problems when building a downstream project.

JakeWharton: Hmm this may use the newer 1.8.21 stdlib but shouldn't be bumping the gradle plugin or anything that would cause problems with using 1.8.20. Otherwise we would see the same problems here.

JakeWharton (approved):
I'll approve, but seems like something else is going wrong if this is actually causing problems downstream.

swankjesse:
The actual problem is when we build our downstream repo, it’s upset that there’s both 1.5.0 and 1.5.1 artifacts in the transitive dependencies. For JVM that’s fine but for our Kotlin/Native build that triggers an error.

PR #1168, opened 2023-06-05
Introduce `ChangeListener` for widget-local end-of-change notification

### LINKED ISSUE #1133: Mechanism for widgets to process a batch of changes
In LazyColumn when the scrolling window moves down by _N_ we perform many changes as a unit:

 * Increment `itemsBefore` by _N_
 * Remove _N_ elements from the start of `items`
 * Add _N_ elements to the end of `items`
 * Decrement itemsAfter by _N_

If we perform each of these changes independently, we have two bad options:

 * Janky scrolling behavior, because we send unnecessary ‘insert’ and ’remove’ events (#1127)
 * Inefficient native binding, because we signal that the entire dataset has changed (#1128)

A fix would be for Redwood to signal ‘the current batch of changes is complete’ to widgets like `LazyList`. It could in turn defer notifying its adapter until it can do so precisely.

This strategy could be also enable optimizations in other widget types. For example, an `Image` widget that has both a baseUrl and size property should build & fetch its Thumbor URL only once when both properties are change.

JakeWharton:
Closes #1133

### DIFF
--- a/redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ChangeListenerTest.kt
+++ b/redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ChangeListenerTest.kt
@@ -1,0 +1,218 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.compose
+
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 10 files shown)


redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ChangeListenerTest.kt
```diff
@@ -0,0 +1,218 @@
...
+    c.setContent {
+      Button(text, onClick = null)
+    }
+    assertThat(c.awaitSnapshot()).containsExactly("modifier Modifier", "text hi", "onClick false", "onEndChanges")
+
+    text = "hello"
+    assertThat(c.awaitSnapshot()).containsExactly("text hello", "onEndChanges")
+  }
+
+  @Test fun unrelatedPropertyChangeDoesNotNotifyWidget() = runTest {
```
swankjesse, line 98: Nice

redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ListeningButton.kt
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+package app.cash.redwood.compose
+
+import app.cash.redwood.Modifier
+import app.cash.redwood.testing.WidgetValue
+import app.cash.redwood.widget.ChangeListener
+import example.redwood.widget.Button
+
+class ListeningButton : Button<WidgetValue>, ChangeListener {
```
swankjesse, line 23: I love how these tests work

swankjesse (approved):

PR #1170, opened 2023-06-05
Avoid generating pointless error strings in composables

### LINKED ISSUE #1166: `RedwoodComposeNode` produces unsavory bytecode junk
```
       250: invokeinterface #68,  1           // InterfaceMethod androidx/compose/runtime/Composer.getApplier:()Landroidx/compose/runtime/Applier;
       255: dup
       256: ldc           #70                 // String null cannot be cast to non-null type app.cash.redwood.compose.RedwoodApplier<P of app.cash.redwood.compose.RedwoodCompositionKt.RedwoodComposeNode>
       258: invokestatic  #76                 // Method kotlin/jvm/internal/Intrinsics.checkNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
```

This GIGANTIC string is emitted for a case which will never happen. The type is already non-null, but since the source property is in a separate library AND in order to maintain the guarantees of the Kotlin type system the null check has to be performed. It's just a shame this big-ass string gets shoved into the output with it.

We can probably do an explicit `!!` and suppress the unnecessary warning.

The offending code is here:

https://github.com/cashapp/redwood/blob/546aa52bd2d07126d4febd57d084e569f02dde07/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt#L139

JakeWharton:
This value can never be null, but the compiler insists on checking. Instead, do a bang-bang which avoids the string. Normally this wouldn't be a big deal, but this function is inlined into every generated composable.

Closes #1166

### DIFF
--- a/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
+++ b/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
@@ -135,8 +135,14 @@ public inline fun <P : Widget.Provider<*>, W : Widget<*>> RedwoodComposeNode(
   currentComposer.startNode()
 
   if (currentComposer.inserting) {
-    @Suppress("UNCHECKED_CAST") // Safe so long as you use generated composition function.
-    val applier = currentComposer.applier as RedwoodApplier<P>
+    // Perform an explicit !! on the return value to avoid the Kotlin compiler inserting a huge
+    // string into the output as an error message for an otherwise implicit null check.
+    @Suppress(
+      "UNCHECKED_CAST", // Safe so long as you use generated composition function.
+      "UNNECESSARY_NOT_NULL_ASSERTION",
+    )
+    val applier = currentComposer.applier!! as RedwoodApplier<P>
+
     currentComposer.createNode {
       @Suppress("UNCHECKED_CAST") // Safe so long as you use generated composition function.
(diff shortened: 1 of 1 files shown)


colinrtwhite (approved):

veyndan (approved):

JakeWharton:
Meant to add this to the commit message:

Before:
```
   192: aload_2
   193: invokeinterface #64,  1           // InterfaceMethod androidx/compose/runtime/Composer.getApplier:()Landroidx/compose/runtime/Applier;
   198: dup
   199: ldc           #66                 // String null cannot be cast to non-null type app.cash.redwood.compose.RedwoodApplier<kotlin.Any>
   201: invokestatic  #72                 // Method kotlin/jvm/internal/Intrinsics.checkNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
   204: checkcast     #74                 // class app/cash/redwood/compose/RedwoodApplier
   207: astore        8
```
After:
```
   192: aload_2
   193: invokeinterface #64,  1           // InterfaceMethod androidx/compose/runtime/Composer.getApplier:()Landroidx/compose/runtime/Applier;
   198: dup
   199: invokestatic  #70                 // Method kotlin/jvm/internal/Intrinsics.checkNotNull:(Ljava/lang/Object;)V
   202: checkcast     #72                 // class app/cash/redwood/compose/RedwoodApplier
   205: astore        8
```

Issue #1152, opened 2023-05-30
iOS UIKit needs to propagate size changes up the DOM


colinrtwhite:
On Android changing a property that changes the size of a view will re-layout all its parents in the view tree automatically. On iOS UIKit this is not the case! We need to manually notify that parent (and parent's parent, etc.) that the child view has changed size.

Currently we hack around this in a few places by doing something like:

```swift
func text(text: String) {
    root.text = text
    root.superview?.setNeedsLayout()
}
```

However, this isn't foolproof since it only re-lays out the parent and not the parent's parent, which also might have changed size due to its child's child changing its text.

How do we solve this? I think we have two main options:

## Automatic

We already handle changes similar to this through the `Widget.Children.onModifierUpdated` method, however this is only called when a child's modifier is updated. I propose we generalize it to:

```kotlin
public fun change(index: Int, property: String)
```

This would be called for all property changes and would pass the property name so Android (and similar platforms) can skip any change where `property != "modifier"`.

## Manual

Optionally, we could have the child widget notify its parent that its size should be recomputed. This has the benefit of the child not having to recompute its size for every property change. For example, the parent doesn't have to recompute the child's size when its text colour changes. The downside to this solution is it's manual and could be error-prone.

JakeWharton:
So as far as I can tell we should not have to do this manually. I've been reading and watching WWDC videos on how iOS layout works since you posted this issue, and all signs point to this being able to be made to work using the built-in mechanisms. I also chatted with @efirestone about it who also believes that it should be something else causing the incorrect behavior.

From your example,

> ```swift
> func text(text: String) {
>     root.text = text
>     root.superview?.setNeedsLayout()
> }
> ```

I would say two things:

1. Whatever type is `root` seems broken. Writes to its `text` property should automatically be invoking `setNeedsLayout()` on its parent, no? Controls like `UILabel` do this.
2. If for some reason there's a really good justification for point 1 to NOT be implemented that way, having the widget wrapper perform the action is an okay solution.

> However, this isn't foolproof since it only re-lays out the parent and not the parent's parent, which also might have changed size due to its child's child changing its text.

This also seems extremely suspect to me. That would mean in a traditional iOS application using UIKit changes to any property affecting layout on any view would have to manually traverse the ancestral chain calling `setNeedsLayout` on each.

Is there a reproducer of the problematic behavior in the samples that I can debug? Or is there one somewhere else internally?

JakeWharton:
[...]

PR #1175, opened 2023-06-07
Fix some stuff with UIKit

JakeWharton:
- Spacer should trigger layout on its parent when its size changes
- Eliminate boolean for tracking when layout is needed–the system does that for us
- Pass needs layout flag up to parent so it bottoms out in the system
- Request a layout pass and not a draw pass for children

Refs #1152

### DIFF
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
@@ -111,8 +111,6 @@ internal class UIViewFlexContainer(
   }
 
   private inner class FlexContainerHostView : UIScrollView(cValue { CGRectZero }) {
-    private var needsLayout = true
-
     override fun intrinsicContentSize(): CValue<CGSize> {
       return CGSizeMake(UIViewNoIntrinsicMetric, UIViewNoIntrinsicMetric)
     }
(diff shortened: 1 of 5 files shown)


redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -121,24 +119,22 @@ internal class UIViewFlexContainer(
...
-          height = container.items.maxOfOrNull { it.bottom } ?: 0.0,
-        ),
-      )
-      superview?.setNeedsLayout()
+      var newWidth = 0.0
+      var newHeight = 0.0
+      for (item in container.items) {
+        newWidth = maxOf(newWidth, item.right)
+        newHeight = maxOf(newHeight, item.bottom)
+      }
```
JakeWharton, line 131: Avoids boxing

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -121,24 +119,22 @@ internal class UIViewFlexContainer(
...
+      contentSize.useContents {
+        if (newWidth != width || newHeight != height) {
+          setContentSize(CGSizeMake(newWidth, newHeight))
+          superview?.setNeedsLayout()
+        }
```
JakeWharton, line 136: Trying to avoid two layout passes, if possible

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -121,24 +119,22 @@ internal class UIViewFlexContainer(
       return measure(size.useContents { toUnsafeSize() }).run { CGSizeMake(width, height) }
     }
 
-    override fun setNeedsLayout() {
-      needsLayout = true
```
JakeWharton, line 125: Adding a `super` call here fixed a bug, but `layoutSubviews` is only invoked when `setNeedsLayout` has been triggered so it turns out this can all be eliminated.

veyndan (approved):

PR #1183, opened 2023-06-08
Import a Kotlin common version of Yoga.

colinrtwhite:
This PR creates a `redwood-yoga` module that contains a Kotlin multiplatform implementation of Facebook's Yoga layout. It's based off of [this implementation](https://github.com/Joingo/yoga-kotlin) with a number of small tweaks and improvements. The entire implementation is marked as `internal`; the public API for the module is in a follow-up PR.

This engine fixes a lot of the edge cases we were seeing with the existing `redwood-flexbox` engine (especially on iOS) and is overall more robust based on my testing. The existing flexbox engine was converted from an Android layout whereas this implementation is platform agnostic.

Sorry in advance for the large PR! This PR strictly adds the `redwood-yoga` module with no other changes to make it easier to review. Long term it might be better to rely on the C++ implementation so it's easier to integrate upstream fixes, but this works for now.


redwood-yoga/build.gradle
```diff
@@ -0,0 +1,18 @@
+import app.cash.redwood.buildsupport.KmpTargets
+
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+apply plugin: 'kotlinx-atomicfu'
+apply plugin: 'com.vanniktech.maven.publish'
+apply plugin: 'org.jetbrains.dokka' // Must be applied here for publish plugin.
```
JakeWharton: Please use the new build plugin for this. Should be something like
```suggestion
apply plugin: 'app.cash.redwood.build'

redwoodBuild {
  publishing()
}
```

redwood-yoga/src/commonMain/kotlin/app/cash/redwood/yoga/internal/YGCachedMeasurement.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
```
JakeWharton, line 2: Oof. And I thought our company name was bad.

JakeWharton (approved):

redwood-yoga/build.gradle
```diff
@@ -19,3 +19,14 @@ kotlin {
...
   }
 }
+
+
+spotless {
+  kotlin {
+    targetExclude(
+      // Internal MIT-licensed files from Meta.
+      fileTree("src/commonMain/kotlin/app/cash/redwood/yoga/internal").filter { it.isFile() },
+      fileTree("src/commonTest/kotlin/app/cash/redwood/yoga/internal").filter { it.isFile() },
```
JakeWharton, line 28: No wildcard support? It's unclear from Spotless docs.

colinrtwhite: There doesn't appear to be unfortunately! `"src/**/kotlin/app/cash/redwood/yoga/internal"` is invalid.

PR #1185, opened 2023-06-08
Update redwood-layout-view to use Yoga engine.

colinrtwhite:
Updates the Android `View` implementation to use the new `redwood-yoga` engine.

~NOTE: CI will fail as some common methods are still required by the the other platform implementations (will add them tomorrow). The [final PR](https://github.com/cashapp/redwood/pull/1187) in the chain should pass.~


redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_columnWithJustifyContentCenter.png
JakeWharton, line 1: Are these screenshot changes all what we expect?

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_longRow.png
JakeWharton, line 1: These ones are the most surprising

redwood-layout-view/build.gradle
```diff
@@ -13,6 +13,7 @@ redwoodBuild {
 dependencies {
   api projects.redwoodLayoutWidget
   implementation projects.redwoodFlexbox
```
JakeWharton, line 14: I assume this module and all the dependencies on it will go away somewhere in a later change

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
```diff
@@ -0,0 +1,196 @@
...
+import android.view.View.MeasureSpec
+import android.view.ViewGroup
+import app.cash.redwood.yoga.MeasureCallback
+import app.cash.redwood.yoga.MeasureMode
+import app.cash.redwood.yoga.Node
+import app.cash.redwood.yoga.Size
+import kotlin.math.roundToInt
+
+internal class YogaLayout(context: Context) : ViewGroup(context) {
+  private val nodes = mutableMapOf<View, Node>()
```
JakeWharton: Might be worth an `ArrayMap` (or `SimpleArrayMap`) in the future.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
```diff
@@ -0,0 +1,196 @@
...
+
+  private fun View.asNode(): Node {
+    val childNode = Node()
+    childNode.measureCallback = ViewMeasureCallback(this)
+    applyLayoutParams(childNode, layoutParams)
+    return childNode
+  }
+}
+
+private class ViewMeasureCallback(val view: View) : MeasureCallback {
```
JakeWharton, line 180: This could be implemented directly on `YogaLayout` to save an allocation. Future change.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
```diff
@@ -0,0 +1,196 @@
...
+    }
+    super.removeAllViewsInLayout()
+  }
+
+  private fun removeViewFromYogaTree(view: View) {
+    val childNode = nodes[view] ?: return
+    val owner = childNode.owner ?: return
+
+    owner.children -= childNode
+    nodes -= view
```
JakeWharton: I'm not sure when only the second null check would fail, but I think we probably still want to remove from the map regardless, right?

```diff
     val childNode = nodes[view] ?: return
+    nodes -= view
     val owner = childNode.owner ?: return

     owner.children -= childNode
-    nodes -= view
```

Also, `remove()` returns the mapped value avoiding the need to perform the hashed lookup twice.
```suggestion
    val childNode = nodes.remove(view) ?: return

    val owner = childNode.owner ?: return
    owner.children -= childNode
```

JakeWharton (approved):

redwood-layout-view/build.gradle
colinrtwhite: Yep! There's a giant removal PR coming.

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_columnWithJustifyContentCenter.png
colinrtwhite: I believe so! Though it is slightly strange how any justify content forces the children to fill. The parent container is set to fill on both dimensions. This might be correct according to Yoga's logic.

It's worth noting that this with the new engine, the Compose UI and View screenshot tests match exactly whereas they didn't before.

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_longRow.png
colinrtwhite: This is mostly expected as the row is set to overflow. It's not expected how it forces the row to fill the available height.

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_columnWithJustifyContentCenter.png
colinrtwhite: This is actually due to a change in the default `AlignItems` value. The engine now defaults to `Stretch` - I updated all the tests to use `FlexStart`. This doesn't impact the Redwood column/row implementations since their property method is always called with `FlexStart` before being rendered.

It's worth noting that with the new engine, the Compose UI and View screenshot tests match exactly whereas they didn't before.

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_longRow.png
colinrtwhite: This was the same issue as mentioned [here](https://github.com/cashapp/redwood/pull/1185#discussion_r1226371057). The drawing offscreen is expected as the row defaults to overflow and won't compress the children.

gradle.properties
```diff
@@ -1,6 +1,7 @@
 kotlin.js.compiler=ir
 kotlin.mpp.androidSourceSetLayoutVersion=2
 kotlin.native.ignoreDisabledTargets=true
+kotlin.native.ignoreIncorrectDependencies=true
```
swankjesse: Comment for this ?

redwood-yoga/src/commonMain/kotlin/app/cash/redwood/yoga/Node.kt
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
```
swankjesse: I don’t believe you

swankjesse: (Unless this is our wrapper? I’ve been continuously frustrated by overzealous formatting plugins trying to change all my copyrights to this one.)

redwood-yoga/src/commonMain/kotlin/app/cash/redwood/yoga/internal/detail/RefObject.kt
```diff
@@ -0,0 +1,9 @@
+package app.cash.redwood.yoga.internal.detail
+
+//----------------------------------------------------------------------------------------
+//	Copyright © 2006 - 2021 Tangible Software Solutions, Inc.
+//	This class can be used by anyone provided that the copyright notice remains intact.
+//
+//	This class is used to replicate the ability to pass arguments by reference in Java.
```
swankjesse: It seems like this ability isn’t the feat it once was

swankjesse (approved):

redwood-yoga/src/commonMain/kotlin/app/cash/redwood/yoga/Node.kt
swankjesse: Upon closer inspection, this looks like it’s properly ours.

veyndan (approved):

PR #1189, opened 2023-06-08
Trigger navigation from EmojiSearch

jingwei99:
We want to exercise state saving, and for that
we need multiple activities.


samples/emoji-search/android-composeui/src/main/java/com/example/redwood/emojisearch/android/composeui/ComposeUiImage.kt
```diff
@@ -28,6 +28,7 @@ import com.example.redwood.emojisearch.widget.Image
 
 internal class ComposeUiImage : Image<@Composable () -> Unit> {
   private var url by mutableStateOf("")
+  private var onClick by mutableStateOf({})
```
JakeWharton, line 32: Looks like you're missing a `.clickable(onClick)` modifier on the `AsyncImage` below

JakeWharton (approved):

JakeWharton:
Also presumably needs iOS support.

jingwei99:
> Also presumably needs iOS support.

Yeah

PR #1191, opened 2023-06-08
Force update bounds of Counter label on text change

JakeWharton:


samples/counter/ios-shared/src/commonMain/kotlin/com/example/redwood/counter/ios/IosText.kt
```diff
@@ -32,5 +32,9 @@ class IosText : Text<UIView> {
 
   override fun text(text: String?) {
     value.text = text
+
+    // This very simple integration wraps the size of whatever text is entered. Calling
+    // this function will update the bounds and trigger relayout in the parent.
+    value.sizeToFit()
```
colinrtwhite, line 38: I think this is the missing piece I was missing with https://github.com/cashapp/redwood/issues/1152. We should probably add this to the emoji sample too.

colinrtwhite (approved):

samples/counter/ios-shared/src/commonMain/kotlin/com/example/redwood/counter/ios/IosText.kt
JakeWharton: Can do. I don't think it's the perfect solution for general usage, but it's certainly enough to achieve correctness at the cost of possibly doing two text layouts (since the parent may constrain the view and cause wrapping).

PR #1193, opened 2023-06-09
Add a settings plugin to apply our build plugin everywhere

JakeWharton:
The primary effect of this is that Spotless once again runs on every project. It will also allow migrating more of the convention-based configuration in the root `build.gradle` moving forward since this effectively replicates `allprojects {}`.

### DIFF
--- a/build-support/settings/README.md
+++ b/build-support/settings/README.md
@@ -1,0 +1,4 @@
+# Redwood Settings Plugin
+
+This is a Gradle plugin applied to the root `settings.gradle` via `includeBuild`
+which automatically adds the `build-support/` Gradle plugin to all projects.
--- a/build-support/settings/src/main/java/app/cash/redwood/buildsettings/RedwoodSettingsPlugin.java
+++ b/build-support/settings/src/main/java/app/cash/redwood/buildsettings/RedwoodSettingsPlugin.java
@@ -1,0 +1,17 @@
+package app.cash.redwood.buildsettings;
+
+import org.gradle.api.Plugin;
+import org.gradle.api.initialization.Settings;
+
+@SuppressWarnings("unused") // Invoked reflectively by Gradle.
+public class RedwoodSettingsPlugin implements Plugin<Settings> {
+  @Override public void apply(Settings target) {
+    target.getGradle().allprojects(project -> {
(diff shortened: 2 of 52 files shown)


build-support/settings/README.md
```diff
@@ -0,0 +1,4 @@
+# Redwood Settings Plugin
+
+This is a Gradle plugin applied to the root `settings.gradle` via `includeBuild`
+which automatically adds the `build-support/` Gradle plugin to all projects.
```
JakeWharton, line 4: Eventually this can likely live within the `build-support/` plugin itself. The reason it doesn't today is that use in the `settings.gradle` via the `plugins {}` DSL moves the plugin and its dependencies off of the buildscript classpath and onto the plugin classpath which breaks some things. If the main build eventually migrates to `plugins {}` then it will be possible to move.

build-support/settings/src/main/java/app/cash/redwood/buildsettings/RedwoodSettingsPlugin.java
```diff
@@ -0,0 +1,19 @@
...
+import org.gradle.api.initialization.Settings;
+import org.gradle.api.plugins.PluginContainer;
+
+@SuppressWarnings("unused") // Invoked reflectively by Gradle.
+public class RedwoodSettingsPlugin implements Plugin<Settings> {
+  @Override public void apply(Settings target) {
+    target.getGradle().beforeProject(project -> {
+      if (!project.getPath().equals(":")) {
+        var plugins = project.getPlugins();
+        plugins.withId("base", plugin -> {
```
JakeWharton: For some reason `project.beforeEvaluate()` isn't working here as the hook on when to apply the other plugin. The build scripts fail to find `redwoodBuild`. If I wait for the lifecycle base plugin though, it works.

¯\\\_(ツ)_/¯

JakeWharton: Made it work!

colinrtwhite (approved):

swankjesse (approved):

JakeWharton:
Also updated this one to remove from new yoga module

PR #1194, opened 2023-06-09
Fix Emoji Search cropping search bar

### LINKED ISSUE #770: `LazyColumn` doesn't play nicely with Flexbox
https://github.com/cashapp/redwood/pull/746#issuecomment-1401845828 and https://github.com/cashapp/redwood/pull/746#issuecomment-1404003920 (assuming these two are due to the same issue).

### LINKED ISSUE #964: [Sample][Emoji] Text input is clipped in Android views
This looks to be something a bit deeper into the Redwood view column impl.

| Pixel 6 emulator  | Pixel 3a emulator |
| ------------- | ------------- |
| Screenshot_1682325586 | Screenshot_1682325587 |

veyndan:
Fixes https://github.com/cashapp/redwood/issues/964. Closes https://github.com/cashapp/redwood/issues/770 as not a bug.

### DIFF
--- a/samples/emoji-search/presenter/src/commonMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearch.kt
+++ b/samples/emoji-search/presenter/src/commonMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearch.kt
@@ -111,6 +111,7 @@ fun EmojiSearch(
       state = searchTerm,
       hint = "Search",
       onChange = { searchTerm = it },
+      modifier = Modifier.shrink(0.0),
     )
     columnProvider.create(
       items = filteredEmojis,


JakeWharton (approved):
I don't really understand why this is the fix, but can't deny the screenshot!

veyndan:
[This line](https://github.com/cashapp/redwood/blob/b462dfd1cfc80266cf563d78b93e4938dfe0114d/redwood-flexbox/src/commonMain/kotlin/app/cash/redwood/flexbox/FlexContainer.kt#LL627C15-L627C15) was causing the height to shorten. Setting shrink to zero tells it not to shrink. This seems to be inline with the CSS flex-shrink property — at least I think that's what the [`Try It` section on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink#try_it) is trying to imply with `flex-shrink: 0`.

PR #1197, opened 2023-06-09
Add iOS change list renderer tests

apatronl:


swankjesse (approved):

redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/DefaultExampleSchemaWidgetFactory.kt
```diff
@@ -0,0 +1,36 @@
...
+import example.redwood.widget.Row
+import example.redwood.widget.ScopedRow
+import example.redwood.widget.Space
+import example.redwood.widget.Text
+import example.redwood.widget.TextInput
+import platform.UIKit.UIView
+
+class DefaultExampleSchemaWidgetFactory : ExampleSchemaWidgetFactory<UIView> {
+  override fun Row(): Row<UIView> = RowBinding()
+  override fun ScopedRow(): ScopedRow<UIView> = TODO("Not yet implemented")
```
swankjesse, line 30: We MIGHT want to change these to `error("unexpected call")` because I think there’s a robot that creates GitHub issues for all of our TODOs in this repo. I don’t think we ever want to actually do this work.

PR #1202, opened 2023-06-10
Eliminate properties used for publishing artifacts

JakeWharton:
The project name is now used as the `artifactId` automatically. These all matched the explicitly-specified ones from before.

I have eliminated the name which is now also set to the project name. Prior to this it was set to a more human-readable version of the `artifactId`.

This was tested by installing all artifacts to a local Maven repo both before and after this change and then diffing the directories. The only changes to the poms is the name change. For example:

    diff -r build/localMavenOld/app/cash/redwood/redwood-treehouse-js/0.5.0/redwood-treehouse-js-0.5.0.pom build/localMaven/app/cash/redwood/redwood-treehouse-js/0.5.0/redwood-treehouse-js-0.5.0.pom
    13c13
    <   <name>Redwood Treehouse</name>
    ---
    >   <name>redwood-treehouse</name>
    15a16

### DIFF
--- a/build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildPlugin.kt
+++ b/build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildPlugin.kt
@@ -24,9 +24,16 @@ import org.gradle.api.Project
 import org.gradle.api.plugins.JavaApplication
 import org.gradle.api.publish.PublishingExtension
 
+private const val redwoodGroupId = "app.cash.redwood"
+// HEY! If you change the major version update release.yaml doc folder.
+private const val redwoodVersion = "0.5.0-SNAPSHOT"
+
 @Suppress("unused") // Invoked reflectively by Gradle.
 class RedwoodBuildPlugin : Plugin<Project> {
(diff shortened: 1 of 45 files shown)


colinrtwhite (approved):
Nice! We can also delete the `gradle.properties` in the new `redwood-yoga` module I just added.

JakeWharton:
Yep. Done!

build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildPlugin.kt
```diff
@@ -96,6 +103,37 @@ private class RedwoodBuildExtensionImpl(private val project: Project) : RedwoodB
       if (project.providers.systemProperty("RELEASE_SIGNING_ENABLED").getOrElse("true").toBoolean()) {
         signAllPublications()
       }
+
+      coordinates(redwoodGroupId, project.name, redwoodVersion)
+
+      pom { pom ->
```
swankjesse, line 109: YAY

swankjesse (approved):
LOVE THIS MUCHLY

veyndan:
🥳

PR #1208, opened 2023-06-12
Rename some old 'java' directories to 'kotlin'

JakeWharton:
This is actually required for Spotless to target the Kotlin sources.


JakeWharton:
Will try to remove the Kotlin-only source folder targeting in a follow-up. Don't remember why we do that instead of doing all `*.kt` under `src/`.

veyndan (approved):

PR #1217, opened 2023-06-13
`redwood-layout` not abiding wrap constraint

veyndan:
@colinrtwhite This is more of a question with code than a PR!

For the test `shortColumn()`, I've set the height of the container to `Constraint.Wrap` (though I believe this is also the implicit value). I've also set the container background to blue.

With these changes, it doesn't appear that the flex container is abiding by the `Constraint.Wrap` constraint. Is this a bug?


colinrtwhite:
Hmm this might be a weird interaction with Paparazzi force stretching its children. I'll take a look.

veyndan:
In `ViewFlexContainer` it looks like the `width` nor `height` fields are being read, which could explain the reason it's not working for `redwood-layout-view` (https://github.com/cashapp/redwood/pull/1226).

Also, the `height` wasn't being correctly delegated (https://github.com/cashapp/redwood/pull/1225).

colinrtwhite:
@veyndan Whoops, good catch!

veyndan:
I've since fixed these in follow-up PRs. Constraints should be working as expected for Android View and Compose.

PR #1218, opened 2023-06-13
Add Paparazzi RTL tests to `redwood-layout-view`

veyndan:
The LTR screenshots are identical to the RTL ones, which is obviously wrong. Added a `TODO` comment so that a GitHub issue is created to remind us to implement it.


JakeWharton (approved):
Excellent

JakeWharton:
Going to do the same for Compose UI and iOS UIKit?

veyndan:
Will do it for Compose UI. Does iOS UIKit have any screenshot tests?

JakeWharton:
Not yet 😭 

I just talked to Kevin G about it. We've been trying to get it working so that we can write them in 100% Kotlin rather than needing a Swift wrapper application, but it's still not clear if it's going to work.

PR #1221, opened 2023-06-13
Add Paparazzi RTL tests to `redwood-lazylayout-view`

veyndan:
LazyLayout counterpart to #1218.

Didn't add a `TODO` as the actual RTL stuff looks okay. Some of the screenshots still look weird, but they don't look related RTL issues.


JakeWharton (approved):
I suppose these will get more interesting once LazyRow is a thing.

veyndan:
It already is! It's just that the width of each row is set to `MATCH_PARENT`, so they just need to be `WRAP_CONTENT`.

PR #1230, opened 2023-06-14
Change `Container`'s `LayoutParams` based on `RecyclerView` direction

veyndan:


redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -99,7 +99,14 @@ internal open class ViewLazyList(context: Context) : LazyList<View> {
 
   final override val placeholder = Placeholders(recyclerView.recycledViewPool)
 
-  private val linearLayoutManager = LinearLayoutManager(recyclerView.context)
+  private val linearLayoutManager = object : LinearLayoutManager(recyclerView.context) {
+    // Identical to the implementation of [LinearLayout.generateDefaultLayoutParams].
```
veyndan, line 103: With a name like `LinearLayoutManager`, one would expect that at least the implementation of `generateDefaultLayoutParams()` to be the same as `LinearLayout`'s.

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -99,7 +99,14 @@ internal open class ViewLazyList(context: Context) : LazyList<View> {
...
+    override fun generateDefaultLayoutParams(): RecyclerView.LayoutParams? = when (orientation) {
+      RecyclerView.HORIZONTAL -> RecyclerView.LayoutParams(WRAP_CONTENT, WRAP_CONTENT)
```
JakeWharton, line 105: Wrap, match?

JakeWharton (approved):

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
veyndan: In `LinearLayout` [they use `WRAP_CONTENT` for both directions](https://developer.android.com/reference/android/widget/LinearLayout#generateDefaultLayoutParams()) when the orientation is horizontal. We definitely don't have to match what they do. I just want it to be the least surprising default.

JakeWharton: Yeah. It seems weird to be asymmetrical, but I don't really have a strong opinion on what is the right thing to do. I suspect most important is to be consistent across all of our platforms.

PR #1232, opened 2023-06-14
Add default event listener for Android

jingwei99:


redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/treehouseAppFactoryAndroid.kt
```diff
@@ -34,7 +36,7 @@ public fun TreehouseAppFactory(
   context: Context,
   httpClient: OkHttpClient,
   manifestVerifier: ManifestVerifier,
-  eventListener: EventListener = EventListener(),
+  eventListener: EventListener = defaultEventListener,
```
jingwei99: An alternative is adding this to `TreehouseApp.kt`, but iOS app is not building right now, so adding it here to start with

JakeWharton (approved):
I believe this event listener implementation should be removed from Treehouse and sent into some downstream project. Logging to Android's Log should never happen in production and should only happen in development. We can actually control that here with variants, but in our app we don't actually want the logs to ever go into Log as we have our own abstraction that we want to use instead. The default should be a no-op/empty implementation.

jingwei99:
> I believe this event listener implementation should be removed from Treehouse and sent into some downstream project. Logging to Android's Log should never happen in production and should only happen in development. We can actually control that here with variants, but in our app we don't actually want the logs to ever go into Log as we have our own abstraction that we want to use instead. The default should be a no-op/empty implementation.

I made this change bc I was working with @swankjesse and it was really helpful to have the logs on to see what's going on with emoji search. 

> The default should be a no-op/empty implementation.

👌

another option is to move this to the Android emoji search app (and then iOS when I get to it)? that'll satisfy the needs (of what motivated this change), what do you think

JakeWharton:
> another option is to move this to the Android emoji search app (and then iOS when I get to it)? that'll satisfy the needs (of what motivated this change), what do you think

Yeah. Absolutely. That sounds perfect. It's definitely useful to have in the samples to demonstrate how things work.

PR #1234, opened 2023-06-15
Add support for `LazyList` margin

veyndan:


redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -150,6 +154,17 @@ internal open class ViewLazyList(context: Context) : LazyList<View> {
     }
   }
 
+  override fun margin(margin: Margin) {
+    with(density) {
+      recyclerView.updatePaddingRelative(
```
veyndan, line 159: It's implemented using padding, as [it appears that](https://github.com/cashapp/redwood/issues/855#issuecomment-1593025007) that's what yoga is doing.

swankjesse: We’re attempting to do CSS box model so padding is technically incorrect - the background color paints padding but not margins. 

But this works for me, for now.

veyndan: The comment that I left in the linked issue is implying that we should rename `margin` back to `padding`, as that's what it technically does. If we want to support margin, that should be a wholly separate thing.

swankjesse: Well, in `Column` I believe it is implemented as a Margin. 

@colinrtwhite do I have this right?

swankjesse (approved):

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
swankjesse: (The best way to be sure is to paint the background of the component and see!)

veyndan: @swankjesse I think we've come full circle 😅 

Screenshot 2023-06-15 at 17 04 15

colinrtwhite: @swankjesse @veyndan is right! [Check out this comment](https://github.com/cashapp/redwood/issues/855#issuecomment-1593540122). Since we can't set a background on the Row/Column and can't size things beyond wrap/fill this is effectively the same as margin currently.

To support margin on a non-child, we'll have to implement it natively for each platform instead of common code!

PR #1237, opened 2023-06-16
Fix infinite recursion when measuring an empty YogaLayout.

colinrtwhite:


redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
```diff
@@ -24,14 +24,10 @@ internal class YogaLayout(context: Context) : ViewGroup(context) {
   var applyModifier: (Node, Int) -> Unit = { _, _ -> }
 
   init {
-    rootNode.measureCallback = ViewMeasureCallback(this)
```
colinrtwhite, line 27: This is the main change. The `rootNode` should never have an associated measure function or it will call its own `measure` infinitely.

JakeWharton (approved):

swankjesse (approved):

PR #1240, opened 2023-06-16
Migrate UIViewLazyList to CollectionView

### LINKED ISSUE #1005: Add support for horizontal scrolling in `UIViewLazyList`

underscoretang:
A first pass at converting this into a CollectionView. It's in a decent place, but there are a few more things to be hammered out:

I put in some logs to illustrate some of the issues here, particularly with the "window" of viewport items not keeping up with the scrolling of the collection view.

🟤 – Redwood bindings
🟢 – CollectionView datasource
🔵 – Viewport Changes
🔴 – Out of range errors

https://github.com/cashapp/redwood/assets/465849/ed1528db-626d-4787-a03d-20987bea6b07

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -26,101 +26,194 @@ import app.cash.redwood.layout.api.Constraint
 import app.cash.redwood.lazylayout.widget.LazyList
 import app.cash.redwood.lazylayout.widget.RefreshableLazyList
 import app.cash.redwood.ui.Margin
+import app.cash.redwood.widget.ChangeListener
 import app.cash.redwood.widget.MutableListChildren
 import app.cash.redwood.widget.Widget
+import kotlin.math.max
+import kotlinx.cinterop.CValue
 import kotlinx.cinterop.ObjCClass
+import kotlinx.cinterop.readValue
+import kotlinx.cinterop.useContents
+import platform.CoreGraphics.CGRect
+import platform.CoreGraphics.CGRectZero
+import platform.CoreGraphics.CGSize
+import platform.CoreGraphics.CGSizeMake
 import platform.Foundation.NSIndexPath
 import platform.Foundation.classForCoder
+import platform.UIKit.UICollectionView
+import platform.UIKit.UICollectionViewCell
+import platform.UIKit.UICollectionViewDataSourceProtocol
+import platform.UIKit.UICollectionViewDelegateFlowLayoutProtocol
+import platform.UIKit.UICollectionViewFlowLayout
+import platform.UIKit.UICollectionViewLayout
+import platform.UIKit.UICollectionViewScrollDirection
 import platform.UIKit.UIControlEventValueChanged
+import platform.UIKit.UIEdgeInsetsMake
 import platform.UIKit.UIRefreshControl
-import platform.UIKit.UITableView
-import platform.UIKit.UITableViewCell
-import platform.UIKit.UITableViewCellStyle
-import platform.UIKit.UITableViewCellStyle.UITableViewCellStyleDefault
-import platform.UIKit.UITableViewDataSourceProtocol
-import platform.UIKit.UITableViewDelegateProtocol
+import platform.UIKit.UIScrollView
 import platform.UIKit.UIView
 import platform.UIKit.item
 import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
-private const val reuseIdentifier = "cell"
+internal interface ViewPortItems : Widget.Children<UIView> {
+  var placeholders: MutableList<Widget<UIView>>
+  var itemsBefore: Int
+  var itemsAfter: Int
+  fun itemForGlobalIndex(index: Int): UIView
+  fun itemCount(): Int
+}
 
-internal open class UIViewLazyList() : LazyList<UIView> {
+internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {
 
-  private val itemsList = mutableListOf<Widget<UIView>>()
+  override val items: ViewPortItems = object : ViewPortItems {
+    override var placeholders = mutableListOf<Widget<UIView>>()
 
-  override val placeholder: Widget.Children<UIView> = MutableListChildren()
+    override var itemsBefore: Int = 0
+    override var itemsAfter: Int = 0
 
-  override val items: Widget.Children<UIView> = object : Widget.Children<UIView> {
+    private val viewPortList = mutableListOf<Widget<UIView>>()
+
     override fun insert(index: Int, widget: Widget<UIView>) {
-      itemsList.add(index, widget)
-      tableView.reloadData()
+      viewPortList.add(index, widget)
     }
 
     override fun move(fromIndex: Int, toIndex: Int, count: Int) {
-      itemsList.move(fromIndex, toIndex, count)
-      tableView.reloadData()
+      viewPortList.move(fromIndex, toIndex, count)
     }
 
     override fun remove(index: Int, count: Int) {
-      itemsList.remove(index, count)
-      tableView.reloadData()
+      viewPortList.remove(index, count = count)
     }
 
-    override fun onModifierUpdated() {
+    override fun onModifierUpdated() {}
+
+    // Fetch the item from the viewPortList relative to the entire collection view
+    override fun itemForGlobalIndex(index: Int): UIView {
+      val viewPortIndex: Int = max(index - itemsBefore, 0)
+
+      viewPortList.getOrNull(viewPortIndex)?.let {
+        return it.value
+      }
+
+      // If we don't have a value, fallback to our pools of placeholders
+      val placeholderIndex = index % placeholders.size
+      return placeholders.get(placeholderIndex).value
+    }
+
+    override fun itemCount(): Int {
+      // TODO: replace with this with "ItemCounts"
+      return max(itemsBefore - 1, 0) + viewPortList.count() + itemsAfter
     }
   }
 
-  private val tableViewDelegate: UITableViewDelegateProtocol =
-    object : NSObject(), UITableViewDelegateProtocol {
-      override fun tableView(
-        tableView: UITableView,
-        willDisplayCell: UITableViewCell,
-        forRowAtIndexPath: NSIndexPath,
-      ) {
-        val content = itemsList[forRowAtIndexPath.item.toInt()]
-        (willDisplayCell as Cell).setView(content.value)
+  override val placeholder: Widget.Children<UIView> = MutableListChildren(list = items.placeholders)
+
+  private val viewPortListCoordinator = object {
+    var minIndex: Int = 0
+    var maxIndex: Int = 0
+
+    fun notifyViewportChanged() {
+      onViewportChanged(minIndex, maxIndex)
+    }
+
+    fun updateViewport(minIndex: Int, maxIndex: Int) {
+      if (minIndex != this.minIndex || maxIndex != this.maxIndex) {
+        this.minIndex = minIndex
+        this.maxIndex = maxIndex
+
+        notifyViewportChanged()
+      }
+    }
+  }
+
+  // A callback to tell LazyList that we have an update to the window of items available
+  private lateinit var onViewportChanged: (firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) -> Unit
+
+  // UICollectionView + Protocols
+  private var collectionViewFlowLayout: UICollectionViewFlowLayout =
+    object : UICollectionViewFlowLayout() {}
+
+  private val collectionViewDataSource: UICollectionViewDataSourceProtocol =
+    object : NSObject(), UICollectionViewDataSourceProtocol {
+
+      override fun collectionView(
+        collectionView: UICollectionView,
+        numberOfItemsInSection: NSInteger,
+      ): NSInteger {
+        return items.itemCount().toLong()
+      }
+
+      override fun collectionView(
+        collectionView: UICollectionView,
+        cellForItemAtIndexPath: NSIndexPath,
+      ): UICollectionViewCell {
+        val cell = collectionView.dequeueReusableCellWithReuseIdentifier(
+          identifier = reuseIdentifier,
+          forIndexPath = cellForItemAtIndexPath,
+        ) as LazyListContainerCell
+
+        val view = items.itemForGlobalIndex(cellForItemAtIndexPath.item.toInt())
+        cell.set(view)
+
+        return cell
       }
     }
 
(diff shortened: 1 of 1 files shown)


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -28,128 +28,284 @@ import app.cash.redwood.lazylayout.widget.RefreshableLazyList
...
   }
 
   override fun onViewportChanged(onViewportChanged: (firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) -> Unit) {
+    println("🟤 onViewportChanged")
     this.onViewportChanged = onViewportChanged
   }
 
   override fun itemsBefore(itemsBefore: Int) {
-    this.itemsBefore = itemsBefore
+    println("🟤 itemsBefore $itemsBefore")
```
underscoretang: I think something that we should consider is having a way for us to pass along a full transaction or model, instead of individual binding callbacks.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -28,128 +28,284 @@ import app.cash.redwood.lazylayout.widget.RefreshableLazyList
...
+        val cell = collectionView.dequeueReusableCellWithReuseIdentifier(
+          identifier = reuseIdentifier,
+          forIndexPath = cellForItemAtIndexPath
+        ) as LazyListContainerCell
 
-      override fun tableView(
-        tableView: UITableView,
-        numberOfRowsInSection: NSInteger,
-      ): NSInteger = itemsList.size.toLong()
+        val widget = items.itemForGlobalIndex(cellForItemAtIndexPath.item.toInt())
```
underscoretang: @veyndan, what if instead of having to manage a list on this end of things. we just have a reference that can be used here and then have the viewport and recycling logic happen on the Redwood LazyList side? This way we can save ourselves the several round trips to make sure we have the item

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -28,128 +28,284 @@ import app.cash.redwood.lazylayout.widget.RefreshableLazyList
...
-import platform.UIKit.UITableViewDelegateProtocol
+import platform.UIKit.UIScrollView
 import platform.UIKit.UIView
+import platform.UIKit.indexPathForRow
 import platform.UIKit.item
 import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
-private const val reuseIdentifier = "cell"
+internal interface Items: Widget.Children<UIView> {
```
dnagler: naming nit: this is kind of a state machine, so maybe something like `ListItemState`?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -28,128 +28,284 @@ import app.cash.redwood.lazylayout.widget.RefreshableLazyList
...
-      tableView.reloadData()
+      println("🟤 insert $index")
+      viewPortList.add(index, widget)
+
+      if (isFinishedInitialLoad) {
+        val indexPath = globalIndexPathForViewportIndex(index)
+       // collectionView.reloadItemsAtIndexPaths(indexPaths = listOf(indexPath))
+      }
+
+      // TODO: We need a signal that the items are done with their initial load..
```
dnagler: Is the idea that the contents of the LazyList will always be paginated with a strict page size?
If so, we could enforce that here--otherwise, it's pretty hard to tell if our "initial load" is done in general.
We could maybe do something along the lines of asking "do the currently visible cells fill the available frame for the view, or is there a gap at the bottom", but that feels like guesswork.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
JakeWharton: The current design is imposed on us by the underlying Compose mechanism. We can't really change it because that would break direct usage (i.e., when not using Treehouse).

You can, however, implement `ChangeListener` on a widget type which gives you a callback after each batch of changes. This should effectively allow you to create the same thing. Instead of doing any action in these individual functions, instead, write the state to properties and then only act when that callback occurs.

veyndan: What @underscoretang said sounds similar to [what I said in the `ChangeListener` proposal issue](https://github.com/cashapp/redwood/issues/1133#issuecomment-1561666108). Considering all three change most of the time, what's the downside of that approach (apart from the class wrapper)?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
veyndan: I'm not totally sure what you mean by:

> we just have a reference that can be used here

veyndan (commented):
This is looking awesome!

> I put in some logs to illustrate some of the issues here, particularly with the "window" of viewport items not keeping up with the scrolling of the collection view.

When this happens, we should display placeholders that are swapped out when the window has caught up with the viewport.

Also, is it possible to do the window shifting in a separate PR from the `CollectionView` migration? I'm just concerned that the PR could get unwieldy if we try and fix everything!

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
underscoretang: i meant migrating over the `Items` object to be handled by redwood directly so that we can just give the LazyList implementations a callback for 

`itemAtIndex(index: Index) -> Item` and not have to do state management here. but this probably comes with a performance cost?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
underscoretang: going to try using the changeListener approach here. the hope is to onlyReload and update the views when all the changes are committed, otherwise we end up with a bunch of intermediate updates that might not have all the state changes ready (esp on agressively scrolling)

underscoretang:
This is ready to merge and works with placeholders as well
https://github.com/cashapp/redwood/assets/465849/ec5a58f9-9e75-4033-8f4b-71bf786c45fb


1. Using `ChangeListener` to determine when to call `collectionView.reloadData()` is perfect
2. Previously, we were removing the `this.view` on the cell, which was a problem because that could be referencing a view that was already moved to another cell. Instead, we're just blowing away the contentViews subviews on prepare for reuse.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
underscoretang: Went with `ViewPortItems` to be a little clearer. it's still a datasource afterall. just with a moving window

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -26,101 +26,197 @@ import app.cash.redwood.layout.api.Constraint
...
+      override fun collectionView(
+        collectionView: UICollectionView,
+        cellForItemAtIndexPath: NSIndexPath,
+      ): UICollectionViewCell {
+        val cell = collectionView.dequeueReusableCellWithReuseIdentifier(
+          identifier = reuseIdentifier,
+          forIndexPath = cellForItemAtIndexPath
+        ) as LazyListContainerCell
+
+        val view = items.itemForGlobalIndex(cellForItemAtIndexPath.item.toInt())
```
dnagler, line 156: This is pretty funky, but if it works it works

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -26,101 +26,197 @@ import app.cash.redwood.layout.api.Constraint
...
-      rowHeight = 64.0 // TODO: size rows by their content.
-      prefetchingEnabled = false
+  private val collectionViewDelegate: UICollectionViewDelegateFlowLayoutProtocol =
+    object: NSObject(), UICollectionViewDelegateFlowLayoutProtocol {
+      override fun collectionView(
+        collectionView: UICollectionView,
+        layout: UICollectionViewLayout,
+        sizeForItemAtIndexPath: NSIndexPath
+      ): CValue<CGSize> {
+        // TODO: Calculate size for individual items. This commented code currently hangs the UI Thread
```
dnagler, line 170: What do you think will be the workaround here? Since we're implementing a synchronous method on the protocol, we can't really be hopping around to different threads.
Would calling `sizeToFit` (instead of `sizeThatFits`) here, or within `cellForItemAtIndexPath` work, or is the issue that this is always going to be called on a background thread?

dnagler (approved):
Seems like a step in the right direction!

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
veyndan: The logic would have to continue to live on the host side, otherwise having `Items` on the guest side would involve  a lot of Zipline serialization — something which we had before and was causing a lot of performance headaches.

Perhaps we could share the implementation across Android View's though? `ViewLazyList` [has a similar-ish concept](https://github.com/cashapp/redwood/blob/e3b9bc927b1496c02264dcafce4f9e6e68b7d6e3/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt#L66-L94) to what you've got here. Can probably be a follow-up though!

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -26,101 +26,197 @@ import app.cash.redwood.layout.api.Constraint
...
+
+      // If we don't have a value, fallback to our pools of placeholders
+      val placeholderIndex = index % placeholders.size
+      return placeholders.get(placeholderIndex).value
     }
 
-    override fun onModifierUpdated() {
+    override fun itemCount(): Int {
+      // TODO: replace with this with "ItemCounts"
+      return max(itemsBefore - 1, 0 ) + viewPortList.count() + itemsAfter
```
veyndan: I'm curious about the `max(itemsBefore - 1, 0 )` bit. Why do we want 1 less than `itemsBefore`?

Essentially, I would've thought it'd be:

```suggestion
      return itemsBefore + viewPortList.count() + itemsAfter
```

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -26,101 +26,197 @@ import app.cash.redwood.layout.api.Constraint
...
-import platform.UIKit.UITableViewDelegateProtocol
+import platform.UIKit.UIScrollView
 import platform.UIKit.UIView
+import platform.UIKit.indexPathForRow
 import platform.UIKit.item
 import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
-private const val reuseIdentifier = "cell"
+internal interface ViewPortItems: Widget.Children<UIView> {
```
veyndan: [nit] I don't think the interface is giving us much. It can probably be safely removed, but it can be done in a follow-up PR

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -26,101 +26,197 @@ import app.cash.redwood.layout.api.Constraint
...
+    override fun itemForGlobalIndex(index: Int): UIView {
+      val viewPortIndex: Int = max(index - itemsBefore, 0)
+
+      viewPortList.getOrNull(viewPortIndex)?.let {
+        return it.value
+      }
+
+      // If we don't have a value, fallback to our pools of placeholders
+      val placeholderIndex = index % placeholders.size
+      return placeholders.get(placeholderIndex).value
```
veyndan, line 101: When the `placeholders.size` is less than the number of items that can fill the view port, we might be reusing an already visible placeholder. Is that a big deal on iOS?

veyndan (approved):
I'm curious how this'd look in Repo Search, as the number of elements increases as you scroll (unlike in Emoji Search). Unfortunately we don't have Repo Search implemented on iOS yet though 😞

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
underscoretang: it's not ideal, but i think it beats maintaining a pool for a really niche case where we have many small cells. we could use the viewport size to increase the initial pool size if needed

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
underscoretang: i think i was seeing an off by 1 error, but i think i got that mixed up with the index. This will get replaced with the ItemCounts anyways

PR #1257, opened 2023-06-23
Restore an overload for backwards-compatibility

swankjesse:


redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineTreehouseUi.kt
```diff
@@ -33,5 +33,15 @@ public interface ZiplineTreehouseUi : ZiplineService, EventSink {
     uiConfigurations: StateFlow<UiConfiguration>,
     stateSnapshot: StateSnapshot?,
   )
+
+  @Deprecated(
+    "delete after 2023-07-23",
```
swankjesse, line 38: How long should we support backwards-incompatible Redwood internals before 1.0 ?

JakeWharton: Once 0.5 drops it has to be all the way to 1.0.

But since 0.5 hasn't dropped, why are we bothering to do this at all?

jingwei99 (approved):

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineTreehouseUi.kt
swankjesse: We’ve got some employees using it already!

JakeWharton: Then they are doing so without a guarantee of stability and breaking them is perfectly within our right.

PR #1274, opened 2023-06-26
Parameterize `AlignItem` in layout test suite

veyndan:


redwood-layout-shared-test/src/main/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -213,3 +186,19 @@ private val movies = listOf(
 private data class CrossAxisAlignmentImpl(
   override val alignment: CrossAxisAlignment,
 ) : HorizontalAlignment, VerticalAlignment
+
+enum class AlignItemsEnum {
```
veyndan, line 190: I tried implementing a `TestParameterValuesProvider` and enumerating the values there, but TestParameterInjector would fail with a type-mismatch error. I'm assuming this is because `FlexDirection` is a value class. This is just a quick-and-dirty workaround.

JakeWharton (approved):

Issue #1261, opened 2023-06-23
Add `CrossAxisAlignment` property to `LazyList`


veyndan:
To better align with the properties `Row` and `Column` provide, we should add `LazyColumn#horizontalAlignment` and `LazyRow#verticalAlignment`, both with default values of `CrossAxisAlignment.Start`.

PR #1277, opened 2023-06-26
Add `CrossAxisAlignment` property to `LazyList`

veyndan:
Closes https://github.com/cashapp/redwood/issues/1261


redwood-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/lazylayout/composeui/ComposeUiLazyList.kt
```diff
@@ -102,9 +104,14 @@ internal class ComposeUiLazyList :
...
   }
 
+  override fun crossAxisAlignment(crossAxisAlignment: CrossAxisAlignment) {
+    this.crossAxisAlignment = crossAxisAlignment
+  }
+
   override val value = @Composable {
     val content: LazyListScope.() -> Unit = {
       items(items.widgets) { item ->
+        // TODO If CrossAxisAlignment is Stretch, pass Modifier.fillParentMaxWidth() to child widget.
```
veyndan, line 114: Any idea how to do this? A further complication is that `Modifier.fillParentMaxWidth` is only available in the `LazyItemScope` (which the `items` block provides).

JakeWharton: Does not seem possible with the current design.

We could change the widget type for Compose UI from `@Composable () -> Unit` to `@Composable (Modifier) -> Unit` but that has huge rippling effects and I'm not actually sure it's even the right thing to do. It would mean that every Compose UI widget would need to accept a `Modifier` from its parent and attach it appropriately to itself.

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -165,6 +166,10 @@ internal open class ViewLazyList(context: Context) : LazyList<View> {
     }
   }
 
+  override fun crossAxisAlignment(crossAxisAlignment: CrossAxisAlignment) {
+    adapter.crossAxisAlignment = crossAxisAlignment
```
JakeWharton, line 170: Need to somehow trigger all of the views to be recreated when this changes

JakeWharton (approved):

redwood-lazylayout-composeui/src/commonMain/kotlin/app/cash/redwood/lazylayout/composeui/ComposeUiLazyList.kt
veyndan: Yep, agreed that propagating a `Modifier` isn't the best solution. I'll just punt this to a future PR.

JakeWharton (approved):

PR #1279, opened 2023-06-26
Add Width and Height modifiers for controlling the size of a widget.

### LINKED ISSUE #855: Width and Height configuration for children of Rows and Columns

colinrtwhite:
Adds modifiers to force specific width and/or height values for children of `Row`/`Column`. Example:

```kotlin
Row {
    Image(
        url = url,
        modifier = Modifier.size(20.dp, 20.dp),
    )
    Text("Some text")
}
```

~Open question: should we also add a `Size` modifier that sets both dimensions at once?~

Resolves: https://github.com/cashapp/redwood/issues/855


JakeWharton:
> Open question: should we also add a Size modifier that sets both dimensions at once?

Compose UI has one. Looks like they all bottom out in a size element. We could just have a size function that combines the two.

JakeWharton (approved):
What's up with the two blank screenshots?

colinrtwhite:
Hmm seems like a bug. Will take a look today.

Also I'll add a `size` function that combines the two modifiers.

veyndan (approved):
🎉 

We should also implement this in `LazyList`. I can do that in a follow-up.

colinrtwhite:
Unfortunately we can't add scoped modifier extensions without context receivers as they require two receivers: `RowScope`/`ColumnScope` and `Modifier`. I'm going to add a `Size` modifier to the schema instead.

colinrtwhite:
The Compose UI issue was caused by `placeable` not being set. This is because the measure callback is skipped if both width and height are fixed.

PR #1285, opened 2023-06-28
Add `Row`/`LazyRow` screenshot tests for `align-self`

veyndan:
The tests are incorrect for `LazyRow` because the `align-self` property isn't supported yet. Surprisingly, the tests are incorrect for `Row` when targetting Android Views, but not Compose UI. Will fix both in follow-up PRs.


redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_layoutWithMarginAndDifferentAlignments[LTR,Row].png
Comment (veyndan) [inline redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_layoutWithMarginAndDifferentAlignments[LTR,Row].png:1]: There are 4 `CrossAxisAlignment` values we want to test: `Start`, `Center`, `End`, `Stretch`. As the items are cut-off in the screenshot tests, we can only see that the first 3 of these values are correctly applied.

Will pick shorter film titles in a follow-up, so more items can fit on the screen.

Comment (veyndan) [reply redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_layoutWithMarginAndDifferentAlignments[LTR,Row].png:1]: https://github.com/cashapp/redwood/pull/1286

JakeWharton (approved):

PR #1287, opened 2023-06-28
Collapse `HorizontalAlignment` and `VerticalAlignment` modifiers into one

veyndan:
In Compose UI, the `Row`/`Column` scope offers a unified `align` modifier. In CSS, this is the same as `align-self`.

If someone changes from a `Row` to a `Column`, I wouldn't also expect that the children modifier names to be changed (previously it would've been from `verticalAlignment` to `horizontalAlignment`.


JakeWharton (approved):

colinrtwhite:
@veyndan I'd argue that we shouldn't make this change as it changes the public API names to be less clear:

```kotlin
Column {
    Image(url, modifier = Modifier.alignment(CrossAxisAlignment.Center)
}
```

Is less clear because it doesn't specify if we're aligning on the horizontal or vertical axis:

```kotlin
Column {
    Image(url, modifier = Modifier.horizontalAlignment(CrossAxisAlignment.Center)
}
```

I could see someone applying `alignment` thinking it changes the vertical alignment in a `Column`.

EDIT: Thought of some more reasons:

- We lose consistency with the `horizontalAlignment` and `verticalAlignment` arguments for the `Row` and `Column` functions.
- When we eventually support a `Box` type we'll need to create separate `HorizontalAlignment` and `VerticalAlignment` modifiers anyways since we need to be able to configure the alignment for an individual item across both dimensions.

veyndan:
@colinrtwhite Responded at https://github.com/cashapp/redwood/pull/1292#issuecomment-1612126907. Happy to have a video chat about this too!

PR #1290, opened 2023-06-28
Add screenshot tests for varying `width` and `height`

veyndan:
Very few of these screenshots are actually correct. These screenshots rely on `CrossAxisAligment.Stretch` being correctly applied. I tried passing a background to the flex container, but that's not possible in Compose UI (unless we expose the `Modifier` for testing purposes only — not the best).

### DIFF
--- a/redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_layoutWithConstraints[LTR,Column,Fill,Fill].png
+++ b/redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_layoutWithConstraints[LTR,Column,Fill,Fill].png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:5d8c5191a9fb3a12ef22d5fdee277a06a9b71349584d52213dfba9e93bf62c1c
+size 6020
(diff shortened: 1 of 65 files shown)


redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_layoutWithConstraints[LTR,Column,Fill,Fill].png
Comment (veyndan) [inline redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_layoutWithConstraints[LTR,Column,Fill,Fill].png:1]: I don't think these are good tests. It exaggerates the number of issues with size constraints. For example, modifying the background of the container to be blue shows that this test is in fact correct, whereas judging by the current screenshot, there's no way of telling if the container applied a height of `Wrap` or `Fill`.

I'm gonna redo these tests.

image

JakeWharton (approved):

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_layoutWithConstraints[LTR,Column,Fill,Fill].png
Comment (veyndan) [reply redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_layoutWithConstraints[LTR,Column,Fill,Fill].png:1]: `trunk` now displays a blue background for the container, so this is no longer a problem.

PR #1291, opened 2023-06-28
Add background color to container in tests

veyndan:
I want to make sure that the `width` and `height` of the container are as expected. I just added a blue background color to the container so we can verify this.


redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiFlexContainer.kt
```diff
@@ -158,6 +160,7 @@ internal class ComposeUiFlexContainer(
         modifier = modifier.verticalScroll(rememberScrollState())
       }
     }
+    modifier = modifier.then(testOnlyModifier)
```
veyndan: I really dislike having test specific code in production code

JakeWharton: If you want to reduce its impact, using a nullable `Modifier` would actually be better. Most shrinking tools on any platform will see the value is never written and will inline the `null`. Then the `if` statement around it would become a constant and get completely eliminated.

But it's not a huge deal either way.

JakeWharton (approved):

colinrtwhite (approved):

redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiFlexContainer.kt
veyndan: Oooo good to know! Updated the PR using a nullable `Modifier`

PR #1292, opened 2023-06-28
Revert "Collapse `HorizontalAlignment` and `VerticalAlignment` modifiers into one (#1287)"

colinrtwhite:
This reverts commit 794ae03db30498b8af84bcc5ecf9c4103b824dc7.

I'd argue that we shouldn't make this change for a couple reasons:

- We lose consistency with the `horizontalAlignment` and `verticalAlignment` arguments for the `Row` and `Column` functions.

- When we support a `Box` type we'll need to create separate `HorizontalAlignment` and `VerticalAlignment` modifiers anyways since we need to be able to configure the alignment for an individual item across both dimensions:

```kotlin
Box {
    Image(
        url = url,
        modifier = Modifier
            .horizontalAlignment(CrossAxisAlignment.Center)
            .verticalAlignment(CrossAxisAlignment.Center),
    )
}
```

- It changes the public API names to be less clear:

```kotlin
Column {
    Image(url, modifier = Modifier.alignment(CrossAxisAlignment.Center)
}
```

It's less clear because it doesn't specify if we're aligning on the horizontal or vertical axis:

```kotlin
Column {
    Image(url, modifier = Modifier.horizontalAlignment(CrossAxisAlignment.Center)
}
```

I could see someone applying `alignment` thinking it changes the vertical alignment in a `Column`.


JakeWharton (approved):

veyndan:
The main motivation for such a change is to implement #1261. The `LazyList` scope doesn't distinguish between whether it's a `LazyColumn` or `LazyRow`, and personally I don't believe it needs to be.

> We lose consistency with the `horizontalAlignment` and `verticalAlignment` arguments for the `Row` and `Column` functions.

I'd argue that we _gain_ consistency with two platforms that people are comfortable with — Android and the web (I can't speak for iOS).

- In Compose UI, in order to align a single item within a `Row`/`Column`, we call `Modifier.align`. They also have this "inconsistency" where `Row`/`Column` gets passed a `horizontalAlignment`/`verticalAlignment` respectively, despite the children calling `align`.
- In web, in order to align a single item within a flex element in either direction, we call `align-self`.

None of these state which direction we need to put it in, because it is evident from the parent.

> It changes the public API names to be less clear

The example you gave is almost exactly the same as what Compose UI does.

> It's less clear because it doesn't specify if we're aligning on the horizontal or vertical axis

I think we should use the CSS definition of alignment (as much of the `Row`/`Column` API is CSS-y).  If we're in a `Row`, you can only _align_ on the vertical axis. By definition, alignment on the horizontal axis doesn't make sense, as moving stuff along the horizontal axis would be to "justify" it (and perhaps we should rename the `MainAxisAligment` properties too).

colinrtwhite:
> The main motivation for such a change is to implement https://github.com/cashapp/redwood/issues/1261. The `LazyList` scope doesn't distinguish between whether it's a `LazyColumn` or `LazyRow`, and personally I don't believe it needs to be.

Ah gotcha, I agree it doesn't have to use separate scopes, however I think it's better as an API consumer to have separate `Modifier` names. Using a single scope for `LazyRow` and `LazyColumn` also locks us into no direction specific modifiers in the future (due to binary compatibility) even if it's unlikely.

> I'd argue that we gain consistency with two platforms that people are comfortable with — Android and the web (I can't speak for iOS).
> The example you gave is almost exactly the same as what Compose UI does.

Compose UI uses `align`, however the major difference is the values passed to `align` imply horizontal or vertical alignment (e.g. `Top`, `CenterVertically`, `Start`, `CenterHorizontally`). `CrossAxisAlignment` uses `Start` and `End` for vertical alignment as well. As a result, we should either change the values passed to `alignment` so they imply direction (and come up with a proposal) or we should keep the directions in the method names.

> I think we should use the CSS definition of alignment

I disagree. The CSS names for horizontal and vertical alignment aren't very intuitive. It's impossible to know which direction `Modifier.justifyContent` changes without understanding the CSS spec. `Modifier.horizontalAlignment` is clear about what it does without searching.

As mentioned, we will need two separate `horizontalAlignment` and `verticalAlignment` modifiers for `Box` so it's better to use `horizontalAlignment` and `verticalAlignment` for our other layout components to remain consistent.

veyndan:
> Ah gotcha, I agree it doesn't have to use separate scopes

If we decide to go with `horizontalAlignment`/`verticalAlignment`, we will have to introduce another scope to `LazyList`, so that the `*Alignment` properties switch the `*AxisAlignment` types (so `horizontalAlignment` will be `MainAxisAlignment` in `Row`, but `CrossAxisAlignment` in `Column`. I think you implied this in the rest of the paragraph, but I just wanted to make it clear.

> Compose UI uses align, however the major difference is the values passed to align imply horizontal or vertical alignment (e.g. Top, CenterVertically, Start, CenterHorizontally).

I like this less than having separate alignments tbh. I suppose my proposal solely aligns with CSS and not Compose UI.

The reason I like this less is that it isn't very accessible. Imo we should not only support LTR and RTL writing modes, but also top-to-bottom (like Japanese). Words like "top" have the same language centricity as "left" or "right" does. Granted, the words "row" and "column" also have this issue too.

> I disagree. The CSS names for horizontal and vertical alignment aren't very intuitive. It's impossible to know which direction Modifier.justifyContent changes without understanding the CSS spec. Modifier.horizontalAlignment is clear about what it does without searching.

I assumed that having a CSS-like API was our north-star, as if someone didn't understand what something means, they can Google it. I'm a fan of that idea, as it means people don't have to learn esoteric APIs, even if they're debatably easier to learn. Also, it's a lot easier for people to jump into this if they have CSS knowledge already, and if they don't, learning CSS things is applicable outside of Redwood.

Granted, I'm thinking about discussions we (the Treehouse team) had about year ago, so perhaps the zeitgeist has since changed.

> As mentioned, we will need two separate horizontalAlignment and verticalAlignment modifiers for Box so it's better to use horizontalAlignment and verticalAlignment for our other layout components to remain consistent.

Coming from CSS, a `Box` is achieved by just using flexbox, so you'd just use the align and justify properties, by first choosing a random direction you want to use (`Row` or `Column`). Again, debatably harder API, but less esoteric.

veyndan:
I think I'm debating a larger change than anticipated. Feel free to merge this, and I can update `LazyList` to take a `horizontalAlignment` and `verticalAlignment`.

JakeWharton:
> Coming from CSS, a `Box` is achieved by just using flexbox

Our box allows overlapping children whereas flexbox does not.

PR #1301, opened 2023-06-30
Upgrade Zipline to 1.1.0

swankjesse:
Now Zipline API changes need to be opted into with this:
  ./gradlew ziplineApiDump

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ jobs:
         timeout-minutes: 5
         continue-on-error: true
 
-      - run: ./gradlew kotlinUpgradeYarnLock build -PredwoodNoSamples
+      - run: ./gradlew kotlinUpgradeYarnLock build ziplineApiCheck -PredwoodNoSamples
 
   connected:
     runs-on: macos-latest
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
         timeout-minutes: 5
         continue-on-error: true
 
-      - run: ./gradlew kotlinUpgradeYarnLock build -PredwoodNoSamples
+      - run: ./gradlew kotlinUpgradeYarnLock build ziplineApiCheck -PredwoodNoSamples
```
JakeWharton: Isn't this already attached to the `check` lifecycle task? I wouldn't think we need to explicitly run it.

redwood-treehouse/api/zipline-api.toml
```diff
@@ -0,0 +1,67 @@
...
+  "moYx+T3e",
+
+  # fun sendChanges(kotlin.collections.List): kotlin.Unit
+  "W8qOuU0t",
+]
+
+[app.cash.redwood.treehouse.ZiplineTreehouseUi]
+
+functions = [
+  # fun close(): kotlin.Unit
```
JakeWharton, line 53: Feels like we should have omitted this one with a special case.

JakeWharton (approved):

.github/workflows/build.yaml
JakeWharton: Confirmed that it is, and removed

PR #1314, opened 2023-07-06
Test empty `LazyColumn` composable

veyndan:
This is to set up the groundwork for testing that the correct properties are emitted from `LazyList` as scrolling and data swapping occurs.

### DIFF
--- a/redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
+++ b/redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
@@ -1,0 +1,60 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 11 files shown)


redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
```diff
@@ -0,0 +1,147 @@
...
+      LazyColumn(placeholder = { Text("Placeholder") }) {
+      }
+    }
+
+    val snapshot = ExampleSchemaTester {
+      setContent(content)
+      awaitSnapshot()
+    }
+
+    val expected = listOf(
```
veyndan: This is a bit unwieldy

JakeWharton: Yeah I'm curious if this is the right layer at which to test this behavior, as opposed to the normal tester and asserting against values.

JakeWharton (approved):

redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
veyndan: I think the normal tester was what I wanted. The protocol level was definitely too low for this sorta thing.

Issue #1048, opened 2023-05-10
Add `LazyList` widget bindings for the HTML DOM


veyndan:
We will probably need to use the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) in order to implement `LazyList#onPositionDisplayed`.

JakeWharton:
Oooo fun

JakeWharton:
This should come with the removal of the `ColumnProvider` abstraction inside Emoji Search.

JakeWharton:
Maybe something to borrow/steal in https://gitlab.com/opensavvy/ui/compose-lazy-html

JakeWharton:
Don't forget to enable `publishing()` in the build file once ready!

PR #1321, opened 2023-07-11
Add truncating `LazyList` DOM implementation

veyndan:
The motivation for this was to get rid of `ColumnProvider` in Emoji Search. The `ColumnProvider` abstraction was making it difficult to add `LazyList` specific types as parameters.

Note that this doesn't properly implement `LazyList` on the web, as no items are loaded on scroll. Also, most of the widget properties haven't been implemented.


redwood-lazylayout-dom/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: 'org.jetbrains.kotlin.js'
+
+redwoodBuild {
+}
```
veyndan, line 4: Deliberately not `publishing()` this, as the current implementation doesn't properly implement lazy loading in the DOM. I wanted to make sure we don't accidentally publish a broken version of this. We should enable publishing once #1048 has been completed.

JakeWharton (approved):
Is this actually "truncating" as described? I didn't see where it happens.

veyndan:
It does, in the sense that an initial N number of items are sent over (as defined in `LazyList.kt`). We never load more, as the host never gives updates to the guest about the current view port, so we only ever display those initial N items.

JakeWharton:
Ah, makes sense.

Issue #1275, opened 2023-06-26
Triggering scrolls from presenter


swankjesse:
I’ve got some presenter code that would like to scroll its scrollable containers. I imagine APIs like this:
 * scrollToTop()
 * scrollToBottom()
 * scrollToItem(someIdentifier)

What’s annoying about these is they’re more like events than properties, but events that flow the opposite direction from our usual ones.

I could synthesize this with properties in a pinch.

JakeWharton:
Since everything is fundamentally rooted in Compose, there's not really a mechanism by which you can do this. The closest is how Compose UI does scrolling, which is by allowing you to adjust scroll on the scroll state object either directly or via animation.

PR #1322, opened 2023-07-12
Allow programmatic scrolling of `LazyList`

veyndan:
Closes https://github.com/cashapp/redwood/issues/1275


redwood-lazylayout-api/src/commonMain/kotlin/app/cash/redwood/lazylayout/api/properties.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * @param id Should only be used to trigger recompositions.
+ */
+@[Immutable Serializable]
+public class ScrollItemIndex(
+  @Suppress("unused") private val id: Int,
+  public val index: Int,
+)
```
veyndan: I dislike this. I'd rather just pass an `index`, but if `scrollToItem` is given the same `index`, then no recomposition occurs. We're passing a constantly incrementing `id` such that we always recompose when a new `index` is passed.

Any alternatives that would remove the need for an `id` would be greatly appreciated!

JakeWharton (approved):
Should need API generation as well to update the XML file.

PR #1326, opened 2023-07-18
Disable Android resources by default in libraries

JakeWharton:
Before:

    $ gw build --dry-run --console=plain | grep SKIPPED | wc -l
        6551

After:

    $ gw build --dry-run --console=plain | grep SKIPPED | wc -l
        6300


veyndan (approved):
Nice

PR #1328, opened 2023-07-20
In UIViewLazyList, calculate and cache the sizeThatFits value for each item on insertion

dnagler:
This value can then be read within the sizeForItemAtIndexPath hook. This change removes the workaround of hard-coding cell heights, and allows for compatibility with both vertical and horizontal scrolling.

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -64,9 +64,15 @@ internal interface ViewPortItems : Widget.Children<UIView> {
   var placeholders: MutableList<Widget<UIView>>
   var itemsBefore: Int
   var itemsAfter: Int
-  fun itemForGlobalIndex(index: Int): UIView
+  var collectionView: UICollectionView?
+  fun itemForGlobalIndex(index: Int): ViewPortItem
   fun itemCount(): Int
 }
+
+internal data class ViewPortItem(
+  val widget: Widget<UIView>,
+  val size: CValue<CGSize>,
+)
 
 internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {
 
@@ -75,11 +81,13 @@ internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {
 
     override var itemsBefore: Int = 0
     override var itemsAfter: Int = 0
+    override var collectionView: UICollectionView? = null
 
-    private val viewPortList = mutableListOf<Widget<UIView>>()
+    private val viewPortList = mutableListOf<ViewPortItem>()
 
     override fun insert(index: Int, widget: Widget<UIView>) {
-      viewPortList.add(index, widget)
+      val size = widget.value.sizeThatFits(containerSize())
+      viewPortList.add(index, ViewPortItem(widget, size))
     }
 
     override fun move(fromIndex: Int, toIndex: Int, count: Int) {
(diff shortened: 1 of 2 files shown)


apatronl (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -170,16 +181,12 @@ internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {
...
-//        }
-
-        // TODO: Handle size for horizontal scrollDirection
-        return CGSizeMake(collectionView.frame().useContents { size.width }, 64.0)
+        val itemSize = items.itemForGlobalIndex(sizeForItemAtIndexPath.item.toInt()).size
+        if (isVertical) {
+          return CGSizeMake(collectionView.frame().useContents { size.width }, itemSize.useContents { height })
+        } else {
+          return CGSizeMake(itemSize.useContents { width }, collectionView.frame().useContents { size.width })
+        }
```
JakeWharton: The vertical case uses the frame width and the item height but the horizontal case uses the item width and the frame width. Should the latter be item width and frame _height_?

Also, if you want, you can hoist the `return` outside the conditional.
```suggestion
        return if (isVertical) {
          CGSizeMake(collectionView.frame().useContents { size.width }, itemSize.useContents { height })
        } else {
          CGSizeMake(itemSize.useContents { width }, collectionView.frame().useContents { size.width })
        }
```

JakeWharton (approved):

veyndan (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: Whoops yeah, meant to use height

dnagler: Fixed!

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -144,6 +152,9 @@ internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {
         collectionView: UICollectionView,
         numberOfItemsInSection: NSInteger,
       ): NSInteger {
+        // NB: This is a bit of a hack, but we don't want to let the items list know
+        // the collectionView's frame until it's had a layout pass.
+        items.containerFrame = collectionView.frame()
```
underscoretang: This method gets called a lot, can we cache this value and only set it once, or better yet when the frame changes?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -218,6 +228,8 @@ internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {
         UICollectionViewScrollDirection.UICollectionViewScrollDirectionHorizontal
       }
     }
+
+    this.isVertical = isVertical
```
underscoretang: should we set this at the top of this method? not 100% about this, but if setting `collectionViewFlowLayout.scrollDirection` triggers a redraw we probably want this value to be set ahead of time. we need a horizontal list test/sample

underscoretang (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: Good point, I can move it up

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: Good point, I'll move this to the `width`/`height` hooks instead

dnagler: Better yet, I'll just cache a reference to the `collectionView` itself on load, and check its frame every time we insert a widget

PR #1330, opened 2023-07-21
Allow configuration of pull-to-refresh content color

veyndan:


redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
```diff
@@ -51,6 +51,7 @@ public data class RefreshableLazyList(
   @Property(9) val margin: Margin,
   @Property(10) val crossAxisAlignment: CrossAxisAlignment,
   @Property(11) val scrollItemIndex: ScrollItemIndex,
+  @Property(12) val pullRefreshContentColor: UInt,
```
JakeWharton, line 54: Where did "content" come from here? Some prior art?

I would think something like "indicator" would be more accurate in describing what this is changing.

JakeWharton (approved):

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
veyndan: I nicked "content" from Compose UI's [`PullToRefreshIndicator`](https://developer.android.com/reference/kotlin/androidx/compose/material/pullrefresh/package-summary#PullRefreshIndicator(kotlin.Boolean,androidx.compose.material.pullrefresh.PullRefreshState,androidx.compose.ui.Modifier,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,kotlin.Boolean)), where `contentColor` is the color of the spinning thing. I don't love the name, but I also don't particularly want to invent another new name for that icon's color.

Android View's uses `setColorSchemeColors` which isn't descriptive at all. iOS uses `tintColor` which is OK (though IMO it's — at best — on par with `contentColor`)?

PR #1331, opened 2023-07-21
UIViewFlexContainer ScrollView Support

underscoretang:
Making this a ScrollView eliminates the need to dynamically wrap the YogaView and deal with proxying sizing back and forth.

This works for a lot of complex nested use cases. For example:

https://github.com/cashapp/redwood/assets/465849/84cc4574-12f2-4e1b-a9f4-54bdcd3bdb7c

Remaining issues:
- [x] Wrapped sizing for Rows doesn't produce a height
- [x] Double Top Margins
- [x] Update width sizing when the data model changes (missing a layout pass for swapping shorter -> longer text)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,6 +14,8 @@ kotlin.native.ignoreIncorrectDependencies=true
 org.gradle.caching=true
 org.gradle.jvmargs=-Xmx4096m
 org.gradle.parallel=true
+# Specify a specific JAVA_HOME for gradle if needed
+# org.gradle.java.home=/Library/Java/JavaVirtualMachines/jdk-17.0.3.jdk/Contents/Home
 
 android.useAndroidX=true
 android.enableJetifier=false
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
@@ -25,6 +25,7 @@ import app.cash.redwood.layout.widget.Row
 import app.cash.redwood.ui.Default
 import app.cash.redwood.ui.Density
 import app.cash.redwood.ui.Margin
+import app.cash.redwood.widget.ChangeListener
 import app.cash.redwood.widget.UIViewChildren
 import app.cash.redwood.yoga.AlignItems
 import app.cash.redwood.yoga.FlexDirection
@@ -35,10 +36,9 @@ import platform.UIKit.UIView
 import platform.darwin.NSInteger
 
 internal class UIViewFlexContainer(
-  direction: FlexDirection,
-) : Row<UIView>, Column<UIView> {
+  private val direction: FlexDirection,
+) : Row<UIView>, Column<UIView>, ChangeListener {
   private val yogaView = YogaUIView()
-
   override val value: UIView get() = yogaView
   override val children = UIViewChildren(
     value,
@@ -64,12 +64,10 @@ internal class UIViewFlexContainer(
 
   override fun width(width: Constraint) {
     yogaView.width = width
-    invalidate()
   }
 
   override fun height(height: Constraint) {
     yogaView.height = height
-    invalidate()
   }
 
   override fun margin(margin: Margin) {
@@ -81,11 +79,10 @@ internal class UIViewFlexContainer(
         marginBottom = margin.bottom.toPx().toFloat()
       }
     }
-    invalidate()
   }
 
   override fun overflow(overflow: Overflow) {
-    invalidate()
+    yogaView.scrollEnabled = overflow == Overflow.Scroll
   }
 
   override fun horizontalAlignment(horizontalAlignment: MainAxisAlignment) {
@@ -106,15 +103,13 @@ internal class UIViewFlexContainer(
 
   private fun alignItems(alignItems: AlignItems) {
     yogaView.rootNode.alignItems = alignItems
-    invalidate()
   }
 
   private fun justifyContent(justifyContent: JustifyContent) {
     yogaView.rootNode.justifyContent = justifyContent
-    invalidate()
   }
 
-  private fun invalidate() {
+  override fun onEndChanges() {
     value.setNeedsLayout()
   }
 }
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
@@ -7,19 +7,26 @@
 package app.cash.redwood.layout.uiview
 
 import app.cash.redwood.layout.api.Constraint
+import app.cash.redwood.yoga.FlexDirection
+import app.cash.redwood.yoga.MeasureCallback
+import app.cash.redwood.yoga.MeasureMode
 import app.cash.redwood.yoga.Node
 import app.cash.redwood.yoga.Size
 import kotlinx.cinterop.CValue
+import kotlinx.cinterop.ObjCClass
 import kotlinx.cinterop.cValue
 import kotlinx.cinterop.useContents
 import platform.CoreGraphics.CGRectMake
 import platform.CoreGraphics.CGRectZero
 import platform.CoreGraphics.CGSize
 import platform.CoreGraphics.CGSizeMake
+import platform.UIKit.UILabel
+import platform.UIKit.UIScrollView
+import platform.UIKit.UIScrollViewContentInsetAdjustmentBehavior
 import platform.UIKit.UIView
 import platform.UIKit.UIViewNoIntrinsicMetric
 
-internal class YogaUIView : UIView(cValue { CGRectZero }) {
+internal class YogaUIView : UIScrollView(cValue { CGRectZero }) {
   val rootNode = Node()
 
   var width = Constraint.Wrap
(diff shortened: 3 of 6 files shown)


gradle.properties
```diff
@@ -14,6 +14,8 @@ kotlin.native.ignoreIncorrectDependencies=true
 org.gradle.caching=true
 org.gradle.jvmargs=-Xmx4096m
 org.gradle.parallel=true
+# Specify a specific JAVA_HOME for gradle if needed
```
underscoretang, line 17: I'm going to leave this as a comment for future reference. Let me know if there's a better alternative here

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -91,15 +90,18 @@ internal class UIViewFlexContainer(
...
     yogaView.rootNode.justifyContent = justifyContent
-    invalidate()
   }
 
   private fun invalidate() {
     value.setNeedsLayout()
   }
+
+  override fun onEndChanges() {
+    invalidate()
```
underscoretang: Invalidating after all changes are done can save us layout passes and doing intermediate calculations

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -7,42 +7,79 @@
...
 
   var width = Constraint.Wrap
   var height = Constraint.Wrap
 
-  var applyModifier: (Node, Int) -> Unit = { _, _ -> }
+  var applyModifier: (Node, Int) -> Unit = { _, _ -> println("applyModifier")}
+
+  init {
+    // TODO: Support Scroll Indicators
+    showsVerticalScrollIndicator = false
```
underscoretang, line 39: The issue with these atm is that they're part of the view hierarchy. when YogaUIView goes to `syncNodes`, it doesn't know how to differentiate or account for these scroll indicator views

samples/emoji-search/ios-uikit/EmojiSearchApp.xcodeproj/xcshareddata/xcschemes/EmojiSearchApp.xcscheme
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
underscoretang, line 1: Adding a scheme here to enable us to Profile the EmojiSample app in Instruments

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -91,15 +90,17 @@ internal class UIViewFlexContainer(
...
```
JakeWharton:
```suggestion
  override fun onEndChanges() {
    value.setNeedsLayout()
```

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -7,42 +7,79 @@
...
 
-internal class YogaUIView : UIView(cValue { CGRectZero }) {
+internal class YogaUIView : UIScrollView(cValue { CGRectZero }) {
   val rootNode = Node()
 
   var width = Constraint.Wrap
   var height = Constraint.Wrap
 
-  var applyModifier: (Node, Int) -> Unit = { _, _ -> }
+  var applyModifier: (Node, Int) -> Unit = { _, _ -> println("applyModifier")}
```
JakeWharton: Revert?

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -7,42 +7,79 @@
...
+    // calculate a size that the container should fit in.
+    val bounds = bounds.useContents {
+      sizeForConstraints(size)
+    }
 
+    if (scrollEnabled) {
+      // When scrolling is enabled, we want to calculate and apply the contentSize
+      // separately and have it grow a much as needed in the flexDirection.
+      // This duplicates the calculation we're doing above, and should be
+      // combined into one call
```
JakeWharton: Going to make a few of these suggestions, as [comments should be sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html).
```suggestion
      // When scrolling is enabled, we want to calculate and apply the contentSize
      // separately and have it grow a much as needed in the flexDirection.
      // This duplicates the calculation we're doing above, and should be
      // combined into one call.
```
I think this is pretty uncontroversial, but it's something that people may have never thought about so it can sometimes be weird to get started thinking about it.

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -7,42 +7,79 @@
...
+      val scrollSize = if (rootNode.flexDirection == FlexDirection.Column) {
+        Size(bounds.width, Size.Undefined)
+      } else {
+        Size(Size.Undefined, bounds.height)
+      }
+      val contentSize = calculateLayoutWithSize(scrollSize)
+      setContentSize(contentSize.toCGSize())
+      calculateLayoutWithSize(bounds)
+    } else {
+      // If we're not scrolling, the contentSize should equal the size of the view
```
JakeWharton:
```suggestion
      // If we're not scrolling, the contentSize should equal the size of the view.
```

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -7,42 +7,79 @@
...
+      val containerSize = calculateLayoutWithSize(bounds)
+      setContentSize(containerSize.toCGSize())
+    }
+
+    // Layout the nodes based on the calculatedLayouts above
```
JakeWharton:
```suggestion
    // Layout the nodes based on the calculatedLayouts above.
```

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt
```diff
@@ -67,7 +67,10 @@ public class UIViewChildren(
   }
 
   private fun invalidate() {
-    parent.setNeedsLayout()
-    parent.invalidateIntrinsicContentSize()
+    parent.superview?.let {
+      it.setNeedsLayout()
+    } ?: run {
+      parent.setNeedsLayout()
+    }
```
JakeWharton: Dying for `if let` rather than this (very common) Kotlin pattern.

How about
```suggestion
    (parent.superview ?: parent).setNeedsLayout()
```
which hopefully is a bit more simple to parse.

The rules around iOS layout still elude me...

samples/emoji-search/presenter/src/commonMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearch.kt
```diff
@@ -60,11 +62,25 @@ interface Navigator {
...
   httpClient: HttpClient,
   navigator: Navigator,
+) {
+  // Different variants of EmojiSearch:
+
+  // Flex Containers
+  NestedFlexBoxContainers(httpClient, navigator)
+
+  // Lazy Column
+  // LazyColumn(httpClient, navigator)
```
JakeWharton: I'm a bit wary of making this sample too complicated, as it's our only real Treehouse one. For now we can leave this as-is, but it's probably wise to add a playground-like app in the future were we can capture more esoteric experiments.

JakeWharton: Also we plan to have automated UI tests against this app to validate behavior as an integration test for each platform (#537), which is another reason we shouldn't add a ton of features to it.

JakeWharton (approved):

samples/emoji-search/presenter/src/commonMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearch.kt
underscoretang: Is the EmojiSample mean to only develop/validate/test the LazyColumn? I think we should start adding more complex examples somewhere in the redwood sample code since that's what we'll end up needing downstream from redwood and running into redwood issues at that stage really bogs down the iteration time.

Since we're using EmojiSample as the main development environment in redwood, it should stress test more complex (real world) cases to get ahead of regressions and not require feedback signals from treehouse guest app development

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/UIViewChildren.kt
underscoretang: me too. this is a hack that I'm hoping to find another way around

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -25,36 +25,36 @@ import app.cash.redwood.layout.widget.Row
...
-
   override val value: UIView get() = yogaView
   override val children = UIViewChildren(value)
   override var modifier: Modifier = Modifier
 
   init {
     yogaView.rootNode.flexDirection = direction
     yogaView.applyModifier = { node, index ->
-      node.applyModifier(children.widgets[index].modifier, Density.Default)
+      if (index < children.widgets.count()) {
```
dnagler: Just curious--why is this change needed? How often is this called with `index >= children.widgets.count()`?

samples/emoji-search/ios-uikit/EmojiSearchApp.xcodeproj/xcshareddata/xcschemes/EmojiSearchApp.xcscheme
dnagler: Nice, thanks

samples/emoji-search/presenter/src/commonMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearch.kt
JakeWharton: It's the only sample that demonstrates Treehouse usage. The counter sample demonstrates raw Redwood.

This is why I would prefer both be kept as simple as possible. Each is there to demonstrate the structure of the technology (which is already complex) and so a complex application can distract or mask what's going on with the setup.

So I would prefer if long-term we either created a playground-style sample in which we can put a bunch of screens with more targeted demonstrations of particular components, or we built a separate sample which is just a more complex app. Or we could do both!

The repo search sample is one which is solely meant to demo LazyColumn usage, I believe. It would fit well within a playground app, or we could expand it to be an exceedingly basic GitHub browser app as a more complex example.

We don't have to do anything in this PR right now, though.

dnagler (approved):

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
underscoretang: i'll pull this out, this was an issue when I nested a LazyList in here and hard scrolled, but after rebasing, it's not crashing anymore

samples/emoji-search/presenter/src/commonMain/kotlin/com/example/redwood/emojisearch/presenter/EmojiSearch.kt
underscoretang: Got it. I'll merge this for now and we can spin up a playground sample (or one dedicated to flexbox).

Let's add a README to https://github.com/cashapp/redwood/tree/trunk/samples to make it clear what each sample is intended to demonstrate

PR #1335, opened 2023-07-24
Extract common windowing logic from Android View's and UIKit

veyndan:


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -99,10 +93,6 @@ internal class ViewPortItems : Widget.Children<UIView> {
     return ViewPortItem(placeholders[placeholderIndex], containerSize())
   }
 
-  internal fun itemCount(): Int {
-    return max(itemsBefore - 1, 0) + viewPortList.count() + itemsAfter
```
veyndan, line 104: Taking 1 off of `itemsBefore` is a bug. Now that our `itemCount()` calculation is consolidated in `WindowedItemList`, such a discrepancy between Android an iOS shouldn't occur.

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -209,7 +205,7 @@ internal open class ViewLazyList(context: Context) : LazyList<View> {
     val delta = itemsAfter - items.itemsAfter
     items.itemsAfter = itemsAfter
 
-    val positionStart = items.itemsBefore + items.widgets.size
+    val positionStart = items.itemsBefore + items.items.size
```
veyndan, line 208: `items.items` is unfortunate. I'll rename it in a follow up.

JakeWharton (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
apatronl: Ahh I think this might fix an issue I'm seeing on iOS activity feed

PR #1337, opened 2023-07-24
Reduce responsibilities of `ViewPortItems`

veyndan:
This is to prepare merging `UIViewLazyList.ViewPortItems` with `ViewLazyList.Items`.

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -64,14 +64,11 @@ import platform.UIKit.item
 import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
-internal class ViewPortItems(
-  private val collectionView: UICollectionView,
-) : Widget.Children<UIView>, WindowedItemList<ViewPortItem> by WindowedItemListImpl() {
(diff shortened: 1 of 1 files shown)


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -83,30 +80,30 @@ internal class ViewPortItems(
...
 
-internal data class ViewPortItem(
+private data class ViewPortItem(
   val widget: Widget<UIView>,
   val size: CValue<CGSize>,
 )
 
 internal open class UIViewLazyList : LazyList<UIView>, ChangeListener {
 
+  // Fetch the item relative to the entire collection view
```
JakeWharton, line 96: I don't want to break your stack, but in the future: [comments should be sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html)

JakeWharton (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
veyndan: 100% agreed. In this case though, I just relocated the comment instead of writing it myself

PR #1339, opened 2023-07-25
Remove redundant conditional

veyndan:
Usually, this would be a bug. In this case though, the value of `isVertical` is baked into the "type"/name of the composable function (i.e., `LazyRow` or `LazyColumn`), thus the value of `isVertical` is only ever set once. The default value of `UICollectionViewFlowLayout` is vertical, so we previously just had an unused code path, which has now been removed.


JakeWharton (approved):

dnagler:
belated LGTM

PR #1341, opened 2023-07-25
Make `UIViewLazyList` placeholders self-sizing

veyndan:
Before this change, each placeholder is the height of the viewport. After this change, each placeholder is the height as defined by itself. This lines up with what `ViewLazyList` already does.

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -82,7 +82,8 @@ internal open class UIViewLazyList : LazyList<UIView>, ChangeListener {
 
     // If we don't have a value, fallback to our pools of placeholders
     val placeholderIndex = index % placeholder.size
-    return ViewPortItem(placeholder[placeholderIndex], collectionView.size())
+    val widget = placeholder[placeholderIndex]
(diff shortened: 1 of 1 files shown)


JakeWharton (approved):
We gotta get some screenshot tests for UIKit rolling soon...

veyndan:
Are we still just waiting on https://github.com/cashapp/redwood/pull/1218#issuecomment-1589650570?

JakeWharton:
It was determined to be not viable, or at least require a ridiculous amount of effort. We can still write them in Kotlin, but we need a Swift entrypoint and to run it via xcrun or whatever.

veyndan:
Gotcha — I just created https://github.com/cashapp/redwood/issues/1343 to track this

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -82,7 +82,8 @@ internal open class UIViewLazyList : LazyList<UIView>, ChangeListener {
 
     // If we don't have a value, fallback to our pools of placeholders
     val placeholderIndex = index % placeholder.size
-    return ViewPortItem(placeholder[placeholderIndex], collectionView.size())
+    val widget = placeholder[placeholderIndex]
+    return ViewPortItem(widget, widget.value.sizeThatFits(collectionView.size()))
```
dnagler, line 86: good catch!

PR #1342, opened 2023-07-25
Remove `ViewPortItem` class

veyndan:
`itemForGlobalIndex` returns a `ViewPortItem`. The function is only called twice, with one of the callers discarding the size calculation found in `ViewPortItem`. As the size is just a function of the widget, by inlining the size calculation, we can skip unnecessary sizing in this case.

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -63,27 +63,12 @@ import platform.UIKit.item
 import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
-private fun UICollectionView.size(): CValue<CGSize> {
-  return frame().useContents { size.readValue() }
-}
-
-private data class ViewPortItem(
-  val widget: Widget<UIView>,
-  val size: CValue<CGSize>,
-)
-
 internal open class UIViewLazyList : LazyList<UIView>, ChangeListener {
 
   // Fetch the item relative to the entire collection view
-  private fun WindowedChildren<UIView>.itemForGlobalIndex(index: Int): ViewPortItem {
-    items.getOrNull(max(index - itemsBefore, 0))?.let {
-      return ViewPortItem(it, it.value.sizeThatFits(collectionView.size()))
-    }
-
-    // If we don't have a value, fallback to our pools of placeholders
-    val placeholderIndex = index % placeholder.size
-    val widget = placeholder[placeholderIndex]
-    return ViewPortItem(widget, widget.value.sizeThatFits(collectionView.size()))
+  private fun WindowedChildren<UIView>.itemForGlobalIndex(index: Int): Widget<UIView> {
+    return items.getOrNull(max(index - itemsBefore, 0))
+      ?: placeholder[index % placeholder.size]
   }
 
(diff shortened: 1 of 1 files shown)


JakeWharton (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -121,7 +106,8 @@ internal open class UIViewLazyList : LazyList<UIView>, ChangeListener {
         layout: UICollectionViewLayout,
         sizeForItemAtIndexPath: NSIndexPath,
       ): CValue<CGSize> {
-        val itemSize = items.itemForGlobalIndex(sizeForItemAtIndexPath.item.toInt()).size
+        val widget = items.itemForGlobalIndex(sizeForItemAtIndexPath.item.toInt())
+        val itemSize = widget.value.sizeThatFits(collectionView.frame().useContents { size.readValue() })
```
dnagler, line 110: Is this performant? The whole reason I added the `ViewPortItem` class in PR #1328 was so that we could pre-compute the `sizeThatFits` on `insert` (called once per widget), rather than here (called repeatedly on scroll).

dnagler: Testing in the EmojiSearch sample app on the latest `trunk`, this definitely feels like a regression--scrolling is no longer performant, and most of the screen is occupied by placeholder loading widgets for long periods of time. I tried reverting the changes from this PR locally, but I'm still seeing regressed performance.

I know that it's still early days for Redwood in general, but given that this class is the core component of the Activity feed, if we were to create a new internal release with the current `trunk`, local development on the Activity team would suffer. I think we'll need to have a system in place going forward to indicate whether or not `trunk` is "green" in terms of being ready for the next internal release, or mid-refactor--ideally, `trunk` will always be "green".

veyndan: I think we should assume a slight degradation in performance because of #1341, where we now have more widgets to layout and size. This wouldn't account for such a large jump in performance though. Running EmojiSearch and removing the usage of `Row` in each item shows _significant_ performance improvements. Currently looking into that.

Regarding your second paragraph, I 100% agree. I've gotten used to Redwood being in development, that a "one step backwards, two steps forward" approach on `trunk` is no longer permissible, due to the broader impact that Redwood has internally. Will definitely keep that in mind in the future!

For the current breakage, I'll continue to try to fix it. If it takes longer, or we need an internal release, I can make the necessary rollbacks, and continue development on a separate branch.

veyndan: I think this is the underlying issue (it is to do with `Row`/`Column`): https://github.com/cashapp/redwood/issues/1205

If you comment out the following line, performance drastically improves. https://github.com/cashapp/redwood/blob/3f256fa271ea6d8913838cad019f2e0bf6cae212/redwood-yoga/src/commonMain/kotlin/app/cash/redwood/yoga/Node.kt#L98

dnagler: Thanks for continuing to dig into this!

PR #1344, opened 2023-07-25
Share `itemsBefore`, `itemsAfter`, and viewport updating logic between View and UIKit

veyndan:


redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
```diff
@@ -0,0 +1,62 @@
...
+  override fun itemsBefore(itemsBefore: Int) {
+    val delta = itemsBefore - items.itemsBefore
+    items.itemsBefore = itemsBefore
+
+    if (delta > 0) {
+      listUpdateCallback.onInserted(itemsBefore - delta, delta)
+    } else {
+      listUpdateCallback.onRemoved(itemsBefore, -delta)
+    }
+  }
```
veyndan, line 49: This logic is from https://github.com/cashapp/redwood/pull/1127/files. With Android overriding this function, and iOS having a no-op `ListUpdateCallback` implementation, this code is essentially unused. When iOS has an actual `ListUpdateCallback` implementation in a follow-up, we'll actually start using the code here.

JakeWharton (approved):

PR #1351, opened 2023-07-26
Insert and remove nodes with child modifications

veyndan:
Node insertion and deletion is solely managed by `UIViewChildren`. Instead of checking if the node children and the view children are in sync on every measure call (and if not, having to recreate all the node children), we can instead just update the node children as we update the view children, so we're certain that they are in sync, removing the need for the check.

### DIFF
--- a/redwood-layout-uiview/build.gradle
+++ b/redwood-layout-uiview/build.gradle
@@ -1,3 +1,4 @@
+import app.cash.redwood.buildsupport.ComposeHelpers
 import app.cash.redwood.buildsupport.FlexboxHelpers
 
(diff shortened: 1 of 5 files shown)


JakeWharton (approved):
May want @colinrtwhite to weigh in here just in case. The View-based flex layout also does this sync node pattern but I don't recall the reasoning.

colinrtwhite (approved):
LGTM - we should add these callbacks to `ViewChildren` as well. I initially went with the diffing approach to avoid duplicating the `typedSubviews` insertion/removal code.

PR #1238, opened 2023-06-16
Store and restore state for widget view in compose (syntax)

jingwei99:
This PR introduces a mechanism to store and restore state for the Android view implementation as part of `onSaveInstanceState` and `onRestoreInstanceState`.

**Android view implementation: works as designed**

https://github.com/cashapp/redwood/assets/57155915/1b79da53-4626-4569-a50d-8df6d6324e90

**Android compose implementation: it's not the focus of this PR, and it's somewhat broken right now, will look into it**

https://github.com/cashapp/redwood/assets/57155915/ec739245-6ad2-4ffc-ab0b-2471860b68ec

**iOS implementation: it just works ... (no work is required)**

PR #1357, opened 2023-07-27
Convert restoredId to a value class

jingwei99:
follow up from https://github.com/cashapp/redwood/pull/1238/files#r1234088328

### DIFF
--- a/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
+++ b/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
@@ -17,6 +17,7 @@ package app.cash.redwood.treehouse
 
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.mutableStateOf
+import kotlin.jvm.JvmInline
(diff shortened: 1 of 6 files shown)


redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
```diff
@@ -33,6 +34,10 @@ public class StateSnapshot(
       entry.value.map { mutableStateOf(it.fromJsonElement()) }
     }
   }
+
+  @JvmInline
+  @Serializable
+  public value class Id(public val value: String?)
```
colinrtwhite, line 40: TIL `value` classes can be nested; I always thought they had to be top level.

colinrtwhite (approved):

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
JakeWharton: That was indeed a limitation a few years ago but it has since been lifted.

PR #1358, opened 2023-07-27
Add guard on canBeSaved in SaveableStateRegistry call

jingwei99:
Follow up from https://github.com/cashapp/redwood/pull/1238/files#r1234089322

### DIFF
--- a/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
+++ b/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
@@ -69,7 +69,7 @@ private class RedwoodZiplineTreehouseUi(
 
     this.saveableStateRegistry = SaveableStateRegistry(
       restoredValues = stateSnapshot?.toValuesMap(),
-      canBeSaved = { stateSnapshot?.canBeSaved() ?: false },
+      canBeSaved = { stateSnapshot?.canBeSaved() ?: true },
     )
 
     composition.bind(
--- a/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
+++ b/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
@@ -69,7 +69,7 @@ private class RedwoodZiplineTreehouseUi(
 
     this.saveableStateRegistry = SaveableStateRegistry(
       restoredValues = stateSnapshot?.toValuesMap(),
-      canBeSaved = { true },
+      canBeSaved = { stateSnapshot?.canBeSaved() ?: false },
     )
 
     composition.bind(
--- a/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
+++ b/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
@@ -40,7 +40,7 @@ public class StateSnapshot(
       // Test if the content is serializable
       toValuesMap()
       true
-    } catch (t: Throwable) {
+    } catch (t: IllegalStateException) {
       false
     }
   }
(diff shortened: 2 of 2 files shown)


redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
```diff
@@ -69,7 +69,7 @@ private class RedwoodZiplineTreehouseUi(
 
     this.saveableStateRegistry = SaveableStateRegistry(
       restoredValues = stateSnapshot?.toValuesMap(),
-      canBeSaved = { true },
+      canBeSaved = { stateSnapshot?.canBeSaved() ?: true },
     )
```
veyndan, line 73: I'm not sure how `canBeSaved` can ever be `false`. We evaluate `stateSnapshot?.toValuesMap()` in `restoredValues` before the `canBeSaved` lambda is invoked. If this `toValuesMap()` throws an exception, we're not catching it anywhere.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
```diff
@@ -35,6 +35,16 @@ public class StateSnapshot(
     }
   }
 
+  public fun canBeSaved(): Boolean {
+    return try {
+      // Test if the content is serializable
+      toValuesMap()
+      true
+    } catch (t: Throwable) {
```
veyndan: Is there a more specific exception we can catch (like `SerializationException`)?

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
jingwei99: - `canBeSaved` is invoked during instantiation of `SaveableStateRegistryImpl`

- `canBeSaved` is a mechanism to signal `SaveableStateRegistryImpl` if `restoredValues` is serializable, it's in this case, the `canBeSaved` happen to come with the underlying implementation that utilizes `stateSnapshot?.toValuesMap()`

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
jingwei99: for now `IllegalStateException`: https://github.com/cashapp/redwood/blob/780941e26fa1f1037f6e212d79ffe9efaa26ad87/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt#L94 I'll tight it up to `IllegalStateException` here

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
veyndan: Maybe I'm misunderstanding something. Here are my assumptions:

* Calling `stateSnapshot?.toValuesMap()` will throw an exception if the values aren't serializable (I'm assuming this by the implementation of your `canBeSaved` function).
* When instantiating the `SaveableStateRegistry` instance, we're immediately invoking `stateSnapshot?.toValuesMap()` (on line 71).
* If that fails because `toValuesMap()` isn't serializable, then an exception is thrown.
* We never got to a point where the `canBeSaved` lambda is invoked, because we threw an exception while instantiating `SaveableStateRegistry`.

As an aside, would it be possible to add a unit test for this?

jingwei99: yep, that's the right assumption. 

The point of this change is to serve as a reminder that `canBeSaved` needs to receive a fair share of attention.

and it's in this case, the `canBeSaved` happen to come with the underlying implementation that utilizes `stateSnapshot?.toValuesMap()`, this will not always be the case, thus the reminder.

unit test could make sense for `fromJsonElement` I think?

veyndan (approved):

PR #1366, opened 2023-07-28
Convert public API data classes to Poko

### LINKED ISSUE #1043: Eliminate `data class` from public API

drewhamilton:
Closes #1043. Depends on #1369.

Figured I'd get this up with the snapshot to confirm it's all working as expected before cutting Poko 0.15. Unfortunately it's blocked by Kotlin 1.9, unless we restore the reflection removed in drewhamilton/poko#196.

Took my best guess on which modules are public API, lmk if I missed anything!


build.gradle
```diff
@@ -15,13 +15,15 @@ buildscript {
...
     classpath libs.paparazzi.gradlePlugin
+    classpath libs.poko.gradlePlugin
     classpath 'app.cash.redwood.build:gradle-plugin'
     classpath 'app.cash.redwood:redwood-gradle-plugin'
   }
   repositories {
     mavenCentral()
     google()
     gradlePluginPortal()
+    maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots") }
```
drewhamilton, line 26: TODO: Remove

gradle/libs.versions.toml
```diff
@@ -74,5 +74,6 @@ coil-core = { module = "io.coil-kt:coil", version.ref = "coil" }
 turbine = "app.cash.turbine:turbine:1.0.0"
 ktlint = "com.pinterest:ktlint:0.50.0"
 googleJavaFormat = "com.google.googlejavaformat:google-java-format:1.17.0"
+poko-gradlePlugin = "dev.drewhamilton.poko:poko-gradle-plugin:0.15.0-SNAPSHOT"
```
drewhamilton, line 77: TODO: Bump to release version

redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
```diff
@@ -48,33 +52,33 @@ public sealed interface Change {
...
+  public val value: JsonElement = JsonNull,
 ) : ValueChange
 
 @Serializable
 @SerialName("modifier")
-public data class ModifierChange(
+@Poko public class ModifierChange(
   override val id: Id,
-  val elements: List<ModifierElement> = emptyList(),
+  public val elements: List<ModifierElement> = emptyList(),
```
drewhamilton: TODO: Array

settings.gradle
```diff
@@ -17,6 +17,7 @@ dependencyResolutionManagement {
   repositories {
     mavenCentral()
     google()
+    maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots") }
```
drewhamilton, line 20: TODO: Remove

drewhamilton:
Most of CI fails with the reflection issue mentioned above. But `build` fails with this:

> Task :redwood-runtime:compileCommonMainKotlinMetadata FAILED
e: java.lang.IllegalStateException: e: Could not find "/.../redwood/redwood-runtime/build/kotlinTransformedMetadataLibraries/commonMain/dev.drewhamilton.poko-poko-annotations-0.15.0-SNAPSHOT-commonMain-oWB5zQ.klib" in [/.../Library/Application Support/kotlin/daemon]

Not sure if this is a configuration problem in this PR or something Poko is setting up wrong.

JakeWharton:
If you rebase on #1369 the build gets much farther.

The problem seems to be that Poko's Gradle plugin has a `compileOnly` dependency on KGP 1.9 so it doesn't cause Gradle to upgrade here. The runtime, however, is built with Kotlin 1.9 metadata which fails to be resolved correctly with 1.8.22 for some reason.

drewhamilton:
Ah OK, I'll rebase on the 1.9.0 PR. My thought behind using `compileOnly` was that a project should explicitly opt-in to upgraded Kotlin compiler versions, rather than be silently upgraded by Poko.

JakeWharton:
Yeah it's a reasonable choice. Both options have upsides and downsides.

drewhamilton:
Apparently KotlinX serialization omits empty lists when encoding, but includes empty arrays. Some of the tests fail as a result, and I'm not sure if they're safe to change (adding the empty array) or if more stuff needs to change along with them.

Examples below:
```
app.cash.redwood.protocol.ProtocolTest[jvm] > eventEmptyArgs[jvm] FAILED
    org.opentest4j.AssertionFailedError: expected:<"{"id":1,"tag":2[]}"> but was:<"{"id":1,"tag":2[,"args":[]]}">

app.cash.redwood.protocol.SnapshotChangeListTest[jvm] > validChangesSerialization[jvm] FAILED
    org.opentest4j.AssertionFailedError: expected:<...e":"modifier","id":1[]},{"type":"property"...> but was:<...e":"modifier","id":1[,"elements":[]]},{"type":"property"...>
```

The failure messages are janked up because assertk is mixing up its own display format with the JSON format, but the actual difference is the presence of e.g. `,"args":[]`.

JakeWharton:
I suspect it's because it checks `.equals` against the default value which works for lists and fails for arrays.

I can write a custom serializer for those properties that omits it when empty.

drewhamilton:
I pushed a serializer for `Event`, but it's very verbose. I couldn't find a way to supply a custom serializer for a single property, but I'm hoping there's a way.

If not, another option could be a constant `(internal?) val EmptyJsonElementArray` and stop using `emptyArray` and `arrayOf()` in generated code and tests.

drewhamilton:
Trying out `SerializersModuleBuilder.contextual` but it's not getting picked up the way I hoped it would.

JakeWharton:
Weird that it didn't retarget. I guess because it's from a fork?

JakeWharton:
Back at a computer. It won't even let me re-open. Want to rebase and make a new PR when you get a chance?

PR #1367, opened 2023-07-28
Invalidate `ViewFlexContainer` after all property changes applied

veyndan:
Aligns with what `UIViewFlexContainer` does.


redwood-layout-shared-test/src/main/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -143,6 +144,7 @@ abstract class AbstractFlexContainerTest<T : Any> {
     }
     verifySnapshot(container, "Center")
     container.alignItems(AlignItems.FlexEnd)
+    container.onEndChanges()
```
veyndan, line 147: I don't like this. I'll refactor `TestFlexContainer` such that modifications will automatically call `onEndChanges`. On first thought, it'll probably be something like:

```kotlin
container.modify { // this: ModifyScope
    alignItems(AlignItems.FlexEnd) // Only available in `ModifyScope`
    // Optional other changes
    // When lambda is completed, call `onEndChanges()`
}
```

colinrtwhite (approved):

PR #1368, opened 2023-07-28
Support custom version or coordinates for Compose compiler

### LINKED ISSUE #1364: Provide the ability to set the JB Compose compiler version in our consumer plugin
This _should_ remove the need to make a release for every Kotlin version, as consumers can bump the version locally as they can with AndroidX Compose compiler through the AGP DSL and JetBrains Compose compiler through JetBrains Compose Gradle plugin DSL.

Unlike AGP and CGP[^1], we should provide the ability to accept the full artifact coordinate AND accept a version catalog reference.

[^1]: No one calls it this…

JakeWharton:
Since this project does not build a Kotlin compiler plugin, we should not be forced to do a release for new versions of Kotlin. This property will allow consumers to upgrade thier Kotlin version by specifying a newer JetBrains Compose compiler version, or a totally different set of Maven coordiantes for a custom Compose compiler aritfact.

Closes #1364

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,7 +125,7 @@ allprojects {
       }
     }
     android.composeOptions {
-      kotlinCompilerExtensionVersion libs.versions.androidxComposeCompiler.get()
+      kotlinCompilerExtensionVersion libs.androidx.compose.compiler.get().version
     }
   }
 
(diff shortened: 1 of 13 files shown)


veyndan (approved):
Nice!

PR #1371, opened 2023-07-28
Test widget property contract instead of the yoga contract

veyndan:
We're testing at a slightly lower level than is necessary. We should be testing what the widget's are sending over (i.e., `MainAxisAlignment` and `CrossAxisAlignment`) instead of what our abstraction over Yoga expects (i.e., `Alignitems` and `JustifyContent`).


redwood-layout-shared-test/src/main/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -114,48 +112,47 @@ abstract class AbstractFlexContainerTest<T : Any> {
...
   }
 
-  @Test fun layoutWithAlignItems(
+  @Test fun layoutWithCrossAxisAlignment(
     @TestParameter flexDirectionEnum: FlexDirectionEnum,
-    @TestParameter alignItemsEnum: AlignItemsEnum,
+    @TestParameter crossAxisAlignmentEnum: CrossAxisAlignmentEnum,
   ) {
     assumeTrue(flexDirectionEnum in listOf(FlexDirectionEnum.Row, FlexDirectionEnum.Column))
-    assumeFalse(alignItemsEnum == AlignItemsEnum.Baseline)
```
veyndan, line 123: As Yoga's `AlignItems` defines `Baseline`, and `CrossAxisAlignment` doesn't, we no longer have to explicitly opt-out of testing `Baseline`.

redwood-layout-shared-test/src/main/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -280,36 +277,34 @@ private fun ConstraintEnum.toConstraint() = when (this) {
...
-  JustifyContentEnum.SpaceAround -> JustifyContent.SpaceAround
-  JustifyContentEnum.SpaceEvenly -> JustifyContent.SpaceEvenly
+private fun MainAxisAlignmentEnum.toMainAxisAlignment() = when (this) {
+  MainAxisAlignmentEnum.Start -> MainAxisAlignment.Start
+  MainAxisAlignmentEnum.Center -> MainAxisAlignment.Center
+  MainAxisAlignmentEnum.End -> MainAxisAlignment.End
+  MainAxisAlignmentEnum.SpaceBetween -> MainAxisAlignment.SpaceBetween
+  MainAxisAlignmentEnum.SpaceAround -> MainAxisAlignment.SpaceAround
+  MainAxisAlignmentEnum.SpaceEvenly -> MainAxisAlignment.SpaceEvenly
 }
```
JakeWharton, line 310: Probably best to put the value inside the enum, like
```kotlin
enum class MainAxisAlignmentEnum(val value: MainAxisAlignment) {
  Start(MainAxisAlignment.Start)
  ...
}
```
to eliminate the need to do a mapping function.

JakeWharton (approved):

redwood-layout-shared-test/src/main/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
JakeWharton: https://github.com/cashapp/redwood/pull/1377

PR #1378, opened 2023-07-28
Extract out common Yoga `FlexContainer` logic

veyndan:


redwood-layout-widget-yoga/src/commonMain/kotlin/app/cash/redwood/layout/widget/yoga/YogaFlexContainer.kt
```diff
@@ -0,0 +1,47 @@
...
+package app.cash.redwood.layout.widget.yoga
+
+import app.cash.redwood.layout.api.CrossAxisAlignment
+import app.cash.redwood.layout.api.MainAxisAlignment
+import app.cash.redwood.layout.widget.FlexContainer
+import app.cash.redwood.ui.Density
+import app.cash.redwood.ui.Margin
+import app.cash.redwood.yoga.Node
+
+interface YogaFlexContainer<W : Any> : FlexContainer<W> {
```
JakeWharton: This is a pretty small module to carry and to put into public API. Could we accomplish this with a new `copyPasta.kt` entry? Or maybe it could be added to the existing `FlexboxHelpers`?

veyndan: I think I was imagining third party implementations that could depend on this module. I've definitely back-tracked on that thinking since this PR, so I've just chucked it in `FlexboxHelpers` as you suggested.

JakeWharton (approved):

PR #1381, opened 2023-08-01
Add comment to canBeSaved instantiation of SaveableStateRegistry 

jingwei99:


redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
```diff
@@ -69,6 +69,8 @@ private class RedwoodZiplineTreehouseUi(
 
     this.saveableStateRegistry = SaveableStateRegistry(
       restoredValues = stateSnapshot?.toValuesMap(),
+      // about `canBeSaved`: values will only be restored by SaveableStateRegistry if `canBeSaved` is true
+      // `canBeSaved` essentially tests if values saved (and to be restored) are serializable
```
veyndan: Can we make this more of a `TODO`? This comment explains what `canBeSaved` does, instead of explaining that it currently doesn't work.

jingwei99: what does a TODO look like here? The way I see it (could be different from what you see it): there's no follow up action needed from this comment. Could you share what you see here?

veyndan (approved):
LGTM!

PR #1386, opened 2023-08-02
Reduce conversions between `Size` and `CGSize`

veyndan:
We are instantiating a `Size` object in many places, only to convert it to a `CGSize`. This indirection is unnecessary, and can be expensive considering `YogaUIView` function invocations occur on the UI thread.

### DIFF
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
@@ -8,19 +8,15 @@ package app.cash.redwood.layout.uiview
 
 import app.cash.redwood.layout.api.Constraint
 import app.cash.redwood.yoga.FlexDirection
-import app.cash.redwood.yoga.MeasureCallback
-import app.cash.redwood.yoga.MeasureMode
 import app.cash.redwood.yoga.Node
 import app.cash.redwood.yoga.Size
 import kotlinx.cinterop.CValue
-import kotlinx.cinterop.ObjCClass
 import kotlinx.cinterop.cValue
 import kotlinx.cinterop.useContents
 import platform.CoreGraphics.CGRectMake
 import platform.CoreGraphics.CGRectZero
 import platform.CoreGraphics.CGSize
 import platform.CoreGraphics.CGSizeMake
-import platform.UIKit.UILabel
 import platform.UIKit.UIScrollView
 import platform.UIKit.UIScrollViewContentInsetAdjustmentBehavior
 import platform.UIKit.UIView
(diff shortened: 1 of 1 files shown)


redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -42,13 +38,12 @@ internal class YogaUIView : UIScrollView(cValue { CGRectZero }) {
   }
 
   override fun intrinsicContentSize(): CValue<CGSize> {
-    return calculateLayoutWithSize(Size(Size.Undefined, Size.Undefined)).toCGSize()
+    return calculateLayoutWithSize(CGSizeMake(Size.Undefined.toDouble(), Size.Undefined.toDouble()))
```
veyndan, line 41: `Size.Undefined.toDouble()` occurs in many places, which could be pulled out into a constant to prevent the `Float` to `Double` conversion happening in many places. I'm assuming that the compiler is smart enough to do such an optimisation though.

dnagler: Why not just use `CGSizeMake(0.0, 0.0)` here? It's standard practice to use `CGRect.zero` on iOS to signal "we don't know the frame right now, but we'll calculate it later", and it feels like we're doing something similar here.

dnagler (approved):

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
veyndan: `Size.Undefined` references a Yoga specific constant. As this eventually gets passed to `Node#measure` (some Yoga code), presumably the fact that this constant equals `Float.NaN` is important?

colinrtwhite: @veyndan Does doing this conversion open us up to any floating point number issues? My worry was `Size.Undefined == Size.Undefined.toDouble().toFloat()` might not be true.

colinrtwhite: @veyndan @dnagler Yep, passing 0 to Yoga code, would measure its children as if it had a 0 width/height. `Size.Undefined` tells the Yoga code that container has an unbounded width/height.

JakeWharton: NaNs do survive both conversions since they're the same concept regardless of how many bits are dedicated to the floating point representation.

PR #1388, opened 2023-08-03
Only measure `UIViewLazyList` rows when in viewport

veyndan:
### Before

On load, `UICollectionViewFlowLayout` will ask **EVERY** row for its size, regardless of whether its in the viewport or not. On every scroll event, `UICollectionViewFlowLayout` will _again_ ask **EVERY** row for its size.

In the case of Emoji Search (which has 1877 rows/emojis), all 1877 rows have their item size recalculated on initial load/ scroll.

https://github.com/cashapp/redwood/assets/6900601/529fd97f-83f4-4f48-8d52-d7e6d9075e04

### After

On load, `UICollectionViewFlowLayout` will ask the rows **within the viewport** for their size, and guess the sizes of the other rows. On every scroll event, `UICollectionViewFlowLayout` will _again_ the rows **within the viewport** for their size.

In the case of Emoji Search (which has 1877 rows/emojis), about 12 rows have their item size recalculated on initial load/ scroll.

https://github.com/cashapp/redwood/assets/6900601/5d1e8bfb-79f4-4d23-b3eb-d175d5f92a66

The only con of this new approach, is that the scroll bar doesn't always scroll smoothly (as it doesn't if its estimated row heights for the rows off screen is correct). This is a completely fine trade-off IMO, particularly given the HUGE performance improvements we receive.

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -39,7 +39,6 @@ import kotlinx.cinterop.useContents
 import platform.CoreGraphics.CGFloat
 import platform.CoreGraphics.CGRect
 import platform.CoreGraphics.CGRectZero
-import platform.CoreGraphics.CGSize
 import platform.CoreGraphics.CGSizeMake
 import platform.Foundation.NSIndexPath
 import platform.Foundation.classForCoder
@@ -48,7 +47,9 @@ import platform.UIKit.UICollectionViewCell
 import platform.UIKit.UICollectionViewDataSourceProtocol
 import platform.UIKit.UICollectionViewDelegateFlowLayoutProtocol
 import platform.UIKit.UICollectionViewFlowLayout
+import platform.UIKit.UICollectionViewFlowLayoutAutomaticSize
 import platform.UIKit.UICollectionViewLayout
+import platform.UIKit.UICollectionViewLayoutAttributes
 import platform.UIKit.UICollectionViewScrollDirection.UICollectionViewScrollDirectionHorizontal
 import platform.UIKit.UICollectionViewScrollDirection.UICollectionViewScrollDirectionVertical
 import platform.UIKit.UICollectionViewScrollPositionTop
@@ -64,7 +65,9 @@ import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
 internal open class UIViewLazyList(
-  private var collectionViewFlowLayout: UICollectionViewFlowLayout = UICollectionViewFlowLayout(),
+  private var collectionViewFlowLayout: UICollectionViewFlowLayout = UICollectionViewFlowLayout().apply {
+    estimatedItemSize = UICollectionViewFlowLayoutAutomaticSize.readValue()
+  },
   internal val collectionView: UICollectionView = UICollectionView(
     CGRectZero.readValue(),
     collectionViewFlowLayout,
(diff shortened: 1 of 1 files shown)


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -218,6 +212,19 @@ private class LazyListContainerCell(frame: CValue<CGRect>) : UICollectionViewCel
     super.layoutSubviews()
     widgetView?.setFrame(this.contentView.bounds)
   }
+
+  override fun preferredLayoutAttributesFittingAttributes(
+    layoutAttributes: UICollectionViewLayoutAttributes,
+  ): UICollectionViewLayoutAttributes {
+    val itemSize = widgetView!!.sizeThatFits(collectionView.frame.useContents { size.readValue() })
+    return layoutAttributes.apply {
+      size = if (collectionViewFlowLayout.scrollDirection == UICollectionViewScrollDirectionVertical) {
```
dnagler, line 221: Have you tried setting `frame.size` instead of `size`? Not sure if that would impact performance one way or the other.

dnagler (approved):
This is huge, thank you!

One other question: do you see any impact on search performance at all?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -95,6 +98,9 @@ internal open class UIViewLazyList(
           forIndexPath = cellForItemAtIndexPath,
         ) as LazyListContainerCell
 
+        cell.collectionView = collectionView
```
apatronl: I'm not sure whether a cell should hold a reference to the collection view, could it cause a reference cycle? Maybe the var should be a weak reference

apatronl: Or just pass the collection view's frame if that's all we need within the cell

JakeWharton (approved):
> The only con of this new approach, is that the scroll bar doesn't always scroll smoothly (as it doesn't if its estimated row heights for the rows off screen is correct).

For heterogeneous contents this makes sense (and is the same on Android and web), but it's weird to see this happening in the after video with homogeneous rows whose height is always the same. I wouldn't have expected that behavior in this sample.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
veyndan: I have but it just didn't apply any constraints. It just defaults to `itemSize`, [which is 50x50 by default](https://developer.apple.com/documentation/uikit/uicollectionviewflowlayout/1617711-itemsize), so ends up appearing in a grid with each item clipped.

dnagler: Got it, thanks!

veyndan:
> One other question: do you see any impact on search performance at all?

Didn't see any performance difference. At least on my machine, the previous search performance was decent though.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
veyndan: Good catch! Just passing in the `CollectionView` frame now.

veyndan:
> > The only con of this new approach, is that the scroll bar doesn't always scroll smoothly (as it doesn't if its estimated row heights for the rows off screen is correct).
> 
> For heterogeneous contents this makes sense (and is the same on Android and web), but it's weird to see this happening in the after video with homogeneous rows whose height is always the same. I wouldn't have expected that behavior in this sample.

True. Presumably it's incorrectly guessing the layout size for rows that are not on screen. I don't know how `UICollectionViewFlowLayoutAutomaticSize` does its estimation, but that's probably the root cause of it. This'd probably require a follow up to figure that out, but I'm somewhat out of my depth here.

PR #1392, opened 2023-08-09
Convert public API data classes to Poko

### LINKED ISSUE #1043: Eliminate `data class` from public API

drewhamilton:
Reboot of #1366. Closes #1043.

Also replaces lists in public API with arrays, which are more performant on JS.

The arrays cause problems with some of the serialization tests: lists were omitted by default; arrays are included by default because different instances of an empty array are not `equals`.

This PR half-implements a verbose solution – custom serializers for all serializable classes containing an array property. Happy to add the remaining custom serializers if this is the preferred solution. I'm hoping there's a less verbose way though.

Another option is to give these classes singleton default empty array values, but this is brittle because it's easy to write new code generation that doesn't know it should use these singletons. Also adds to the public API in a weird way.

A final option is to punt the array migration to a separate PR and just migrate to Poko with the existing lists in this PR.

Before merging:
- [x] Decide and implement preferred array solution
- [x] Update to a legit Poko release

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,6 @@ buildscript {
     mavenCentral()
     google()
     gradlePluginPortal()
-    maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots") }
   }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@ buildscript {
     classpath libs.buildConfigPlugin
     classpath libs.zipline.gradlePlugin
     classpath libs.paparazzi.gradlePlugin
+    classpath libs.poko.gradlePlugin
     classpath 'app.cash.redwood.build:gradle-plugin'
     classpath 'app.cash.redwood:redwood-gradle-plugin'
   }
@@ -22,6 +23,7 @@ buildscript {
     mavenCentral()
     google()
     gradlePluginPortal()
+    maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots") }
   }
 }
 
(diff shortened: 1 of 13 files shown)


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/composeProtocolGeneration.kt
```diff
@@ -704,14 +705,15 @@ internal fun generateComposeProtocolModifierSerialization(
...
-        .returns(LIST.parameterizedBy(Protocol.ModifierElement))
+        .returns(ARRAY.parameterizedBy(Protocol.ModifierElement))
         .beginControlFlow("return %M", Stdlib.buildList)
         .addStatement(
           "this@%L.forEach { element -> add(element.%M(json)) }",
           name,
           schema.modifierToProtocol,
         )
         .endControlFlow()
+        .addCode("⇥.toTypedArray()⇤")
```
drewhamilton: Not convinced that just adding this actually helps perf since it's still building a list first. But I think to build an array directly `Modifier` would need a `size` and an iterator.

JakeWharton:
> A final option is to punt the array migration to a separate PR and just migrate to Poko with the existing lists in this PR.

Yeah I think this is the play. It lets us land this, the blog post gets to say it uses it, and the decisions around the additional pain of arrays can be done separately.

drewhamilton:
Done. I'll open the array PR after this is merged, to avoid it auto-closing.

JakeWharton (approved):

PR #1403, opened 2023-08-11
Move common Android config into build plugin

JakeWharton:


build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildPlugin.kt
```diff
@@ -112,6 +115,7 @@ class RedwoodBuildPlugin : Plugin<Project> {
    */
   private fun Project.configureCommonCompose() {
     plugins.withId("com.android.base") {
+      println("XXXXX $path")
```
colinrtwhite: Leftover debug code?

colinrtwhite (approved):

veyndan (approved):

build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildPlugin.kt
JakeWharton:
```suggestion
```

JakeWharton: Whoops, yep!

Issue #1054, opened 2023-05-11
`UiConfiguration` should be part of Redwood, not Treehouse


JakeWharton:
All of the things it exposes (and will expose) are useful to Compose regardless of whether it's running natively or within Zipline.

PR #1410, opened 2023-08-16
Move `StateSnapshot` into Redwood

veyndan:
`TreehouseView` depends on `StateSnapshot`. In order to introduce a `RedwoodView` that is similar to `TreehouseView` (https://github.com/cashapp/redwood/issues/1054), `StateSnapshot` needs to be moved into Redwood.

### DIFF
--- a/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/StateSnapshot.kt
+++ b/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/StateSnapshot.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
```diff
@@ -15,79 +15,8 @@
...
-      // TODO add other primitive types (double, float, long) when needed
-    }
-
-    is JsonArray -> listOf({ this.forEach { it.toJsonElement() } })
-    // TODO: map, numbers
-    // is Map<*, *> -> JsonElement
-    else -> error("unexpected type: $this")
-  }
-}
+public typealias StateSnapshot = app.cash.redwood.StateSnapshot
```
veyndan, line 22: This doesn't work, because `StateSnapshot.Id` isn't type aliased. Any ideas?

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/StateSnapshot.kt
veyndan, line 1: I'm assuming that `redwood-runtime` is the correct place to put this. Better ideas are welcome!

JakeWharton:
I'm not sure this has to move. This type is probably specific to the Treehouse implementation of state saving as the thing which moves across the JS bridge. For Redwood with direct Compose, we can tie the `SavedStateRegistry` directly to the one installed on the view tree or Compose UI tree and basically get normal state saving.

veyndan:
@JakeWharton That makes sense. I'll close this one out in favour of #1426.

PR #1413, opened 2023-08-17
Expose viewport size in `UiConfiguration`

JakeWharton:
Views:

views

Compose UI:

composeui

UIKit:

Simulator Screen Shot - iPhone 14 Pro - 2023-08-22 at 13 26 33

### DIFF
--- a/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/UiConfiguration.kt
+++ b/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/UiConfiguration.kt
@@ -23,6 +23,15 @@ import kotlinx.serialization.Serializable
 public class UiConfiguration(
   public val darkMode: Boolean = false,
   public val safeAreaInsets: Margin = Margin.Zero,
+  /**
+   * The size of the viewport into which the composition is rendering. This could be as lage as the
+   * entire screen or as small as an individual view within a larger native screen.
+   *
+   * This does not offer any information on the size of the individual composables which are
+   * rendering within the composition, but is the frame into which they will render. The root
+   * composable if stretching to fill the viewport will match this size.
+   */
+  public val viewportSize: Size = Size.Zero,
 ) {
   public companion object
 }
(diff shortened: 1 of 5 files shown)


JakeWharton:
We probably want to add density to this otherwise you can't size yourself appropriately. Alternatively we could expose it as DP directly.

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/UiConfiguration.kt
```diff
@@ -23,6 +23,15 @@ import kotlinx.serialization.Serializable
...
   public val safeAreaInsets: Margin = Margin.Zero,
+  /**
+   * The size of the viewport into which the composition is rendering. This could be as lage as the
+   * entire screen or as small as an individual view within a larger native screen.
+   *
+   * This does not offer any information on the size of the individual composables which are
+   * rendering within the composition, but is the frame into which they will render. The root
+   * composable if stretching to fill the viewport will match this size.
+   */
+  public val viewportSize: Size = Size.Zero,
```
colinrtwhite, line 34: I'd vote that we expose this as `Dp` values directly. We currently expose the `safeAreaInsets` with density independent pixels mostly so its values can be passed to other layout components (e.g. `Column(margin = safeAreaInsets)`.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIKitView.kt
```diff
@@ -72,16 +75,21 @@ public class TreehouseUIKitView(
...
 }
 
 private fun computeUiConfiguration(
   traitCollection: UITraitCollection,
+  bounds: CValue<CGRect>,
 ): UiConfiguration {
   return UiConfiguration(
     darkMode = traitCollection.userInterfaceStyle == UIUserInterfaceStyleDark,
     safeAreaInsets = computeSafeAreaInsets(),
+    viewportSize = bounds.useContents { Size(size.width.toInt(), size.height.toInt()) },
```
colinrtwhite: I think we should probably have `Size` use `Double`s (or `Dp`) instead of `Int`s. Unfortunately, iOS's coordinate system using `Double`s means we either use decimals everywhere or lose precision.

colinrtwhite (approved):

colinrtwhite:
@JakeWharton That's a good point. I think my preference is to expose everything in `UiConfiguration` using `Dp`s since that's what the guest code mostly deals with and that's what the layout components accept as parameters. We could still add `Density` to `UiConfiguration` in case they want to get the raw pixel value from a `Dp`.

JakeWharton:
Views:

views

Compose UI:

composeui

UIKit:

Simulator Screen Shot - iPhone 14 Pro - 2023-08-22 at 13 26 33

PR #1426, opened 2023-08-25
Introduce `RedwoodView`

### LINKED ISSUE #1116: Introduce `RedwoodView`
This is similar to `TreehouseView` but at the Redwood layer.

As `HostConfiguration` is moving to be within Redwood (#1054), we need the machinery that automatically monitors configuration changes and sends them into the composition to be turnkey code.

It _might_ be possible to eliminate `TreehouseView` as a result in favor of a smaller decorator around a `RedwoodView`.

veyndan:
The only Treehouse specific part of `TreehouseView` was the state restoration, which has now been pulled out into a separate interface, to be implemented on all platforms that support it (currently, just Android Views).

Closes #1116.

### DIFF
--- a/redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/RedwoodView.kt
+++ b/redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/RedwoodView.kt
@@ -13,42 +13,32 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package app.cash.redwood.treehouse
+package app.cash.redwood.protocol.widget
 
-import app.cash.redwood.protocol.widget.ProtocolMismatchHandler
-import app.cash.redwood.protocol.widget.ProtocolNode
 import app.cash.redwood.ui.UiConfiguration
 import app.cash.redwood.widget.Widget
 import kotlin.native.ObjCName
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.serialization.json.Json
 
-@ObjCName("TreehouseView", exact = true)
-public interface TreehouseView {
+@ObjCName("RedwoodView", exact = true)
+public interface RedwoodView {
   public val children: Widget.Children<*>
   public val uiConfiguration: StateFlow<UiConfiguration>
   public val widgetSystem: WidgetSystem
   public val readyForContent: Boolean
   public var readyForContentChangeListener: ReadyForContentChangeListener?
-  public var saveCallback: SaveCallback?
-  public val stateSnapshotId: StateSnapshot.Id
 
   /** Invoked when new code is loaded. This should at minimum clear all [children]. */
   public fun reset()
 
-  @ObjCName("TreehouseViewReadyForContentChangeListener", exact = true)
+  @ObjCName("RedwoodViewReadyForContentChangeListener", exact = true)
   public fun interface ReadyForContentChangeListener {
-    /** Called when [TreehouseView.readyForContent] has changed. */
-    public fun onReadyForContentChanged(view: TreehouseView)
+    /** Called when [RedwoodView.readyForContent] has changed. */
+    public fun onReadyForContentChanged(view: RedwoodView)
   }
 
-  @ObjCName("TreehouseViewSaveCallback", exact = true)
-  public interface SaveCallback {
-    /** Called on the UI dispatcher to save the state for the current content. */
-    public fun performSave(id: String)
-  }
-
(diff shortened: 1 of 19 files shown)


redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/RedwoodView.kt
veyndan, line 1: Where should `RedwoodView` exist?

I was running into a circular dependency issue when putting it in `redwood-widget` (which is where I think it should exist). The circular dependency is the `protocol-widget` references, which is only referenced via the `WidgetSystem` subclass.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/Saveable.kt
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+import kotlin.native.ObjCName
+
+@ObjCName("Saveable", exact = true)
+public interface Saveable {
```
veyndan, line 21: Open to better names!

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/RedwoodView.kt
```diff
@@ -13,42 +13,32 @@
...
 
-@ObjCName("TreehouseView", exact = true)
-public interface TreehouseView {
+@ObjCName("RedwoodView", exact = true)
+public interface RedwoodView {
   public val children: Widget.Children<*>
   public val uiConfiguration: StateFlow<UiConfiguration>
   public val widgetSystem: WidgetSystem
   public val readyForContent: Boolean
   public var readyForContentChangeListener: ReadyForContentChangeListener?
```
JakeWharton, line 30: The need for these three things is very suspect to me. We shouldn't need any of these in pure Redwood.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/RedwoodView.kt
```diff
@@ -13,42 +13,32 @@
...
-  public var saveCallback: SaveCallback?
-  public val stateSnapshotId: StateSnapshot.Id
 
   /** Invoked when new code is loaded. This should at minimum clear all [children]. */
   public fun reset()
```
JakeWharton, line 33: Same with this–Redwood will never invoke this.

JakeWharton (commented):
I think we should probably start with a new interface that exposes only the children + ui config flow and try to create implementations for that. Then, once that's working, we can see whether it makes sense for `TreehouseView` to extend that new interface, or maybe just wrap each implementation.

veyndan:
Makes sense about the separate `RedwoodView` interface in addition to `TreehouseView`. 

> […] start with a new interface that exposes only the children + ui config flow and try to create implementations for that.

Created a PR for that at https://github.com/cashapp/redwood/pull/1435. I'll see if `TreehouseView` can extend from `RedwoodView` in a follow-up PR.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/RedwoodView.kt
veyndan: Superseded by https://github.com/cashapp/redwood/pull/1435#discussion_r1310631358.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/RedwoodView.kt
veyndan: `RedwoodView` has been introduced in https://github.com/cashapp/redwood/pull/1435 and no longer includes these three things, nor `reset()`.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/Saveable.kt
veyndan: No longer relevant. See https://github.com/cashapp/redwood/pull/1435.

veyndan:
Superseded by https://github.com/cashapp/redwood/pull/1435.

PR #1427, opened 2023-08-25
Try to avoid the Compose compiler plugin being applied twice

JakeWharton:
This can cause errors which are hard to debug such as impossible ClassCastExceptions.


samples/emoji-search/android-composeui/build.gradle
```diff
@@ -1,6 +1,5 @@
 apply plugin: 'com.android.application'
 apply plugin: 'org.jetbrains.kotlin.android'
-apply plugin: 'app.cash.redwood'
```
JakeWharton, line 3: This is why #1421 is failing. The rest of the PR is a check for this condition.

JakeWharton: Also #1430

colinrtwhite (approved):

PR #1435, opened 2023-08-30
Introduce `RedwoodView` and create platform-specific implementations

veyndan:


redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/RedwoodWidgetView.kt
veyndan, line 1: I'll move this class and `RedwoodUIKitView` to `redwood-widget` in a follow-up PR.

JakeWharton: Interesting. I never really thought where these types would have to end up. We might need a new module for this eventually, but for now let's try to keep it in -widget.

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/RedwoodWidgetView.kt
```diff
@@ -0,0 +1,78 @@
...
+import app.cash.redwood.ui.Density
+import app.cash.redwood.ui.Size
+import app.cash.redwood.ui.UiConfiguration
+import app.cash.redwood.widget.RedwoodView
+import app.cash.redwood.widget.ViewGroupChildren
+import app.cash.redwood.widget.Widget
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+public abstract class RedwoodWidgetView(
```
JakeWharton:
```suggestion
public open class RedwoodWidgetView(
```
I think this is usable as-is for normal Redwood usage.

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseWidgetView.kt
```diff
@@ -60,22 +49,8 @@ public class TreehouseWidgetView(
...
-
-  init {
-    setOnWindowInsetsChangeListener { insets ->
-      mutableUiConfiguration.value = computeUiConfiguration(insets = insets.safeDrawing)
-    }
-  }
-
   override fun reset() {
-    _children.remove(0, _children.widgets.size)
+    children.remove(0, (children as ViewGroupChildren).widgets.size)
```
JakeWharton, line 53: Hmm don't love that we have to do this, but can deal with it later.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIKitView.kt
```diff
@@ -17,30 +17,17 @@ package app.cash.redwood.treehouse
...
 import platform.UIKit.UITraitCollection
-import platform.UIKit.UIUserInterfaceStyle.UIUserInterfaceStyleDark
 import platform.UIKit.UIView
 
 @ObjCName("TreehouseUIKitView", exact = true)
 public class TreehouseUIKitView(
   override val widgetSystem: WidgetSystem,
-) : TreehouseView {
-  public val view: UIView = RootUiView(this)
+) : TreehouseView, RedwoodUIKitView() {
```
JakeWharton: Future change: We probably can drop the "Kit" from these names. Feels out of place and doesn't fit the normal naming pattern of their types.

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/RedwoodView.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+package app.cash.redwood.widget
+
+import app.cash.redwood.ui.UiConfiguration
+import kotlin.native.ObjCName
+import kotlinx.coroutines.flow.StateFlow
+
+@ObjCName("RedwoodView", exact = true)
+public interface RedwoodView {
+  public val children: Widget.Children<*>
```
JakeWharton, line 24: This type should probably have a generic which is then used for `children`. Future usages of this type will use the generic to ensure you are using a correctly-typed `Widget.Factory<T>`.

JakeWharton (approved):

JakeWharton:
Another future change: Migrate the counter sample to use these views.

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/RedwoodWidgetView.kt
veyndan: I just made it abstract to align with the requirement that `RedwoodUIKitView` is abstract. `RedwoodUIKitView` probably should be `open` as well though.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIKitView.kt
veyndan: Should we also rename `TreehouseWidgetView`? All the implementations of `TreehouseView` are "widget" views. Something like `TreehouseViewView` would be awful but consistent. `TreehouseAndroidView` perhaps?

veyndan: Actually, not a big fan of `TreehouseAndroidView` either. I'm drawing a blank on a better name tbh

JakeWharton: `TreehouseLayout`?

veyndan: Is "Layout" Android specific enough?

veyndan: Renamed to `TreehouseLayout` in https://github.com/cashapp/redwood/pull/1451.
Renamed to `TreehouseUIView` in https://github.com/cashapp/redwood/pull/1452.

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/RedwoodView.kt
veyndan: Fixed in https://github.com/cashapp/redwood/pull/1442.

PR #1436, opened 2023-08-30
AndroidX Compose compiler 1.5.3

JakeWharton:
Split from #1419, again


JakeWharton:
Weird I thought this supported both 1.9.0 and 1.9.10. Guess not.

PR #1440, opened 2023-08-30
Move `RedwoodUIKitView` to `redwood-widget`

veyndan:


redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIKitView.kt
```diff
@@ -53,6 +52,29 @@ public open class RedwoodUIKitView(
   }
 }
 
+public class RootUiView : UIView(cValue { CGRectZero }) {
```
veyndan: I really don't want to make this public. I found it hard to appease all the Kotlin/Native constraints and making this non-public (i.e., disallowing an open class extending `UIView`, disallowing a class extending `UIView` and a Kotlin class).

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIKitView.kt
```diff
@@ -43,7 +42,8 @@ public class TreehouseUIKitView private constructor(
   public constructor(widgetSystem: WidgetSystem) : this(widgetSystem, RootUiView())
 
   init {
-    (view as RootUiView).treehouseView = this
+    (view as RootUiView).redwoodView = this
+    view.didMoveToSuperview = ::superviewChanged
```
veyndan: This is terrible

JakeWharton: I'm pretty sure this cycle will create a memory leak. @efirestone?

But this is feeling worse and worse in terms of layering. We gotta do something different here long-term.

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIKitView.kt
JakeWharton: Yeah this shouldn't need to be in the public API surface. We definitely need to figure out something else here. Maybe it's best not to try and migrate the Treehouse views right away and instead focus on migrating the Counter sample to the new views. Then we can reconcile how Treehouse and Redwood layer on top of each other later.

JakeWharton (approved):

JakeWharton:
Hah we commented on exactly the same things.

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIKitView.kt
veyndan: Working on migrating the Counter sample first sounds good.

I'm still trying to grok how `RedwoodView` would help us within a Redwood-only app, when `RedwoodComposition` is already (somewhat) solving the problem. Is the only benefit having `LocalUiConfiguration` set up? If so, I'm assuming the migration depends on finishing up the work to close https://github.com/cashapp/redwood/issues/1054?

If you have any psuedocode for what you think [what `onCreate` in `MainActivity`](https://github.com/cashapp/redwood/blob/trunk/samples/counter/android-views/src/main/kotlin/com/example/redwood/counter/android/views/MainActivity.kt) should look like, it'd be much appreciated!

JakeWharton: So there's a near-term and a long-term ideal.

For the near-term, I'd be happy with
```kotlin
class MainActivity : AppCompatActivity() {
  private val scope = CoroutineScope(AndroidUiDispatcher.Main)

  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    val redwoodView = RedwoodWidgetView(this)
    setContentView(redwoodView)

    val composition = RedwoodComposition(
      scope = scope,
      container = redwoodView.children,
      provider = SchemaWidgetFactories(
        Schema = AndroidWidgetFactory(this),
        RedwoodLayout = ViewRedwoodLayoutWidgetFactory(this),
      ),
    )
    composition.setContent {
      Counter()
    }
  }

  override fun onDestroy() {
    scope.cancel()
    super.onDestroy()
  }
}
```

Long term I think we want:
```kotlin
class MainActivity : AppCompatActivity() {
  private val scope = CoroutineScope(AndroidUiDispatcher.Main)
    super.onCreate(savedInstanceState)

    val redwoodView = RedwoodLayout(this)
    setContentView(redwoodView)

    redwoodView.setWidgetProvider(
      SchemaWidgetFactories(
        Schema = AndroidWidgetFactory(this),
        RedwoodLayout = ViewRedwoodLayoutWidgetFactory(this),
      )
    )
    redwoodView.setContent {
      Counter()
    }
  }
}
```

> Is the only benefit having `LocalUiConfiguration` set up? If so, I'm assuming the migration depends on finishing up the work to close https://github.com/cashapp/redwood/issues/1054?

This is the goal, yes. But the dependency is inverted. We need a standard mechanism of getting a `StateFlow<UiConfiguration>` before we can make it a required argument for `RedwoodComposition`.

veyndan: Sweet, thanks!

> We need a standard mechanism of getting a `StateFlow<UiConfiguration>`

Isn't the standard mechanism just `RedwoodView#uiConfiguration`?

JakeWharton: It will be, yes. But we need one of those for Android views, iOS UIKit, Compose UI, and the DOM and so far we only have about half sorta working. Once those implementations are done we can update `RedwoodComposition` to take a `RedwoodView` and automatically bind the composition local, thus completing #1054. And somewhere in there we can update the samples.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIKitView.kt
veyndan: Resolving in favour of the more descriptive comment below 😄

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIKitView.kt
veyndan: I made this `private` within `TreehouseUIView` again.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIKitView.kt
veyndan: Managed to remove the need for `view.didMoveToSuperview = ::superviewChanged`. Not too sure how to remove `(view as RootUiView).redwoodView = this`, but at least this PR no longer exacerbates the problem!

PR #1453, opened 2023-09-06
Add `RedwoodView` which wraps `HTMLElement`

JakeWharton:
Refs #1116

### DIFF
--- a/redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
+++ b/redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.widget
+
+import app.cash.redwood.ui.Size
+import app.cash.redwood.ui.UiConfiguration
+import app.cash.redwood.ui.dp
+import app.cash.redwood.widget.Widget.Children
+import kotlinx.browser.window
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.update
+import org.w3c.dom.HTMLElement
+import org.w3c.dom.MediaQueryList
+
+public fun HTMLElement.asRedwoodView(): RedwoodView<HTMLElement> {
+  checkNotNull(parentNode) {
+    "Element $this must be attached to Document to be used as RedwoodView"
+  }
+  return RedwoodHTMLElementView(this)
+}
+
+private class RedwoodHTMLElementView(
+  element: HTMLElement,
+) : RedwoodView<HTMLElement> {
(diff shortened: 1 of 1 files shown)


redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.coroutines.flow.update
+import org.w3c.dom.HTMLElement
+import org.w3c.dom.MediaQueryList
+
+public fun HTMLElement.asRedwoodView(): RedwoodView<HTMLElement> {
+  checkNotNull(parentNode) {
+    "Element $this must be attached to Document to be used as RedwoodView"
+  }
+  return RedwoodHTMLElementView(this)
+}
```
veyndan, line 33: If we go with an extension function on the platform specific widgets to create a `RedwoodView`, we should probably exclusively allow instantiation of a `RedwoodView` via an extension function on other platforms too, for consistency/consistent discovery.

veyndan (approved):

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
```diff
@@ -0,0 +1,67 @@
...
+  element: HTMLElement,
+) : RedwoodView<HTMLElement> {
+  override val children: Children<HTMLElement> = HTMLElementChildren(element)
+  override val uiConfiguration = MutableStateFlow(
+    UiConfiguration(
+      darkMode = window.matchMedia("(prefers-color-scheme: dark)").matches,
+      safeAreaInsets = Margin.Zero,
+      viewportSize = Size(width = element.offsetWidth.dp, height = element.offsetHeight.dp),
+      density = 1.0,
+    )
```
veyndan: Remove explicit setting of defaults?

```suggestion
    UiConfiguration(
      darkMode = window.matchMedia("(prefers-color-scheme: dark)").matches,
      viewportSize = Size(width = element.offsetWidth.dp, height = element.offsetHeight.dp),
    )
```

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
JakeWharton: So I want to eventually switch this to be a web component where you do actually create it yourself, but as an actual node in the DOM tree (kinda like an Android view) and then you look up the reference. But that requires emitting ES2015 JS which Kotlin doesn't fully support yet. Without that, we have no choice but to wrap an existing `HTMLElement` type. I guess I could make it more restrictive such as requiring that it's a blank `<div>`, but I don't think we have a good way to make it consistent on every platform because of this difference.

veyndan: I wasn't even considering that you have to create an instance of `HTMLElement` in order to use this. My thinking was more along the lines of:

1. Android Views — Get a `RedwoodView` by instantiating a `RedwoodLayout(…)`
2. iOS — Get a `RedwoodView` by instantiating a `RedwoodUIView(…)`
3. DOM — Get a `RedwoodView` via the `asRedwoodView()` extension function (which is surprising if you've implemented things with Android Views or iOS).

My thinking was that on the DOM we should get a `RedwoodView` by instantiating a `RedwoodHTMLElementView(…)`, or we update the ways of instantiating a `RedwoodView` on the other platforms, i.e., `Context.asRedwoodView()` (not the greatest), and `UIView.asRedwoodView()`.

Not a big deal if we plan to move over to web components at some point.

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
JakeWharton: Yep, done.

PR #1454, opened 2023-09-06
Make `UiConfiguration` part of Redwood

### LINKED ISSUE #1054: `UiConfiguration` should be part of Redwood, not Treehouse

veyndan:
@JakeWharton Regarding your comment at https://github.com/cashapp/redwood/pull/1440#discussion_r1312048461, I wanted to ensure that I have the correct north star, by implementing what you mentioned and having it only work for Android Views (for now). Please lmk if this is/isn't what you had in mind!

I'll create the remaining `RedwoodView` implementations so this can work on other platforms in some other PRs and keep rebasing this PR and keeping at as a draft until all the implementations are complete.

Closes https://github.com/cashapp/redwood/issues/1054

### DIFF
--- a/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
+++ b/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
@@ -19,13 +19,17 @@ import androidx.compose.runtime.Applier
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.ComposeNode
 import androidx.compose.runtime.Composition
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.DisallowComposableCalls
 import androidx.compose.runtime.MonotonicFrameClock
 import androidx.compose.runtime.Recomposer
 import androidx.compose.runtime.Updater
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.currentComposer
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.snapshots.Snapshot
 import app.cash.redwood.RedwoodCodegenApi
+import app.cash.redwood.widget.RedwoodView
 import app.cash.redwood.widget.Widget
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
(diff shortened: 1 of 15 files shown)


redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
```diff
@@ -43,16 +47,17 @@ public interface RedwoodComposition {
  */
 public fun <W : Any> RedwoodComposition(
   scope: CoroutineScope,
-  container: Widget.Children<W>,
+  view: RedwoodView<W>,
   provider: Widget.Provider<W>,
   onEndChanges: () -> Unit = {},
 ): RedwoodComposition {
```
JakeWharton: If you keep an overload of this which only adds the `StateFlow<UiConfiguration>` to the original signature and then unpack the `RedwoodView` into its two parts in the new overload...

redwood-protocol-compose/src/commonMain/kotlin/app/cash/redwood/protocol/compose/ProtocolRedwoodComposition.kt
```diff
@@ -32,8 +36,13 @@ public fun ProtocolRedwoodComposition(
   bridge: ProtocolBridge,
   changesSink: ChangesSink,
   widgetVersion: UInt,
+  uiConfigurations: StateFlow<UiConfiguration>,
 ): RedwoodComposition {
-  val composition = RedwoodComposition(scope, bridge.root, bridge.provider) {
+  val view = object : RedwoodView<Nothing> {
+    override val children: Widget.Children<Nothing> = bridge.root
+    override val uiConfiguration: StateFlow<UiConfiguration> = uiConfigurations
+  }
```
JakeWharton: ...that will eliminate the need to create this temporary object.

Plus I think the testing support will probably also want to use an overload where they can explicitly pass the `StateFlow` + `Widget.Children` separately.

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
```diff
@@ -75,7 +80,12 @@ private class WidgetRedwoodComposition(
   }
 
   override fun setContent(content: @Composable () -> Unit) {
-    composition.setContent(content)
+    composition.setContent {
+      val uiConfiguration by view.uiConfiguration.collectAsState()
```
JakeWharton: and that has a side-effect of removing this `view.uiConfiguration` call which will happen on each recomposition to determine if the `StateFlow` instance changed.

JakeWharton (approved):
Yep this is precisely what I had in mind.

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
veyndan: Good idea! I just added a new commit with this (and the following) changes.

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
veyndan: My knowledge of recomposition is mediocre. Does the latest commit prevent the recomposition you're talking about?

JakeWharton: Yep!

PR #1459, opened 2023-09-07
Move insets to Redwood-only module

veyndan:
This is in preparation for allowing `UiConfiguration` to be provided within a Redwood-only & Compose UI-specific app.

Refs https://github.com/cashapp/redwood/issues/1054

### DIFF
--- a/redwood-composeui/build.gradle
+++ b/redwood-composeui/build.gradle
@@ -14,6 +14,7 @@ kotlin {
   sourceSets {
     commonMain {
       dependencies {
+        implementation libs.jetbrains.compose.foundation
         api projects.redwoodCompose
         implementation projects.redwoodWidgetCompose
       }
--- a/redwood-composeui/src/androidMain/kotlin/app/cash/redwood/composeui/insets.kt
+++ b/redwood-composeui/src/androidMain/kotlin/app/cash/redwood/composeui/insets.kt
@@ -15,7 +15,7 @@
  */
 @file:JvmName("insetsAndroid")
 
-package app.cash.redwood.treehouse.composeui
+package app.cash.redwood.composeui
 
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.WindowInsets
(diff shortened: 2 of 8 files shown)


redwood-widget-composeui/src/commonMain/kotlin/app/cash/redwood/widget/composeui/insets.kt
```diff
@@ -13,11 +13,11 @@
...
-package app.cash.redwood.treehouse.composeui
+package app.cash.redwood.widget.composeui
 
 import androidx.compose.runtime.Composable
 import app.cash.redwood.ui.Margin
 
 /** Return the device's insets in screen density-independent pixels. */
 @Composable
-internal expect fun safeAreaInsets(): Margin
+public expect fun safeAreaInsets(): Margin
```
JakeWharton: Does this revert once `RedwoodContent` exists in this module?

redwood-widget-composeui/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+    publishLibraryVariants('release')
+  }
+
+  KmpTargets.addAllTargets(project, true /* skipJs */)
+
+  sourceSets {
+    commonMain {
+      dependencies {
+        implementation libs.jetbrains.compose.foundation
+        implementation projects.redwoodRuntime
```
JakeWharton:
```suggestion
        api projects.redwoodRuntime
```
For `Margin`

JakeWharton (approved):

redwood-widget-composeui/src/commonMain/kotlin/app/cash/redwood/widget/composeui/insets.kt
JakeWharton: After looking at the other PR, I think this should move into `redwood-composeui` and not into a new module.

veyndan: > Does this revert once RedwoodContent exists in this module?

Nope, cause we need it for `TreehouseContent`. Though if we make `TreehouseContent` a wrapper around `RedwoodContent`, we could revert it.

JakeWharton: I think long term that should be the goal, but it's hard to reason about it right now. We can leave it as-is while we figure it out.

redwood-widget-composeui/build.gradle
veyndan: I've moved the insets over to `redwood-composeui`. `redwood-composeui` has an `api` dependency on `redwood-compose`, which has an `api` dependency on `redwood-runtime`, so we no longer need to explicitly specify a dependency on `redwood-runtime`.

Issue #883, opened 2023-03-24
Parameterize usage of `TreehouseView`


veyndan:
I've just come into the case where I have a `CodeListener` and I want to do something with the `TreehouseView` argument of `CodeListener#onInitialCodeLoading(TreehouseView)`. This `TreehouseView` is _definitely_ a `TreehouseWidgetView`, as the host widgets are Android Views. At the moment, I have to cast the `TreehouseView` to a `TreehouseWidgetView` which feels icky and potentially prone to future runtime errors.

It'd be awesome if `TreehouseView` was a type parameter in `CodeListener`. This will probably result in this parametrization being propagated to many other places too.

JakeWharton:
See #865. It is not safe to do this and it only works on Android views and not Compose UI or UIKit.

Ideally you are capturing the concrete instance of the view in some kind of listener and then using that to listen to code changes. The current design isn't one that will work long term.

JakeWharton:
There's also direct arguments against this, for what it's worth. You could be driving two separate compositions with different UI toolkits within the same Zipline.

veyndan:
@JakeWharton Are there any arguments against parametrising the `Widget.Children` within `TreehouseView`? It should work for Android Views, Compose UI, and UIKit.

```kotlin
public interface TreehouseView<W : Any> {
  public val children: Widget.Children<W>
  public val widgetSystem: WidgetSystem<W>
  public var readyForContentChangeListener: ReadyForContentChangeListener<W>?
  // etc.

  public fun interface ReadyForContentChangeListener<W : Any> {
    public fun onReadyForContentChanged(view: TreehouseView<W>)
  }

  public fun interface WidgetSystem<W : Any> {
    public fun widgetFactory(
      json: Json,
      protocolMismatchHandler: ProtocolMismatchHandler,
    ): ProtocolNode.Factory<W>
  }
}
```

Such a change would mean that `TreehouseView` could extend from `RedwoodView`.

PR #1468, opened 2023-09-08
Give `TreehouseView` a generic for its Widget type

veyndan:
A PR to showcase what https://github.com/cashapp/redwood/issues/883#issuecomment-1711507473 could look like.

### DIFF
--- a/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseView.kt
+++ b/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseView.kt
@@ -24,12 +24,12 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.serialization.json.Json
 
 @ObjCName("TreehouseView", exact = true)
-public interface TreehouseView {
-  public val children: Widget.Children<*>
+public interface TreehouseView<W : Any> {
+  public val children: Widget.Children<W>
   public val uiConfiguration: StateFlow<UiConfiguration>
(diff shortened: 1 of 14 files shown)


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseView.kt
```diff
@@ -24,22 +24,22 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.serialization.json.Json
 
 @ObjCName("TreehouseView", exact = true)
-public interface TreehouseView {
-  public val children: Widget.Children<*>
+public interface TreehouseView<W : Any> {
+  public val children: Widget.Children<W>
```
veyndan, line 28: Having `children` parametrised is technically the only requirement for having `TreehouseView` extend from `RedwoodView`. The other parametrisations within the file are purely in the hope of improving type safety.

JakeWharton (approved):
I think this is an improvement. Long term I'd still like to investigate eliminating TreehouseView subtypes in favor of a single wrapper now that all the complexity was pushed to RedwoodView.

veyndan:
How would we eliminate `TreehouseView` subtypes, when the implementation of `TreehouseView`-specific fields/methods are platform-specific (e.g., in `TreehouseLayout` we have to override `onAttachedToWindow` in order to set `readyForContent = true`).

JakeWharton:
Expose the attached stats and also use it to pause the recomposition for plain Redwood.

PR #1470, opened 2023-09-11
Elimate cocoapods usage in samples

### LINKED ISSUE #1428: Eliminate cocoapods usage in samples
Copy https://github.com/JetBrains/compose-multiplatform-ios-android-template/pull/19

veyndan:
Closes https://github.com/cashapp/redwood/issues/1428 (apologies @ahmed3elshaer if you began any work on this!)


samples/counter/README.md
```diff
@@ -33,9 +33,7 @@ Running Counter on iOS
 
 Run this:
 ```
-cd samples/counter/ios-uikit
-pod install
-open CounterApp.xcworkspace
+open samples/counter/ios-uikit/CounterApp.xcworkspace
```
JakeWharton: These are also .gitignore'd. Do we need to commit this now?

samples/counter/ios-shared/build.gradle
```diff
@@ -15,17 +15,15 @@
...
-  iosSimulatorArm64()
-
-  cocoapods {
-    noPodspec()
-    framework {
+  [
+    iosArm64(),
+    iosX64(),
+    iosSimulatorArm64(),
+  ].each { iosTarget ->
```
JakeWharton, line 25: I _think_ we can just do this:
```suggestion
  targets.each { iosTarget ->
```

JakeWharton (approved):

samples/counter/ios-shared/build.gradle
JakeWharton: Moreover, it's super weird that we can invoke these functions twice and still only create a single target.

So, alternatively, consider deleting the three lines above and only having
```kotlin
[ iosArm64(), etc. ].each { .. }
```
so that they're only invoked once.

samples/counter/README.md
veyndan: If we switch it to `open samples/counter/ios-uikit/CounterApp.xcodeproj` it opens alright 🤷  Just gonna do that unless there's some benefit of `xcworkspace` here that I'm not seeing

samples/counter/ios-shared/build.gradle
veyndan: > Moreover, it's super weird that we can invoke these functions twice and still only create a single target.

The first invocation of the function is happening within the list constructor. Where's the invocation of the function happening the second time?

JakeWharton: Oh you deleted the original ones. Sorry reviewing on my phone so harder to keep track of what's going on.

ahmed3elshaer:
> Closes #1428 (apologies @ahmed3elshaer if you began any work on this!)

Glad that you beat me to it 😄 

I will pick something else

PR #1471, opened 2023-09-11
Reduce viewport updates in `LazyList` for improved perf with Zipline

### LINKED ISSUE #1256: Optimize this for less operations
https://github.com/cashapp/redwood/blob/a4bbf1c2c8e718becede7d49ab06acc9813b8f80/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt#L186

```text

import app.cash.redwood.lazylayout.widget.LazyList
import app.cash.redwood.lazylayout.widget.RefreshableLazyList
import app.cash.redwood.widget.ChangeListener
import app.cash.redwood.widget.MutableListChildren
import kotlin.math.max
import kotlinx.cinterop.CValue
import kotlinx.cinterop.ObjCClass
import kotlinx.cinterop.readValue
import kotlinx.cinterop.useContents
import platform.CoreGraphics.CGRect
import platform.CoreGraphics.CGRectZero
import platform.CoreGraphics.CGSize
import platform.CoreGraphics.CGSizeMake
import platform.Foundation.NSIndexPath
import platform.Foundation.classForCoder
import platform.UIKit.UICollectionView
import platform.UIKit.UICollectionViewCell
import platform.UIKit.UICollectionViewDataSourceProtocol
import platform.UIKit.UICollectionViewDelegateFlowLayoutProtocol
import platform.UIKit.UICollectionViewFlowLayout
import platform.UIKit.UICollectionViewLayout
import platform.UIKit.UICollectionViewScrollDirection
import platform.UIKit.UIControlEventValueChanged
import platform.UIKit.UIEdgeInsetsMake
import platform.UIKit.UIRefreshControl
import platform.UIKit.UIScrollView
import platform.UIKit.UIView
import platform.UIKit.item
import platform.darwin.NSInteger
import platform.darwin.NSObject

internal interface ViewPortItems : Widget.Children<UIView> {
  var placeholders: MutableList<Widget<UIView>>
  var itemsBefore: Int
  var itemsAfter: Int
  fun itemForGlobalIndex(index: Int): UIView
  fun itemCount(): Int
}

internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {

  override val items: ViewPortItems = object : ViewPortItems {
    override var placeholders = mutableListOf<Widget<UIView>>()

    override var itemsBefore: Int = 0
    override var itemsAfter: Int = 0

    private val viewPortList = mutableListOf<Widget<UIView>>()

    override fun insert(index: Int, widget: Widget<UIView>) {
      viewPortList.add(index, widget)
    }

    override fun move(fromIndex: Int, toIndex: Int, count: Int) {
      viewPortList.move(fromIndex, toIndex, count)
    }

    override fun remove(index: Int, count: Int) {
      viewPortList.remove(index, count = count)
    }

    override fun onModifierUpdated() {}

    // Fetch the item from the viewPortList relative to the entire collection view
    override fun itemForGlobalIndex(index: Int): UIView {
      val viewPortIndex: Int = max(index - itemsBefore, 0)

      viewPortList.getOrNull(viewPortIndex)?.let {
        return it.value
      }

      // If we don't have a value, fallback to our pools of placeholders
      val placeholderIndex = index % placeholders.size
      return placeholders.get(placeholderIndex).value
    }

    override fun itemCount(): Int {
      // TODO: replace with this with "ItemCounts"
      return max(itemsBefore - 1, 0) + viewPortList.count() + itemsAfter
    }
  }

  override val placeholder: Widget.Children<UIView> = MutableListChildren(list = items.placeholders)

  private val viewPortListCoordinator = object {
    var minIndex: Int = 0
    var maxIndex: Int = 0

    fun notifyViewportChanged() {
      onViewportChanged(minIndex, maxIndex)
    }

    fun updateViewport(minIndex: Int, maxIndex: Int) {
      if (minIndex != this.minIndex || maxIndex != this.maxIndex) {
        this.minIndex = minIndex
        this.maxIndex = maxIndex

        notifyViewportChanged()
      }
    }
  }

  // A callback to tell LazyList that we have an update to the window of items available
  private lateinit var onViewportChanged: (firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) -> Unit

  // UICollectionView + Protocols
  private var collectionViewFlowLayout: UICollectionViewFlowLayout =
    object : UICollectionViewFlowLayout() {}

  private val collectionViewDataSource: UICollectionViewDataSourceProtocol =
    object : NSObject(), UICollectionViewDataSourceProtocol {

      override fun collectionView(
        collectionView: UICollectionView,
        numberOfItemsInSection: NSInteger,
      ): NSInteger {
        return items.itemCount().toLong()
      }

      override fun collectionView(
        collectionView: UICollectionView,
        cellForItemAtIndexPath: NSIndexPath,
      ): UICollectionViewCell {
        val cell = collectionView.dequeueReusableCellWithReuseIdentifier(
          identifier = reuseIdentifier,
          forIndexPath = cellForItemAtIndexPath,
        ) as LazyListContainerCell

        val view = items.itemForGlobalIndex(cellForItemAtIndexPath.item.toInt())
        cell.set(view)

        return cell
      }
    }

  private val collectionViewDelegate: UICollectionViewDelegateFlowLayoutProtocol =
    object : NSObject(), UICollectionViewDelegateFlowLayoutProtocol {
      override fun collectionView(
        collectionView: UICollectionView,
        layout: UICollectionViewLayout,
        sizeForItemAtIndexPath: NSIndexPath,
      ): CValue<CGSize> {
        // TODO: Calculate size for individual items. This commented code currently hangs the UI Thread
//        val item = items.itemForGlobalIndex(sizeForItemAtIndexPath.item.toInt())
//        collectionView.frame().useContents {
//          item.sizeThatFits(size.readValue()).useContents {
//            return CGSizeMake(width, height)
//          }
//        }

        // TODO: Handle size for horizontal scrollDirection
        return CGSizeMake(collectionView.frame().useContents { size.width }, 64.0)
      }

      override fun scrollViewDidScroll(scrollView: UIScrollView) {
        val visibleIndexPaths = collectionView.indexPathsForVisibleItems()

        if (visibleIndexPaths.isNotEmpty()) {
          // TODO: Optimize this for less operations
          viewPortListCoordinator.updateViewport(
            visibleIndexPaths.minOf { (it as NSIndexPath).item.toInt() },
            visibleIndexPaths.maxOf { (it as NSIndexPath).item.toInt() },
          )
        }
      }
    }

  internal val collectionView = UICollectionView(
    frame = CGRectZero.readValue(),
    collectionViewLayout = this.collectionViewFlowLayout,
  ).apply {
    dataSource = collectionViewDataSource
    delegate = collectionViewDelegate
    prefetchingEnabled = true

    registerClass(
      LazyListContainerCell(CGRectZero.readValue()).classForCoder() as ObjCClass?,
      reuseIdentifier,
    )
  }

  // LazyList
  override fun isVertical(isVertical: Boolean) {
    if (!isVertical) {
      collectionViewFlowLayout.scrollDirection = if (isVertical) {
        UICollectionViewScrollDirection.UICollectionViewScrollDirectionVertical
      } else {
        UICollectionViewScrollDirection.UICollectionViewScrollDirectionHorizontal
      }
    }
  }


```

veyndan:
Closes https://github.com/cashapp/redwood/issues/1256.

### DIFF
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
@@ -45,24 +45,24 @@ internal fun LazyList(
   content: LazyListScope.() -> Unit,
 ) {
   val itemProvider = rememberLazyListItemProvider(content)
-  var lastVisibleItemIndex by remember { mutableStateOf(0) }
+  var lastPagedItemIndex by remember { mutableStateOf(0) }
   val itemsBefore = remember(state.firstVisibleItemIndex) { (state.firstVisibleItemIndex - OffscreenItemsBufferCount / 2).coerceAtLeast(0) }
-  val itemsAfter = remember(lastVisibleItemIndex, itemProvider.itemCount) { (itemProvider.itemCount - (lastVisibleItemIndex + OffscreenItemsBufferCount / 2).coerceAtMost(itemProvider.itemCount)).coerceAtLeast(0) }
+  val itemsAfter = remember(lastPagedItemIndex, itemProvider.itemCount) { (itemProvider.itemCount - (lastPagedItemIndex + OffscreenItemsBufferCount / 2).coerceAtMost(itemProvider.itemCount)).coerceAtLeast(0) }
   val scrollItemIndex = remember(state.scrollToItemTriggeredId) { ScrollItemIndex(state.scrollToItemTriggeredId, state.firstVisibleItemIndex) }
   var placeholderPoolSize by remember { mutableStateOf(20) }
   LazyList(
     isVertical,
     itemsBefore = itemsBefore,
     itemsAfter = itemsAfter,
-    onViewportChanged = { localFirstVisibleItemIndex, localLastVisibleItemIndex ->
-      val visibleItemCount = localLastVisibleItemIndex - localFirstVisibleItemIndex
+    onViewportChanged = { localFirstPagedItemIndex, localLastPagedItemIndex ->
+      val visibleItemCount = localLastPagedItemIndex - localFirstPagedItemIndex
       val proposedPlaceholderPoolSize = visibleItemCount + visibleItemCount / 2
       // We only ever want to increase the pool size.
       if (placeholderPoolSize < proposedPlaceholderPoolSize) {
         placeholderPoolSize = proposedPlaceholderPoolSize
       }
-      state.firstVisibleItemIndex = localFirstVisibleItemIndex
-      lastVisibleItemIndex = localLastVisibleItemIndex
+      state.firstVisibleItemIndex = localFirstPagedItemIndex
+      lastPagedItemIndex = localLastPagedItemIndex
     },
     width = width,
     height = height,
@@ -97,24 +97,24 @@ internal fun RefreshableLazyList(
   content: LazyListScope.() -> Unit,
 ) {
   val itemProvider = rememberLazyListItemProvider(content)
-  var lastVisibleItemIndex by remember { mutableStateOf(0) }
+  var lastPagedItemIndex by remember { mutableStateOf(0) }
   val itemsBefore = remember(state.firstVisibleItemIndex) { (state.firstVisibleItemIndex - OffscreenItemsBufferCount / 2).coerceAtLeast(0) }
-  val itemsAfter = remember(lastVisibleItemIndex, itemProvider.itemCount) { (itemProvider.itemCount - (lastVisibleItemIndex + OffscreenItemsBufferCount / 2).coerceAtMost(itemProvider.itemCount)).coerceAtLeast(0) }
+  val itemsAfter = remember(lastPagedItemIndex, itemProvider.itemCount) { (itemProvider.itemCount - (lastPagedItemIndex + OffscreenItemsBufferCount / 2).coerceAtMost(itemProvider.itemCount)).coerceAtLeast(0) }
   val scrollItemIndex = remember(state.scrollToItemTriggeredId) { ScrollItemIndex(state.scrollToItemTriggeredId, state.firstVisibleItemIndex) }
   var placeholderPoolSize by remember { mutableStateOf(20) }
   RefreshableLazyList(
     isVertical,
     itemsBefore = itemsBefore,
     itemsAfter = itemsAfter,
-    onViewportChanged = { localFirstVisibleItemIndex, localLastVisibleItemIndex ->
-      val visibleItemCount = localLastVisibleItemIndex - localFirstVisibleItemIndex
+    onViewportChanged = { localFirstPagedItemIndex, localLastPagedItemIndex ->
+      val visibleItemCount = localLastPagedItemIndex - localFirstPagedItemIndex
       val proposedPlaceholderPoolSize = visibleItemCount + visibleItemCount / 2
       // We only ever want to increase the pool size.
       if (placeholderPoolSize < proposedPlaceholderPoolSize) {
         placeholderPoolSize = proposedPlaceholderPoolSize
       }
-      state.firstVisibleItemIndex = localFirstVisibleItemIndex
-      lastVisibleItemIndex = localLastVisibleItemIndex
+      state.firstVisibleItemIndex = localFirstPagedItemIndex
+      lastPagedItemIndex = localLastPagedItemIndex
     },
     refreshing = refreshing,
     onRefresh = onRefresh,
--- a/redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
+++ b/redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
@@ -19,9 +19,9 @@ public abstract class WindowedLazyList<W : Any>(
   private val listUpdateCallback: ListUpdateCallback,
 ) : LazyList<W> {
 
-  private var firstVisibleItemIndex = 0
-  private var lastVisibleItemIndex = 0
-  private var onViewportChanged: ((firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) -> Unit)? = null
+  private var firstPagedItemIndex = 0
+  private var lastPagedItemIndex = 0
+  private var onViewportChanged: ((firstPagedItemIndex: Int, lastPagedItemIndex: Int) -> Unit)? = null
 
   final override val items: WindowedChildren<W> = WindowedChildren(listUpdateCallback)
 
@@ -29,11 +29,17 @@ public abstract class WindowedLazyList<W : Any>(
     this.onViewportChanged = onViewportChanged
   }
 
-  protected fun updateViewport(firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) {
-    if (firstVisibleItemIndex != this.firstVisibleItemIndex || lastVisibleItemIndex != this.lastVisibleItemIndex) {
-      this.firstVisibleItemIndex = firstVisibleItemIndex
-      this.lastVisibleItemIndex = lastVisibleItemIndex
-      onViewportChanged?.invoke(firstVisibleItemIndex, lastVisibleItemIndex)
+  public fun updateViewport(firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) {
+    // Paginate the results so `onViewportChanged` isn't blasted with a bunch of updates. This is
+    // particularly important when using `LazyList` in Treehouse (compared to plain Redwood), as
+    // the serialization cost between the host-guest bridge can be expensive.
+    val viewportItemCount = lastVisibleItemIndex - firstVisibleItemIndex + 1
+    val firstPagedItemIndex = (((firstVisibleItemIndex / viewportItemCount) * viewportItemCount) - viewportItemCount).coerceAtLeast(0)
+    val lastPagedItemIndex = (((lastVisibleItemIndex / viewportItemCount) * viewportItemCount) + viewportItemCount * 2).coerceAtMost(items.items.size)
+    if (firstPagedItemIndex != this.firstPagedItemIndex || lastPagedItemIndex != this.lastPagedItemIndex) {
+      this.firstPagedItemIndex = firstPagedItemIndex
+      this.lastPagedItemIndex = lastPagedItemIndex
+      onViewportChanged?.invoke(firstPagedItemIndex, lastPagedItemIndex)
     }
   }
 
--- a/redwood-lazylayout-widget/src/commonTest/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyListTest.kt
+++ b/redwood-lazylayout-widget/src/commonTest/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyListTest.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.lazylayout.widget
+
+import app.cash.redwood.Modifier
+import app.cash.redwood.layout.api.Constraint
+import app.cash.redwood.layout.api.CrossAxisAlignment
+import app.cash.redwood.lazylayout.api.ScrollItemIndex
+import app.cash.redwood.ui.Margin
+import app.cash.redwood.widget.MutableListChildren
+import app.cash.redwood.widget.Widget
+import app.cash.turbine.test
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import kotlin.test.Test
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.runTest
+
+class WindowedLazyListTest {
+  @Test fun updateViewport_0() = updateViewport(40..49, 30..60)
+  @Test fun updateViewport_1() = updateViewport(41..50, 30..70)
+  @Test fun updateViewport_2() = updateViewport(42..51, 30..70)
+  @Test fun updateViewport_3() = updateViewport(49..58, 30..70)
+  @Test fun updateViewport_4() = updateViewport(59..68, 40..80)
+
+  private fun updateViewport(
+    visibleItemIndexRange: IntRange,
+    expectedViewportChange: IntRange,
+  ) = runTest {
+    val windowedLazyList = FakeWindowedLazyList(this, "foo")
+    repeat(100) {
+      windowedLazyList.items.insert(
+        it,
+        object : Widget<String> {
+          override val value: String = it.toString()
+          override var modifier: Modifier = Modifier
+        },
+      )
+    }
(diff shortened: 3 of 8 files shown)


redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
```diff
@@ -30,10 +30,16 @@ public abstract class WindowedLazyList<W : Any>(
...
 
   protected fun updateViewport(firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) {
-    if (firstVisibleItemIndex != this.firstVisibleItemIndex || lastVisibleItemIndex != this.lastVisibleItemIndex) {
-      this.firstVisibleItemIndex = firstVisibleItemIndex
-      this.lastVisibleItemIndex = lastVisibleItemIndex
-      onViewportChanged?.invoke(firstVisibleItemIndex, lastVisibleItemIndex)
+    // Paginate the results so `onViewportChanged` isn't blasted with a bunch of updates. This is
+    // particularly important when using `LazyList` in Treehouse (compared to plain Redwood), as
+    // the serialization cost between the host-guest bridge can be expensive.
+    val viewportItemCount = lastVisibleItemIndex - firstVisibleItemIndex
```
JakeWharton:
```suggestion
    val viewportItemCount = lastVisibleItemIndex - firstVisibleItemIndex + 1
```
If you're showing 0..1 the visible count is 2

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
```diff
@@ -30,10 +30,16 @@ public abstract class WindowedLazyList<W : Any>(
...
+    val firstPagedItemIndex = (((firstVisibleItemIndex / viewportItemCount) * viewportItemCount) - viewportItemCount).coerceAtLeast(0)
+    val lastPagedItemIndex = ((lastVisibleItemIndex / viewportItemCount) * viewportItemCount) + viewportItemCount
```
JakeWharton: Now that we have logic in here, can we write tests for this class?

I'm specifically worried that this algorithm is still a bit off in terms of what it requests. Today, if we're showing 19..29 the count would be 10, the first paged index would be 0 (19/10=1, 1\*10=10, 10-10=0) and the last paged index would be 30 (29/10=2, 2*10=20, 20+10=30). This means moving down the list will cause placeholders to be visible since there's only a single item offscreen in that direction, whereas there's 19 offscreen items in the other direction. A quick fix would be to `+1` after the last index division.

JakeWharton (approved):

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
veyndan: Good point. I just added a basic test case for this class to ensure that it's working. As you described, the logic was slightly off, so it's been since updated.

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
```diff
@@ -19,21 +19,27 @@ public abstract class WindowedLazyList<W : Any>(
...
-      this.firstVisibleItemIndex = firstVisibleItemIndex
-      this.lastVisibleItemIndex = lastVisibleItemIndex
-      onViewportChanged?.invoke(firstVisibleItemIndex, lastVisibleItemIndex)
+  public fun updateViewport(firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) {
+    // Paginate the results so `onViewportChanged` isn't blasted with a bunch of updates. This is
+    // particularly important when using `LazyList` in Treehouse (compared to plain Redwood), as
+    // the serialization cost between the host-guest bridge can be expensive.
+    val viewportItemCount = lastVisibleItemIndex - firstVisibleItemIndex + 1
+    val firstPagedItemIndex = (((firstVisibleItemIndex / viewportItemCount) * viewportItemCount) - viewportItemCount).coerceAtLeast(0)
+    val lastPagedItemIndex = ((lastVisibleItemIndex / viewportItemCount) * viewportItemCount) + viewportItemCount * 2
```
JakeWharton: Does this need a `coerceAtMost` of the list size? Or do we not know the upper bound?

veyndan: We know the upper bound that the `LazyList` is aware of, though that upper bound can dynamically increase (e.g., in the case of Paging). I don't think we necessarily _need_ `coerceAtMost` (or even `coerceAtLeast`), but it'll marginally reduce the number of unnecessary `onViewportChanged` lambda invocations for small lists.

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
```diff
@@ -19,21 +19,27 @@ public abstract class WindowedLazyList<W : Any>(
...
-    if (firstVisibleItemIndex != this.firstVisibleItemIndex || lastVisibleItemIndex != this.lastVisibleItemIndex) {
-      this.firstVisibleItemIndex = firstVisibleItemIndex
-      this.lastVisibleItemIndex = lastVisibleItemIndex
-      onViewportChanged?.invoke(firstVisibleItemIndex, lastVisibleItemIndex)
+  public fun updateViewport(firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) {
+    // Paginate the results so `onViewportChanged` isn't blasted with a bunch of updates. This is
+    // particularly important when using `LazyList` in Treehouse (compared to plain Redwood), as
+    // the serialization cost between the host-guest bridge can be expensive.
+    val viewportItemCount = lastVisibleItemIndex - firstVisibleItemIndex + 1
+    val firstPagedItemIndex = (((firstVisibleItemIndex / viewportItemCount) * viewportItemCount) - viewportItemCount).coerceAtLeast(0)
```
swankjesse, line 37: Why subtract a page?

```suggestion
    val firstPagedItemIndex = (firstVisibleItemIndex / viewportItemCount) * viewportItemCount
```

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
```diff
@@ -19,21 +19,27 @@ public abstract class WindowedLazyList<W : Any>(
...
+    val lastPagedItemIndex = (((lastVisibleItemIndex / viewportItemCount) * viewportItemCount) + viewportItemCount * 2).coerceAtMost(items.items.size)
```
swankjesse, line 38: To round up, do this?
```suggestion
    val lastPagedItemIndex = (lastVisibleItemIndex + viewportItemCount - 1) / viewportItemCount * viewportItemCount
```

redwood-lazylayout-widget/src/commonTest/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyListTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+  init {
+    onViewportChanged { firstPagedItemIndex, lastPagedItemIndex ->
+      scope.launch {
+        _viewportChanges.emit(firstPagedItemIndex..lastPagedItemIndex)
+      }
+    }
+  }
+
+  override fun isVertical(isVertical: Boolean) = TODO()
```
swankjesse: Will we ever do these todos? Maybe replace with `error("unexpected call")` ?

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
```diff
@@ -19,21 +19,27 @@ public abstract class WindowedLazyList<W : Any>(
...
-  protected fun updateViewport(firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) {
-    if (firstVisibleItemIndex != this.firstVisibleItemIndex || lastVisibleItemIndex != this.lastVisibleItemIndex) {
-      this.firstVisibleItemIndex = firstVisibleItemIndex
-      this.lastVisibleItemIndex = lastVisibleItemIndex
-      onViewportChanged?.invoke(firstVisibleItemIndex, lastVisibleItemIndex)
+  public fun updateViewport(firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) {
+    // Paginate the results so `onViewportChanged` isn't blasted with a bunch of updates. This is
+    // particularly important when using `LazyList` in Treehouse (compared to plain Redwood), as
+    // the serialization cost between the host-guest bridge can be expensive.
+    val viewportItemCount = lastVisibleItemIndex - firstVisibleItemIndex + 1
```
swankjesse, line 36: Would you consider `coerceAtLeast()` instead of `+ 1` to make sure it's nonzero?

redwood-lazylayout-widget/src/commonTest/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyListTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import kotlin.test.Test
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.runTest
+
+class WindowedLazyListTest {
+  @Test
+  fun updateViewPort() = runTest {
```
swankjesse: test for empty list?

swankjesse: test updates are deduplicated?

swankjesse: test updates that don’t change paging boundaries are deduplicated?

swankjesse (commented):
I wanna nail down exactly what the calls to onViewportChanged() should be, so the guest code has the maximum amount of control over preloading...

I assume that the bounds on the viewport are inclusive, and the two arguments to `onViewportChanged()` are both inclusive. To make the math easier I’m making the visible end index exclusive below.

Suppose the page size is 10. I think my preference is:

 - scrolls to `[40..50)`, we call `onViewportChanged(40, 49)` to show 10 elements.
 - scrolls to `[41..51)`, we call `onViewportChanged(40, 59)` to show 20 elements
 - scrolls to `[42, 52)`, we do nothing
 - scrolls to `[49, 59)`, we do nothing
 - scrolls to `[50, 60)`, we call `onViewportChanged(50, 59)` to show 10 elements

Alternately we could subtract 1 off at the end to adjust back to inclusive/inclusive?

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
veyndan: Replacing `+ 1` with `coerceAtLeast()` would mean that we are no longer calculating the number of items in the viewport. The `+ 1` is to ensure that that is what we're calculating, and the fact that it produces a nonzero is happy happenstance.

redwood-lazylayout-widget/src/commonTest/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyListTest.kt
veyndan: Probably not! Replaced them

veyndan:
I just updated the tests to showcase what the current solution does. In the current case, the "page size" is identical to the "viewport item count". Therefore, we request a page above and below what where we're currently at. With your proposed solution, hard coding a page size of 10 would be okay in most cases, but if you have very large items (think Instagram posts, where you can only fit one full post on a mobile screen), a page size of 10 is **much** too large.

swankjesse:
Sorry, I didn’t mean to imply a hardcoded page size. I just used examples where the page size as always 10.

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
swankjesse: What does the + 1 do? Inclusive to exclusive?

JakeWharton:
I would not let the viewport get within 1 row of needing placeholders. I would think at minimum we should probably keep at least half a viewport's amount of row's offscreen in either direction.

redwood-lazylayout-widget/src/commonTest/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyListTest.kt
```diff
@@ -0,0 +1,115 @@
...
+          override var modifier: Modifier = Modifier
+        },
+      )
+    }
+    windowedLazyList.updateViewport(41, 50)
+    windowedLazyList.updateViewport(41, 50)
+    windowedLazyList.updateViewport(49, 58)
+    windowedLazyList.updateViewport(59, 68)
+    windowedLazyList.viewportChanges.test {
+      assertThat(awaitItem()).isEqualTo(30..70)
```
swankjesse, line 79: Expanding it by one page on either side feels like you’re doing the guest code’s work on the host side. Would you consider changing these to 40..60 and 50..70 ?

Are these inclusive or exclusive? 40..60 seems like it’s 10 elements too large for 41..50 as input

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/WindowedLazyList.kt
veyndan: I'm kinda confused by what you're asking, as we're not dealing with a range here. As Jake pointed (https://github.com/cashapp/redwood/pull/1471#discussion_r1321731027), we need to do a `+ 1` to ensure we're calculating the number of items in the viewport.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
```diff
@@ -45,24 +45,24 @@ internal fun LazyList(
...
+    onViewportChanged = { localFirstPagedItemIndex, localLastPagedItemIndex ->
+      val visibleItemCount = localLastPagedItemIndex - localFirstPagedItemIndex
       val proposedPlaceholderPoolSize = visibleItemCount + visibleItemCount / 2
       // We only ever want to increase the pool size.
       if (placeholderPoolSize < proposedPlaceholderPoolSize) {
         placeholderPoolSize = proposedPlaceholderPoolSize
       }
-      state.firstVisibleItemIndex = localFirstVisibleItemIndex
-      lastVisibleItemIndex = localLastVisibleItemIndex
+      state.firstVisibleItemIndex = localFirstPagedItemIndex
```
veyndan, line 64: @swankjesse @JakeWharton I didn't notice it before, but such a paginated viewport introduces a bug here. This PR means we can no longer reliably expose to the caller of `LazyList` what the first visible item index is.

JakeWharton: Can we add a third parameter to the lambda for the actual first index?

veyndan: If we did that and didn't update the frequency of the lambda invocation on every scroll, then the first visible item index would still be incorrect due to the new deduplication logic. If the new deduplication logic took into account the actual first index, then we're serializing and deserializing the viewport updates at the same frequency as before.

veyndan:
Closing because of https://github.com/cashapp/redwood/pull/1471#discussion_r1324251565.

PR #1479, opened 2023-09-13
Simplify `LazyListValue` construction in test cases

veyndan:
I want to add more tests to this class, and I don't want to have to keep copy-pasting the construction of `LazyListValue` when 90% of it will always be the same.

### DIFF
--- a/redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
+++ b/redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
@@ -20,6 +20,7 @@ import app.cash.redwood.layout.api.Constraint
 import app.cash.redwood.layout.api.CrossAxisAlignment
 import app.cash.redwood.lazylayout.api.ScrollItemIndex
 import app.cash.redwood.lazylayout.widget.LazyListValue
+import app.cash.redwood.testing.WidgetValue
 import app.cash.redwood.ui.Margin
 import app.cash.redwood.ui.dp
 import assertk.assertThat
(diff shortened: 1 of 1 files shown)


redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
```diff
@@ -78,20 +68,53 @@ class LazyListTest {
...
+  width = Constraint.Wrap,
+  height = Constraint.Wrap,
+  margin = Margin(0.0.dp),
+  crossAxisAlignment = CrossAxisAlignment.Start,
+  scrollItemIndex = ScrollItemIndex(0, 0),
+  placeholder = emptyList(),
+  items = emptyList(),
+)
+
+private fun LazyListValue.copy(
```
veyndan, line 106: Ideally, this'd be generated within `LazyListValue`.

veyndan: I'm guessing we don't want `copy` to be generated, as it'd need to be part of the public API which [we don't want](https://github.com/cashapp/redwood/issues/1043). Though these `WidgetValue` instances are so far exclusively used in tests, so maybe that doesn't matter?

JakeWharton (approved):

redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
JakeWharton: Test API definitely has more flexibility, although breaking it could still break people.

PR #1485, opened 2023-09-14
Test scrolling of `LazyList`

veyndan:
Integrates `WindowedLazyList` in the test, so any change to `WindowedLazyList#updateViewport` will be tested here.

### DIFF
--- a/redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
+++ b/redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
@@ -16,19 +16,31 @@
 package app.cash.redwood.lazylayout.compose
 
 import app.cash.redwood.Modifier
+import app.cash.redwood.RedwoodCodegenApi
 import app.cash.redwood.layout.api.Constraint
 import app.cash.redwood.layout.api.CrossAxisAlignment
+import app.cash.redwood.layout.widget.RedwoodLayoutTestingWidgetFactory
 import app.cash.redwood.lazylayout.api.ScrollItemIndex
 import app.cash.redwood.lazylayout.widget.LazyListValue
+import app.cash.redwood.lazylayout.widget.ListUpdateCallback
+import app.cash.redwood.lazylayout.widget.RedwoodLazyLayoutTestingWidgetFactory
+import app.cash.redwood.lazylayout.widget.RedwoodLazyLayoutWidgetFactory
+import app.cash.redwood.lazylayout.widget.WindowedLazyList
+import app.cash.redwood.testing.TestRedwoodComposition
 import app.cash.redwood.testing.WidgetValue
 import app.cash.redwood.ui.Margin
+import app.cash.redwood.ui.UiConfiguration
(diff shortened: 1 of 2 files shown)


redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
```diff
@@ -86,6 +98,136 @@ class LazyListTest {
...
+
+@OptIn(RedwoodCodegenApi::class)
+private suspend fun <R> TestSchemaTester(
+  factories: TestSchemaWidgetFactories<WidgetValue> = TestSchemaWidgetFactories(
+    TestSchema = TestSchemaTestingWidgetFactory(),
+    RedwoodLayout = RedwoodLayoutTestingWidgetFactory(),
+    RedwoodLazyLayout = RedwoodLazyLayoutTestingWidgetFactory(),
+  ),
+  body: suspend TestRedwoodComposition<List<WidgetValue>>.() -> R,
+): R =
```
JakeWharton, line 144: Isn't this supposed to be generated for the testing code automatically?

JakeWharton (approved):

redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
veyndan: I commented on this but I forgot to press submit 🤦  This is what I originally wrote about it:

> This is a copy-paste of the generated `TestSchemaTester`, with the only change being that factories is now a parameter, thus can be changed. Would such an API change make sense within the generated code?

For reference, this is what the generated one looks like:

```kotlin
@OptIn(RedwoodCodegenApi::class)
public suspend fun <R>
    TestSchemaTester(body: suspend TestRedwoodComposition<List<WidgetValue>>.() -> R): R =
    coroutineScope {
  val factories = TestSchemaWidgetFactories(
    TestSchema = TestSchemaTestingWidgetFactory(),
    RedwoodLayout = RedwoodLayoutTestingWidgetFactory(),
    RedwoodLazyLayout = RedwoodLazyLayoutTestingWidgetFactory(),
  )
  val container = MutableListChildren<WidgetValue>()
  val tester = TestRedwoodComposition(this, factories, container,
      MutableStateFlow(UiConfiguration())) {
    container.map { it.value }
  }
  try {
    tester.body()
  } finally {
    tester.cancel()
  }
}
```

JakeWharton: Yeah I'll think about this tomorrow

veyndan: I'm gonna merge this for now. I'm happy to create a follow-up PR that updates the generation of `TestSchemaTester` such that this custom `TestSchemaTester` can be deleted, either by updating it as it is here, or in a different way that'd allow custom factory implementations to be passed in.

Issue #324, opened 2022-08-25
BackHandler Composable


swankjesse:
Lets hook up something like Compose UI’s BackHandler.

```
@Composable
fun BackHandler(enabled: Boolean = true, onBack: () -> Unit) {
}
```

swankjesse:
```
interface FrameContext : ZiplineService {
  var backHandler: BackHandler?
}

interface BackHandler : ZiplineService {
  fun onBack()
}
```

JakeWharton:
In Compose:
```kotlin
BackHandler(true /* or state derived */) {
  // must handle
}
```

JakeWharton:
I've since come around to this being supported at the Redwood layer because it relies on things like #1116 which previously did not exist as a concept.

PR #1487, opened 2023-09-14
Add simple navigation to test app

JakeWharton:
For some reason iOS is broken and never receives the click within the JS. Will look into separately.

Screenshot_20230914_131019
Screenshot_20230914_131027
Screen Shot 2023-09-14 at 1 14 40 PM
Screen Shot 2023-09-14 at 1 14 43 PM

### DIFF
--- a/test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
+++ b/test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.redwood.testing.presenter
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import app.cash.redwood.Modifier
+import app.cash.redwood.layout.api.Constraint
+import app.cash.redwood.layout.api.Constraint.Companion
+import app.cash.redwood.layout.api.CrossAxisAlignment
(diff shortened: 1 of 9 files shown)


JakeWharton:
Would like to eventually bind a toolbar. Also a great candidate for hooking up #324

test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
```diff
@@ -0,0 +1,79 @@
...
+      width = Constraint.Fill,
+      height = Constraint.Fill,
+    ) {
+      Button("Back", onClick = { screen.value = null })
+      activeScreen.Show(httpClient, modifier = Modifier.grow(1.0))
+    }
+  }
+}
+
+@Suppress("unused") // Used via reflection.
```
veyndan, line 50: What is supposedly unused? When deleting the annotation, I'm not seeing an unused warning (either in the IDE or the CLI).

test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
```diff
@@ -0,0 +1,79 @@
...
+import app.cash.redwood.layout.api.Overflow
+import app.cash.redwood.layout.compose.Column
+import app.cash.redwood.ui.Margin
+import app.cash.redwood.ui.dp
+import com.example.redwood.testing.compose.Button
+import com.example.redwood.testing.compose.Text
+
+@Composable
+fun TestApp(httpClient: HttpClient) {
+  val screen = remember { mutableStateOf<Screen?>(Screen.RepoSearch) }
```
veyndan: Should we default to showing the `HomeScreen`?

```suggestion
  val screen = remember { mutableStateOf<Screen?>(null) }
```

veyndan (approved):
Exciting!

test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
JakeWharton: The enum values were marked in the IDE

test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
JakeWharton: Yes. Was testing if this fixed iOS. It didn't.

test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
JakeWharton: Probably wasn't because I was referring to RepoSearch as the default value. But with the switch to null they'll be unused directly now.

Issue #1384, opened 2023-08-02
Allow arbitrarily serializable content within `rememberSaveable`


veyndan:
Currently, `rememberSaveable` only allows a `MutableState<T>`, where `T` is one of the `JsonPrimitive` (`String`, `Boolean`, `Number`), `List<JsonPrimitive>`, or `Map<JsonPrimitive, JsonPrimitive>`. If you want to serialize anything beyond that, we currently need to add explicit support for it in `StateSnapshot.kt`, and if it's a custom type, you're out of luck.

We should instead use the serializers provided by the user via their `Json` instance, so that the user can serialize anything they can write a serializer for.

```kotlin
Json {
  serializersModule = SerializersModule {
    contextual(CustomSerializer)
  }
}

object CustomSerializer : KSerializer<Custom> { … }

class Custom(val value: Int) { … }

rememberSaveable { Custom(42) }
```

jingwei99:
> use the serializers provided by the user via their Json instance

This can be a good option. I think we need to talk over about the implications of this option (read: to serialize the whole Activity list feels pretty darn daunting for me at this moment, but maybe there are things I don't know so I'm over reacting)

veyndan:
> read: to serialize the whole Activity list feels pretty darn daunting for me at this moment

I don't want to serialize the whole Activity list! I essentially want to implement a `Saver` similar to how [Compose UI saves scroll position](https://github.com/androidx/androidx/blob/a733905d282ecdba574bc5e35d6b0ebf83c82dcd/compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/lazy/LazyListState.kt#L410-L421), which is serializing a couple of integers surrounded by a wrapper class.

jingwei99:
oh that works!

PR #1494, opened 2023-09-19
Preserve scroll position on Android Activity restoration

veyndan:
This doesn't work yet, as `rememberSaveable` needs to support saving of values that _aren't_ `MutableState`'s. Currently, when trying to save the state, we get the following error:

```
app.cash.zipline.ZiplineException: IllegalStateException: unexpected type: {-exvejj=[MutableState(value=medal)@632700585], udeaau=[0]}
                 	at captureStack (runtime/coreRuntime.kt:12)
                 	at IllegalStateException_init_$Create$_0 (kotlin-kotlin-stdlib-js-ir.js)
                 	at toStateSnapshot (commonMainSources/libraries/stdlib/src/kotlin/collections/Maps.kt:84)
                 	at <anonymous> (redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt:114)
                 	at <anonymous> (redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineTreehouseUi.kt)
                 	…
```

One trade-off here is that we're just saving the first visible item index, and thus we only restore the first visible item index when we come back. What this means is that if we scroll past 4 items, and the top of the `LazyList` has scrolled half of the fifth item, when we restore the scroll position, we will show the whole of the fifth item instead of half of it. I think this is perfectly okay for now, as the gains we get from restoring the scroll index is ginormous.

### DIFF
--- a/redwood-lazylayout-compose/build.gradle
+++ b/redwood-lazylayout-compose/build.gradle
@@ -13,6 +13,7 @@ kotlin {
   sourceSets {
     commonMain {
       dependencies {
+        api libs.jetbrains.compose.runtime.saveable
         api projects.redwoodLazylayoutWidget
       }
     }
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyListState.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyListState.kt
@@ -18,14 +18,15 @@ package app.cash.redwood.lazylayout.compose
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.Saver
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 
 @Composable
 public fun rememberLazyListState(
   initialFirstVisibleItemIndex: Int = 0,
 ): LazyListState {
-  return remember {
+  return rememberSaveable(saver = LazyListState.Saver) {
     LazyListState(
       initialFirstVisibleItemIndex,
     )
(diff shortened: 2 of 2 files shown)


jingwei99:
this PR right now is pending on `StateSnapshot` being able to save types (values) that's not `MutableState`. last time @veyndan and I talked on this (it was a month or so ago? maybe more), Veyndan brought an idea about handling serialization for types in `StateSnapshot` (it's just I forgot the details of the idea 🤦🏻‍♀️), I'm talking with Veyndan to get that part done, and then we can land this PR to make the scroll position work!

veyndan:
The idea we were discussing is summarised in https://github.com/cashapp/redwood/issues/1384. We can either go that route of making arbitrary content serializable, or we can just add serialization support for non-`MutableState` values, and implement that issue another day!

redwood-lazylayout-compose/build.gradle
```diff
@@ -13,6 +13,7 @@ kotlin {
   sourceSets {
     commonMain {
       dependencies {
+        implementation libs.jetbrains.compose.runtime.saveable
```
JakeWharton:
```suggestion
        api libs.jetbrains.compose.runtime.saveable
```
This is exposed in public API

jingwei99:
I tested this change locally, with [the change for view id](https://github.com/cashapp/redwood/commit/57f78b982e1a6f7411b4ee0f1e62157ca34491c2), now scroll position is saved!! ✨

jingwei99 (approved):

veyndan:
@jingwei99 Did you test this with the Activity screen, or one of the sample apps? I'm just asking because I tried testing it on Emoji Search and it isn't working

jingwei99:
> @jingwei99 Did you test this with the Activity screen, or one of the sample apps? I'm just asking because I tried testing it on Emoji Search and it isn't working

I was testing it on the sample app, and it worked. 

🤔 maybe I didn't have the `Don't keep activities` setting on? (I think that's what it is, my emulator has been weird, and I've been wiping its data often)

veyndan:
Superseded by https://github.com/cashapp/redwood/pull/1599

PR #1502, opened 2023-09-20
Update CI to JDK 21

JakeWharton:


JakeWharton:
We can fix this, but it's better to not have to specify the Kotlin version on these test projects. In Kotlin 1.9.20 it will have JDK 21 support, but it also will change to select the newest possible version it knows about rather than the absolute latest. So when JDK 22 comes out it will still target 21 until some future Kotlin version adds 22 support.

Thus, just going to sit on this until we land 1.9.20.

swankjesse (approved):

PR #1505, opened 2023-09-21
Remove the statement from past to reflect current state for the project

jingwei99:


swankjesse (approved):

swankjesse:
The `0.x` version number says everything we want to say about stability, without scaring people off.

PR #1507, opened 2023-09-22
fix ButtonBinding.swift click action error！

### LINKED ISSUE #1506: test-app ios project ButtonBinding logic error
Button interface onClick(onClick: (() -> Void)? = nil)
root action #selector  is  clicked() call interface method onClick，should is property onClick call

hxxyyangyong:
The interface method of Button is implemented in ButtonBinding, but the root button executes the interface method again, resulting in the inability to call back the Unit in Kotlin. Here, the assigned attribute OnClick should be called.

Closes #1506

### DIFF
--- a/test-app/ios-uikit/TestApp/ButtonBinding.swift
+++ b/test-app/ios-uikit/TestApp/ButtonBinding.swift
@@ -48,7 +48,7 @@ class ButtonBinding: Button {
     
     @objc func clicked() {
         if (self.onClick != nil) {
-            self.onClick()
+            self.onClick?()
         }
     }
 }
--- a/test-app/ios-uikit/TestApp/ButtonBinding.swift
+++ b/test-app/ios-uikit/TestApp/ButtonBinding.swift
@@ -47,8 +47,6 @@ class ButtonBinding: Button {
     }
     
     @objc func clicked() {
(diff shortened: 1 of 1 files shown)


hxxyyangyong:
@JakeWharton

test-app/ios-uikit/TestApp/ButtonBinding.swift
```diff
@@ -48,7 +48,7 @@ class ButtonBinding: Button {
     
     @objc func clicked() {
         if (self.onClick != nil) {
-            self.onClick()
+            self.onClick?()
```
hoc081098: I think `self.onClick?()` without `if` is better 😂

hxxyyangyong: This has nothing to do with the problem itself. Prioritize solving the actual problem. The main reason is that the calling method is wrong. Adding if is more standard.

JakeWharton: But why would we execute the `if` when adding the conditional chaining operator also performs a conditional? Aren't the two equivalent thus redundant?

JakeWharton (approved):

PR #1509, opened 2023-09-24
Support RTL layout direction in redwood-layout-dom

### LINKED ISSUE #1244: Support RTL layout direction in `redwood-layout-dom`
`Margin#start` and `Margin#end` are mapped to `margin-left` and `margin-right`, instead of `margin-inline-start` and `margin-inline-end` respectively.

https://github.com/cashapp/redwood/blob/afe1c9f5f95eec3cff46837a4b2749cbaf72af8b/redwood-layout-dom/src/main/kotlin/app/cash/redwood/layout/dom/HTMLElementRedwoodLayoutWidgetFactory.kt#L79-L80

ahmed3elshaer:
### Changes 
- `marginInlineStart` and `marginInlineEnd` as extension properties for `CSSStyleDeclaration` that represents ""margin-inline-start" and ""margin-inline-end"
- Replaced `marginLeft` and `marginRight` with `marginInlineStart` and `marginInlineEnd` respectively 

Fixes #1244

### DIFF
--- a/redwood-layout-dom/src/commonMain/kotlin/app/cash/redwood/layout/dom/HTMLElementRedwoodLayoutWidgetFactory.kt
+++ b/redwood-layout-dom/src/commonMain/kotlin/app/cash/redwood/layout/dom/HTMLElementRedwoodLayoutWidgetFactory.kt
@@ -77,8 +77,8 @@ private class HTMLFlexContainer(
 
   override fun margin(margin: Margin) {
     value.style.apply {
-      marginLeft = margin.start.toPxString()
-      marginRight = margin.end.toPxString()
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):
LGTM.

I don't quite understand how to choose between `margin-inline-start` and `margin-block-start` though. Since we don't have any DOM customers I'm inclined to just merge and figure out later. What do you think?

veyndan (approved):
Thanks! This works well if the direction is set to RTL. It'll change the positioning vertically when changing to certain writing modes, but I'm not sure whether that's desired or not. Agreed that we should just merge it and see later!

PR #1512, opened 2023-09-25
Stick to the safe area when rendering test app

JakeWharton:
Did I do any of this right? It looks right...

Screenshot 2023-09-25 at 4 32 18 PM

### DIFF
--- a/test-app/ios-uikit/TestApp/TestAppViewController.swift
+++ b/test-app/ios-uikit/TestApp/TestAppViewController.swift
@@ -30,9 +30,7 @@ class TestAppViewController : UIViewController {
         super.viewDidLoad()
 
         view.backgroundColor = .white
-    }
 
-    override func loadView() {
         let testAppLauncher = TestAppLauncher(nsurlSession: urlSession, hostApi: IosHostApi())
         let treehouseApp = testAppLauncher.createTreehouseApp()
         let widgetSystem = TestSchemaWidgetSystem()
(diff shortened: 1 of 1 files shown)


test-app/ios-uikit/TestApp/TestAppViewController.swift
```diff
@@ -42,7 +40,15 @@ class TestAppViewController : UIViewController {
             codeListener: CodeListener()
         )
         ExposedKt.bindWhenReady(content: content, view: treehouseView)
-        view = treehouseView.view
+
+        let tv = treehouseView.view
+        tv.translatesAutoresizingMaskIntoConstraints = false
+
+        self.view.addSubview(tv)
+        let safeGuide = self.view.safeAreaLayoutGuide
```
dnagler, line 48: I think instead of all of this, you could maybe just do
```
tv.bounds = view.safeAreaLayoutGuide.layoutFrame
```

JakeWharton: Oh sweet. Will try.

dnagler: If this works, it's worth pointing out that it's a bit backwards/unconventional--a `UIView`'s `bounds` rect is expressed relative to the view's own coordinate system, whereas its `frame` rect is expressed relative to its `superview`'s coordinate system. It's more typical to declare `mySubview.frame = mySuperview.bounds`, to get `mySubview` to fill `mySuperview`'s rect.

In this case, `view.safeAreaLayoutGuide.layoutFrame` is expressed `view`'s coordinate system, not `tv`'s.

JakeWharton: Looks like they're not exactly in the same coordinate system and the bound origin or something gets skewed by half the height/width. Even when I fix it (or try to) it occupies the whole screen still rather than only the safe area.

Screenshot 2023-09-25 at 8 14 55 PM

JakeWharton: Going to merge as-is. Hopefully soon we'll have a scaffolding-like composable in this app which gets us a traditional top/nav bar on Android and iOS (and will allow drawing edge-to-edge again).

dnagler: Good to know--it was worth a shot. Definitely safe to merge as is!

dnagler (approved):

PR #1520, opened 2023-09-27
Don't crash running old guest code on new hosts

swankjesse:


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
```diff
@@ -388,35 +389,49 @@ private class ViewContentCodeBinding<A : AppService>(
...
-                cancel()
-              }
+      try {
+        treehouseUi.start(
+          changesSink = this@ViewContentCodeBinding,
+          onBackPressedDispatcher = onBackPressedDispatcherService,
+          uiConfigurations = uiConfigurationFlow,
+          stateSnapshot = restoredState,
+        )
+      } catch (e: ZiplineApiMismatchException) {
```
swankjesse, line 399: This is the new thing

JakeWharton: How should we keep track of when we can delete things like this?

JakeWharton (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
swankjesse: Very good question. How about we borrow an idea from Renovate and do a GitHub tracking issue with all of our in-flight deprecations?
https://github.com/cashapp/redwood/issues/1539

Issue #1518, opened 2023-09-27
Test app needs to save screen location across recreation


JakeWharton:
So a config change puts you back onto the same screen rather than the home screen.

ahmed3elshaer:
For some reason using `rememberSaveable` doesn't work properly.
Inside the `TestApp`
I replaced 
```kotlin
  val screenKeyState = remember { mutableStateOf<String?>(null) }
```
with 
```kotlin
  val screenKeyState = rememberSaveable { mutableStateOf<String?>(null) }
```
And still when navigating to `Repo Search` screen and change rotation i get back to `HomeScreen`

For the sake of debugging i provided the `Saver` and added some print lines
```kotlin
  val screenKeyState = rememberSaveable(
    saver = Saver(
      save = {
        println("${it.value} screen is saved")
        it
      },
      restore = {
        println("${it.value} screen is restored")
        it
      },
    ),
  ) { mutableStateOf<String?>(null) }

  val screenKey = screenKeyState.value
  println("screenKey: $screenKey")
```
And i noticed while navigating to `Repo Search` the `save` lambda doesn't get called.
here is the Logcat
```
screenKey: null
null screen is saved
screenKey: Repo Search
Rotation=========
screenKey: null
null screen is saved
```
Any thoughts ? 🤔

JakeWharton:
Honestly I don't know anything about how it works. Maybe #1531 or #1533 is the fix?

ahmed3elshaer:
Maybe, Specially #1533 
I will test again when it's merged and let you know

ahmed3elshaer:
I have tested the app and found that it does not preserve the state across configurations.
 I have debugged the issue and found that `onSaveInstanceState` is not invoked in `TreehouseLayout` when navigating to a new screen. Moreover, the app crashed once when rotating the device and here are the logs:
```
09:17:26.097 Parcel             Reading a NULL string not supported here.
09:17:26.098 Parcel             Reading a NULL string not supported here.
09:17:26.099 Parcel             Reading a NULL string not supported here.
09:17:26.100 Parcel             Reading a NULL string not supported here.
09:17:26.100 Parcel             Reading a NULL string not supported here.
09:17:26.101 Parcel             Reading a NULL string not supported here.
09:17:26.102 Parcel             Reading a NULL string not supported here.
09:17:27.666 AndroidRuntime     FATAL EXCEPTION: Treehouse
                                Process: com.example.redwood.testing.android.views, PID: 21992
[...]

PR #1525, opened 2023-09-28
Simplify creation of new screens

JakeWharton:
This also changes from an enum key to a string key so that we can switch to rememberSaveable soon (#1518).

### DIFF
--- a/test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
+++ b/test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
@@ -17,6 +17,7 @@ package com.example.redwood.testing.presenter
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.Stable
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import app.cash.redwood.Modifier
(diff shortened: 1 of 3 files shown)


test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
```diff
@@ -30,62 +31,60 @@ import app.cash.redwood.ui.dp
...
-      // TODO This should be a Box.
-      Column(
-        width = Fill,
-        horizontalAlignment = Stretch,
-        modifier = Modifier.grow(1.0).horizontalAlignment(Stretch),
-      ) {
-        activeScreen.Show(httpClient)
+      val content = screens[screenKey]
+      if (content == null) {
+        Text("No screen found with key '$screenKey'!")
```
JakeWharton, line 59: I thought about automatically falling back to home, but if we get `rememberSaveable`, you rename a key, and the app reloads, I want what happened to be slightly more clear.

JakeWharton:
We get nice screen names now because they're strings.

Screenshot_20230927_170830
Screenshot 2023-09-27 at 5 08 42 PM
Screenshot 2023-09-27 at 5 09 01 PM

test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
```diff
@@ -30,62 +31,60 @@ import app.cash.redwood.ui.dp
...
 
+private val screens = buildMap<String, @Composable TestContext.() -> Unit> {
+  put("Repo Search") { RepoSearch(httpClient) }
+  put("UI Configuration") { UiConfigurationValues() }
+}
+
+@Stable
+class TestContext(
+  val httpClient: HttpClient,
+)
```
JakeWharton, line 42: This will become a context receiver eliminating the need to explicitly pass any values in the lambdas above.

veyndan (approved):

ahmed3elshaer:
@JakeWharton I appreciate your effort in doing this task. I tried to implement #1518  yesterday, but it did not work as expected. I realized that the reason was the enum screen  now.

JakeWharton:
Yeah a string should work much better with `rememberSaveable`, at least as it's written today. Eventually it'll support custom savers where we could have written an adapter for an enum, but a string is easier all-around.

PR #1531, opened 2023-09-28
Fix saving of `rememberSaveable` values

veyndan:
I split the PR into two commits, where each of the commits are fixing a bug in its own right.

To summarise:

- `Saveable` didn't have a serialization adapter, so it'd crash because of that.
- When deserialising a `MutableState`, we didn't deserialise the value within.


JakeWharton (approved):

jingwei99:
🙏

PR #1533, opened 2023-09-28
Set Android View ID on `TreehouseLayout` by default

veyndan:
If one doesn't explicitly set an Android View ID, `onSaveInstanceState` is never invoked, and thus `rememberSaveable` will never work as intended.

### DIFF
--- a/redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseLayout.kt
+++ b/redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseLayout.kt
@@ -25,6 +25,7 @@ import androidx.activity.OnBackPressedDispatcher as AndroidOnBackPressedDispatch
 import app.cash.redwood.treehouse.TreehouseView.ReadyForContentChangeListener
 import app.cash.redwood.treehouse.TreehouseView.WidgetSystem
 import app.cash.redwood.widget.RedwoodLayout
+import app.cash.treehouse.host.R
 import java.util.UUID
 
 @Deprecated(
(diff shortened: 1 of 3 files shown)


jingwei99 (approved):

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseLayout.kt
```diff
@@ -57,6 +57,12 @@ public class TreehouseLayout(
 
   override var saveCallback: TreehouseView.SaveCallback? = null
 
+  init {
+    // The view needs to have an id for Android to populate saved data back
+    @SuppressLint("ResourceType")
+    id = 9000
```
JakeWharton: This is really unsafe. You can put an `<item type="id">treehouse_layout</item>` into resources and use it via `R.id.treehouse_layout` so that the build system guarantees a unique ID value.

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseLayout.kt
```diff
@@ -57,6 +57,12 @@ public class TreehouseLayout(
 
   override var saveCallback: TreehouseView.SaveCallback? = null
 
+  init {
+    // The view needs to have an id for Android to populate saved data back
```
JakeWharton: nit:
```suggestion
    // The view needs to have an ID for Android to populate saved data back.
```
[Comments should be sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html).

JakeWharton (approved):

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/TreehouseLayout.kt
JakeWharton: You will need to turn on resources for this module as they're disabled by default.

https://github.com/cashapp/redwood/blob/89750b170fe3e8987227271dba45353d7bbc83d4/redwood-layout-view/build.gradle#L33-L36

PR #1534, opened 2023-09-28
Hoist property binding up into abstract types

JakeWharton:
No need to reimplement in each subtype.

Also switch over to a JVM library as there's nothing Android-specific in use in the abstract types.


JakeWharton:
The babiest of baby steps toward iOS screenshot tests....

colinrtwhite (approved):

PR #1536, opened 2023-09-28
Save screen location across recreation

### LINKED ISSUE #1518: Test app needs to save screen location across recreation

ahmed3elshaer:
- Changed `remember` to `rememberSavable` to survive configuration changes.

It depends on merging #1533 to be tested.


swankjesse (approved):

jingwei99 (approved):

JakeWharton:
We can merge this because it no longer crashes, at least, but it does not actually fix the problem as discussed on the issue.

Issue #1500, opened 2023-09-20
Dynamically adjust the number of placeholders


veyndan:
`LazyList` passes down a hardcoded number of placeholder instances to each host view ([currently set to 20](https://github.com/cashapp/redwood/blob/c0a1b3dbe733e37a7a356a520d727d2a65b9e5d7/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt#L52)). This should be sufficient for the majority of screens, but for `LazyList`'s that are sufficiently tall or whose row item heights are sufficiently small, there may be cases where 20 placeholders aren't enough.

The `ViewLazyList` for Android already has a mechanism for requesting more placeholders if the existing ones aren't sufficient, as outlined in the following:

https://github.com/cashapp/redwood/blob/c0a1b3dbe733e37a7a356a520d727d2a65b9e5d7/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt#L172-L179

This mechanism should be copied/shared with iOS.

Issue #1498, opened 2023-09-20
On iOS, don't refresh all cells when items are inserted/removed


veyndan:
When any property of `LazyList` changes on iOS, we invoke `collectionView.reloadData()`. This is an expensive operation which shouldn't be necessary to call, as we are given atomic change updates within `UICollectionViewListUpdateCallback`.

Currently, I don't have a lot of confidence in `UICollectionViewListUpdateCallback` being given the correct atomic change updates. This is due to the lack of tests, and the fact that on Android we invoke `notifyDatasetChanged` (the equivalent to iOS' `reloadData`) on a specific code path, further reducing trust that these atomic change updates are correct.

Therefore, I would argue that this is blocked by:

- [x] https://github.com/cashapp/redwood/issues/1497
- [x] https://github.com/cashapp/redwood/issues/1128

swankjesse:
Implemented by https://github.com/cashapp/redwood/pull/1538

PR #1538, opened 2023-09-29
Update UITableView to track changes precisely

swankjesse:
Previously we called reloadData() whenever anything changed in the underlying model.

With this change we fire precise events to the UI when inserts and removes happen in the source LazyList. This attempts to coalesce changes to itemsBefore() and itemsAfter() with adjacent changes and the beginning and end of the items list. To implement this the changes are buffered into an intermediate model before they are applied.

When a cell changes from being a placeholder to a loaded item, or the opposite, this does a cell content change without firing an event through the UITableView. To implement this we must track which cells are currently displaying placeholders. This PR includes a new datastructure, SparseList, to implement this tracking.

This PR also changes the number of placeholders that guest code offers to host code. I found experimentally that 20 placeholders was not enough for UITableView.

This PR changes LazyList from UICollectionView to
UITableView. This change was potentially unnecessary, though UITableView has fewer features that we don't need.


redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
```diff
@@ -0,0 +1,388 @@
...
+      }
+    }
+
+    override fun move(fromIndex: Int, toIndex: Int, count: Int) {
+      edits += Edit.Move(fromIndex, toIndex, count)
+    }
+
+    override fun remove(index: Int, count: Int) {
+      val last = edits.lastOrNull()
+      if (last is Edit.Remove && index in last.index - count until last.index + 1) {
```
swankjesse, line 90: Need to test fencepost logic on these.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -25,133 +25,144 @@ import app.cash.redwood.Modifier
...
-  private var collectionViewFlowLayout: UICollectionViewFlowLayout = UICollectionViewFlowLayout().apply {
-    estimatedItemSize = UICollectionViewFlowLayoutAutomaticSize.readValue()
-  },
-  internal val collectionView: UICollectionView = UICollectionView(
+  internal val tableView: UITableView = UITableView(
     CGRectZero.readValue(),
-    collectionViewFlowLayout,
   ),
-) : WindowedLazyList<UIView>(UICollectionViewListUpdateCallback(collectionView)), ChangeListener {
+) : LazyList<UIView>, ChangeListener {
```
veyndan, line 67: The goal of `WindowedLazyList` was to have `UIViewLazyList` and `ViewLazyList` to share the bulk of their implementations. They are fairly similar (e.g., view based, platform provides a recycling list view that allows atomic updates). The integration of `WindowedLazyList` with `UIViewLazyList` wasn't complete, namely that `UICollectionViewListUpdateCallback` didn't yet delegate to atomic updates.

What are your thoughts on continuing down that path vs switching approach as you did here?

veyndan (changes requested):
Just a drive-by review for now:

> 

> This PR changes LazyList from UICollectionView to UITableView. This change was potentially unnecessary, though UITableView has fewer features that we don't need.

Switching it back to `UITableView` means that `LazyRow` can no longer work (see https://github.com/cashapp/redwood/pull/942#discussion_r1172570297). I'd be very hesitant to break support for this!

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
veyndan: There's this issue from last week that has more info: https://github.com/cashapp/redwood/issues/1498

veyndan:
> This PR also changes the number of placeholders that guest code offers to host code. I found experimentally that 20 placeholders was not enough for UITableView.

The underlying issue is probably https://github.com/cashapp/redwood/issues/1500. Increasing to 30 placeholders increases the number of serialized views, which'll be unnecessary for the majority of lists.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -25,133 +25,144 @@ import app.cash.redwood.Modifier
...
+
+  private val processor = object : LazyListUpdateProcessor<LazyListContainerCell, UIView>() {
+    override fun createCell(
+      cell: Cell<LazyListContainerCell, UIView>,
+      widget: Widget<UIView>,
+      index: Int,
+    ): LazyListContainerCell {
+      val result = tableView.dequeueReusableCellWithIdentifier(
+        identifier = reuseIdentifier,
+        forIndexPath = NSIndexPath.indexPathForItem(index.convert(), 0.convert()),
```
dnagler, line 83: Calling `indexPathForItem` this frequently might be expensive--it might be worth trying to just create our own index paths, and enforcing that the table view only has a single section (so that every index path is initialized with `.init(section: 0, row: index)`

dnagler, line 83: nit: is it possible to use the more modern/bridged `IndexPath` type here?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -25,133 +25,144 @@ import app.cash.redwood.Modifier
...
+  override val items: Widget.Children<UIView> = processor.items
 
-        return cell
-      }
-    }
+  private val dataSource = object : NSObject(), UITableViewDataSourceProtocol {
+    override fun tableView(
+      tableView: UITableView,
+      numberOfRowsInSection: NSInteger,
+    ) = processor.size.toLong()
```
dnagler: nit: You might want to add an assertion here that `section == 0`

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -25,133 +25,144 @@ import app.cash.redwood.Modifier
...
   init {
-    collectionView.apply {
-      dataSource = collectionViewDataSource
-      delegate = collectionViewDelegate
+    tableView.apply {
+      dataSource = this@UIViewLazyList.dataSource
+      delegate = tableViewDelegate
       prefetchingEnabled = true
+      rowHeight = UITableViewAutomaticDimension
+      estimatedRowHeight = 44.0
```
dnagler: nit: is this needed? 44 is the default value since iOS 11

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -25,133 +25,144 @@ import app.cash.redwood.Modifier
...
+    this.onViewportChanged = onViewportChanged
+  }
+
   override fun isVertical(isVertical: Boolean) {
-    collectionViewFlowLayout.scrollDirection = if (isVertical) {
-      UICollectionViewScrollDirectionVertical
-    } else {
-      UICollectionViewScrollDirectionHorizontal
-    }
+    // TODO: support horizontal LazyLists.
```
dnagler, line 167: Do we plan to support this in the future, or can/should we punt this indefinitely? (Has this already been built on the Android side?)

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: This new class started out as a copy-paste of WindowedLazyList, addressing problems on iOS by adding precise updates.

Are precise updates useful for RecyclerView? If precise updates will make RecyclerView faster, then we should follow-up by migrating RecyclerView to LazyListUpdateProcessor (and delete the old WindowedLazyList). If precise updates are not necessary for RecyclerView, then we should not do precise updates; it’s a bit of work to provide them.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -160,69 +171,109 @@ internal open class UIViewLazyList(
...
+    contentView.translatesAutoresizingMaskIntoConstraints = false
+    setNeedsLayout()
   }
 
   override fun layoutSubviews() {
     super.layoutSubviews()
-    widgetView?.setFrame(this.contentView.bounds)
+
+    val widgetView = this.widgetView ?: return
+    widgetView.setFrame(bounds)
```
dnagler, line 268: nit: since `widgetView` is a subview of `contentView`, I would probably do
```
contentView.setFrame(bounds)
widgetView.setFrame(contentView.bounds)
```
instead, but it should have the same effect

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -160,69 +171,109 @@ internal open class UIViewLazyList(
...
-  ): UICollectionViewLayoutAttributes {
-    val itemSize = widgetView!!.sizeThatFits(collectionViewFrame.useContents { size.readValue() })
-    return layoutAttributes.apply {
-      size = if (collectionViewFlowLayout.scrollDirection == UICollectionViewScrollDirectionVertical) {
-        CGSizeMake(collectionViewFrame.useContents { size.width }, itemSize.useContents { height })
-      } else {
-        CGSizeMake(itemSize.useContents { width }, collectionViewFrame.useContents { size.height })
-      }
+  override fun sizeThatFits(size: CValue<CGSize>): CValue<CGSize> {
+    return widgetView?.sizeThatFits(size) ?: return super.sizeThatFits(size)
```
dnagler, line 271: this can probably just return `contentView.sizeThatFits(size)`

swankjesse:
@veyndan ooooh I like that technique to handle lack of placeholders. Right now the app crashes if there aren’t enough placeholders. I’d like to introduce this performance regression to fix that crash, and then later we can implement #1500 as an optimization.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: Reading that issue makes it clear to me that this new thing will benefit both platforms. I’d still like to land it one-platform-at-a-time, mostly because our test coverage here is lacking.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: I’d like to see what happens if I swap out UITableView for UICollectionView. If I’m lucky it’ll just be tracking slight differences in the APIs with no performance impact. I think I’ll be lucky.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: I couldn’t find it in the Kotlin/Native APIs.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: Done.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: we'll also definitely want to call `widgetView.sizeToFit()` here

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: Fixed!

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: After playing with this I’ve learned that contentView is disinterested in delegating to its children for stuff like sizing, so the cell needs to control both. I’m gonna keep it as-is for now ’cause it seems to be working!

dnagler:
After pairing on this (and seeing marked improvements over current performance!!), I'm in favor of landing this as-is and continuing to iterate, especially since this will be a great help to continuing our QA process on the iOS side (right now our QA team is testing a very broken experience).

dnagler (approved):

swankjesse:
@veyndan

I spent a couple days trying to switch this back to UICollectionView and I’ve been unsuccessful thus far. I can’t figure out how to get dynamic heights working without it exploding because it runs out of placeholders.

Could you please release your merge-block on this PR? We can do the UITableView to UICollectionView switch in a follow-up, but I’d like to unblock the stuff we need right now in Cash App.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
```diff
@@ -49,14 +49,14 @@ internal fun LazyList(
...
-  var placeholderPoolSize by remember { mutableStateOf(20) }
+  var placeholderPoolSize by remember { mutableStateOf(30) }
   LazyList(
     isVertical,
     itemsBefore = itemsBefore,
     itemsAfter = itemsAfter,
     onViewportChanged = { localFirstVisibleItemIndex, localLastVisibleItemIndex ->
       val visibleItemCount = localLastVisibleItemIndex - localFirstVisibleItemIndex
-      val proposedPlaceholderPoolSize = visibleItemCount + visibleItemCount / 2
+      val proposedPlaceholderPoolSize = visibleItemCount * 2
```
veyndan: Revert? This just change just affects Android, as iOS doesn't know how to request more placeholders.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
```diff
@@ -49,14 +49,14 @@ internal fun LazyList(
   val itemsBefore = remember(state.firstVisibleItemIndex) { (state.firstVisibleItemIndex - OffscreenItemsBufferCount / 2).coerceAtLeast(0) }
   val itemsAfter = remember(lastVisibleItemIndex, itemProvider.itemCount) { (itemProvider.itemCount - (lastVisibleItemIndex + OffscreenItemsBufferCount / 2).coerceAtMost(itemProvider.itemCount)).coerceAtLeast(0) }
   val scrollItemIndex = remember(state.scrollToItemTriggeredId) { ScrollItemIndex(state.scrollToItemTriggeredId, state.firstVisibleItemIndex) }
-  var placeholderPoolSize by remember { mutableStateOf(20) }
+  var placeholderPoolSize by remember { mutableStateOf(30) }
```
veyndan, line 54: Can we add a `TODO` to drop this down to 20 again? It's really just blocked on iOS being unable to request more placeholders.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
veyndan: That sounds good. Do you mind adding a `TODO` here (or manually create an issue) so that we remember that these two mechanisms should ideally be merged?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
veyndan: @dnagler To answer your question directly:

1. We plan to support this in the future. We'll need this within Cash for things like the QAB, but also, `LazyRow` should be provided for external consumers (i.e., non Cash people), as these components are supposed to be the lazy counterparts to the existing `Column`/`Row` components.
2. Yes, it's already been built on the Android side.

veyndan (approved):
Stamping to unblock

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
swankjesse: Done.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyList.kt
swankjesse: Done.

```
  // TODO(jwilson): drop this down to 20 once this is fixed:
  //     https://github.com/cashapp/redwood/issues/1551
```

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: Here’s a tracking issue: https://github.com/cashapp/redwood/issues/1552

PR #1545, opened 2023-10-02
Add README for TestApp

underscoretang:
Need this selfishly so i can remember the gradle commands. Just cloned emoji-search/README

### DIFF
--- a/test-app/README.md
+++ b/test-app/README.md
@@ -1,0 +1,76 @@
+Test App
+============
+
+This is an app to aid in redwood development. The app itself is a list of smaller apps that can be expanded upon to develop particular use cases.
+
+
+Prerequisites
+-------------
+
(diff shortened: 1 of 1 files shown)


dnagler (approved):
thanks!

test-app/README.md
```diff
@@ -0,0 +1,76 @@
+Test App
+============
+
+This is an app to aid in redwood development. The app itself is a list of smaller apps that can be expanded upon to develop particular use cases.
```
JakeWharton, line 4: I'm not sure we should call these smaller apps since they can't exist in isolation in any way. Maybe smaller demos? Or even sticking to screens as is used in code?

JakeWharton (approved):

test-app/README.md
underscoretang: sorry this automerged. will update this in another one

PR #1563, opened 2023-10-05
Embed the Zipline app in Android app assets

JakeWharton:
This allows the app to function without the server running, or without ever having to had connect to the server in the past. It will also allow automated testing of these applications in the future.

Refs #537

Views:
Screenshot_20231005_123738

Compose UI:
Screenshot_20231005_124117

### DIFF
--- a/build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildExtension.kt
+++ b/build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildExtension.kt
@@ -26,6 +26,16 @@ interface RedwoodBuildExtension {
    */
   fun publishing()
 
-  /** Bundle a zip with dependencies and startup scripts. */
-  fun application(name: String, mainClass: String)
+  /** Bundle a zip with dependencies and startup scripts for a CLI. */
+  fun cliApplication(name: String, mainClass: String)
+
+  /** Bundle a zip of a Zipline application's compiled `.zipline` files. */
+  fun ziplineApplication()
+
+  /**
+   * Consume a Zipline application in an Android application and embed it within assets.
+   *
+   * @name Name of the Treehouse application. Will be used to prefix the Zipline manifest file.
+   */
+  fun embedZiplineApplication(dependencyNotation: Any, name: String)
 }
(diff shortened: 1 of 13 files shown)


build-support/src/main/kotlin/app/cash/redwood/buildsupport/ZiplineAppAssetCopyTask.kt
```diff
@@ -0,0 +1,81 @@
...
+    outputManifestFile.writeText(outputManifest.encodeJson())
+
+    // Rewrite all .zipline files to their SHA-256 hashes which is how Zipline loads when embedded.
+    for (module in outputManifest.modules.values) {
+      val inputFile = checkNotNull(fileMap.remove(module.url)) {
+        "No ${module.url} file found in input files as specified by the manifest"
+      }
+      val outputFile = outputDirectory.file(module.sha256.hex()).asFile
+      inputFile.copyTo(outputFile)
+    }
```
JakeWharton, line 79: Future change: going to move all this to happen on the production-side rather than the consumer-side so as to support iOS more easily.

build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildExtension.kt
```diff
@@ -26,6 +26,16 @@ interface RedwoodBuildExtension {
    */
   fun publishing()
 
-  /** Bundle a zip with dependencies and startup scripts. */
-  fun application(name: String, mainClass: String)
+  /** Bundle a zip with dependencies and startup scripts for a CLI. */
+  fun cliApplication(name: String, mainClass: String)
```
swankjesse, line 30: I really like how this plugin is structured

build-support/src/main/kotlin/app/cash/redwood/buildsupport/RedwoodBuildPlugin.kt
```diff
@@ -374,4 +376,74 @@ private class RedwoodBuildExtensionImpl(private val project: Project) : RedwoodB
       }
     }
   }
+
+  override fun ziplineApplication() {
```
swankjesse, line 380: I aught to move this behavior to Zipline itself!

build-support/src/main/kotlin/app/cash/redwood/buildsupport/ZiplineAppAssetCopyTask.kt
swankjesse: And move this to the Zipline Gradle plugin!

swankjesse (approved):
Love it.

PR #1567, opened 2023-10-06
Serialize `MutableState` as a normal `JsonObject`

veyndan:
We don't need to special case it by wrapping it in a custom `Saveable` object with a boolean flag.

Refs #1384

### DIFF
--- a/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
+++ b/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
@@ -21,23 +21,23 @@ import kotlin.jvm.JvmInline
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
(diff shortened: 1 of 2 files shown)


jingwei99:
did you get to test `remembersaveable` behaviors after this change in the example app?

veyndan:
Yep, I tested Emoji Search and the search field is restored

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
```diff
@@ -74,22 +68,22 @@ private fun Any?.toJsonElement(): JsonElement {
...
+    this is JsonPrimitive -> {
       if (this.isString) return content
       return booleanOrNull ?: doubleOrNull ?: intOrNull ?: error("unexpected type: $this")
       // TODO add other primitive types (float, long) when needed
     }
 
-    is JsonArray -> listOf({ this.forEach { it.toJsonElement() } })
+    this is JsonArray -> listOf({ this.forEach { it.toJsonElement() } })
+    this is JsonObject && containsKey(MutableStateKey) ->
+      mutableStateOf(getValue(MutableStateKey).fromJsonElement())
```
jingwei99, line 80: I see!

jingwei99 (approved):

PR #1569, opened 2023-10-06
Refactor processor to support externally-created views

swankjesse:


swankjesse:
This is intended to be a pure refactoring with no behavior change.

JakeWharton (approved):

PR #1574, opened 2023-10-10
Promote AbstractFlexContainerTest to commonMain

swankjesse:
This requires parameterized tests to be manually exploded into their cartesian product.


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Square, Inc.
+ * Copyright (C) 2023 Square, Inc.
```
swankjesse, line 2: I assume a robot did this? I have no recollection.

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -223,6 +330,11 @@ abstract class AbstractFlexContainerTest<T : Any> {
     container.onEndChanges()
     verifySnapshot(container, "updated")
   }
+
+  /** We don't have assume() on kotlin.test. Tests that fail here should be skipped instead. */
+  private fun assumeTrue(b: Boolean) {
+    assertTrue(b)
+  }
```
JakeWharton, line 337: Delete? And delete call-sites? Presumably you did not create invalid variants when exploding them.

JakeWharton (approved):

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
JakeWharton: If you F6 a file it will update the copyright for some reason.

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
swankjesse: I didn’t, but I want to keep the assume calls around for when we later do burst.

PR #1581, opened 2023-10-12
Share snapshot callback and improve its output

JakeWharton:


redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testColumnThenRow.1.png
JakeWharton, line 1: If no name is provided a counter is implicitly used.

redwood-layout-uiview/RedwoodLayoutUIViewTests/SnapshotTestingCallback.swift
```diff
@@ -0,0 +1,17 @@
+import RedwoodLayoutUIViewTestKt
+import SnapshotTesting
+import UIKit
+
+final class SnapshotTestingCallback : UIViewSnapshotCallback {
+    private let testName: String
+    private let fileName: StaticString
+
+    init(named testName: String, _ fileName: StaticString = #file) {
```
JakeWharton, line 9: I don't quite understand how it works, but this default is evaluated in the context of the caller which is very different from how Kotlin does it. As a result, we get to capture the name of the calling file automatically.

colinrtwhite (approved):

redwood-layout-uiview/RedwoodLayoutUIViewTests/SnapshotTestingCallback.swift
JakeWharton: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/expressions/#Macro-Expansion-Expression

PR #1595, opened 2023-10-15
Using FileStateStore instead of MemoryStore in TestApp

ahmed3elshaer:
As discussed in #1518 `FileStateStore` is required to survive configuration changes in Android.

After testing every screen state is saved as expected 🚀 , thanks to @veyndan 🙏🏼 

Side notes : Sometimes the back button functionality doesn't work in `Repo Search` screen


veyndan (approved):
I think copy-pasting `FileStateStore` from Emoji Search is a good trade-off until we have a better default `StateStore` implementation. Thanks for the PR!

ahmed3elshaer:
@veyndan glad that i can help 🙏🏼

PR #1600, opened 2023-10-17
Make default saver for LazyListState private

jingwei99:


redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyListState.kt
```diff
@@ -72,11 +67,18 @@ public class LazyListState {
     this.firstVisibleItemIndex = firstVisibleItemIndex
   }
 
-  public companion object {
+  @Composable
+  public fun rememberLazyListState(): LazyListState {
+    return rememberSaveable(saver = Saver) {
+      LazyListState()
+    }
+  }
```
jingwei99: this feels a bit off for me?

JakeWharton: You can keep the function top-level, but instead move the `private val Saver` to also be a top-level property. The function then will have access to it, and the `LazyListState` type won't need a companion at all.

jingwei99: done!

JakeWharton (approved):

PR #1603, opened 2023-10-17
Allow arbitrarily serializable content within `rememberSaveable`

### LINKED ISSUE #1384: Allow arbitrarily serializable content within `rememberSaveable`

veyndan:
Closes #1384.

### DIFF
--- a/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
+++ b/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
@@ -18,75 +18,69 @@ package app.cash.redwood.treehouse
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.mutableStateOf
 import kotlin.jvm.JvmInline
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Polymorphic
+import kotlinx.serialization.PolymorphicSerializer
+import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.json.JsonArray
-import kotlinx.serialization.json.JsonElement
-import kotlinx.serialization.json.JsonNull
-import kotlinx.serialization.json.JsonObject
-import kotlinx.serialization.json.JsonPrimitive
-import kotlinx.serialization.json.booleanOrNull
-import kotlinx.serialization.json.buildJsonObject
-import kotlinx.serialization.json.doubleOrNull
-import kotlinx.serialization.json.intOrNull
-
-private const val MutableStateKey = "androidx.compose.runtime.MutableState"
+import kotlinx.serialization.SerializationStrategy
+import kotlinx.serialization.builtins.ListSerializer
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.polymorphic
+import kotlinx.serialization.modules.subclass
 
 @Serializable
 public class StateSnapshot(
-  public val content: Map<String, List<JsonElement>>,
+  public val content: Map<String, List<@Polymorphic Any?>>,
 ) {
-  public fun toValuesMap(): Map<String, List<Any?>> {
-    return content.mapValues { entry ->
-      entry.value.map {
-        it.fromJsonElement()
-      }
-    }
-  }
 
   @JvmInline
   @Serializable
   public value class Id(public val value: String?)
 }
 
-/**
- * Supported types:
- * String, Boolean, Int, List (of supported primitive types), Map (of supported primitive types)
- */
-public fun Map<String, List<Any?>>.toStateSnapshot(): StateSnapshot = StateSnapshot(
-  mapValues { entry -> entry.value.map { element -> element.toJsonElement() } },
-)
-
-private fun Any?.toJsonElement(): JsonElement {
-  return when (this) {
-    is MutableState<*> -> JsonMutableState(value.toJsonElement())
-    is String -> JsonPrimitive(this)
-    is Int -> JsonPrimitive(this)
-    is List<*> -> JsonArray(map { it.toJsonElement() })
-    is JsonElement -> this
-    null -> JsonNull
-    else -> error("unexpected type: $this")
-    // TODO: add support to Map<*, *>
+// TODO Add support for rest of built-ins serializers.
+public val SaveableStateSerializersModule: SerializersModule = SerializersModule {
+  polymorphic(Any::class) {
+    subclass(Boolean::class)
+    subclass(Double::class)
+    subclass(Float::class)
+    subclass(Int::class)
+    subclass(String::class)
   }
-}
-
-private fun JsonElement.fromJsonElement(): Any? {
-  return when {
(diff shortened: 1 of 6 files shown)


samples/emoji-search/launcher/src/commonMain/kotlin/com/example/redwood/emojisearch/launcher/EmojiSearchAppSpec.kt
```diff
@@ -19,13 +19,15 @@ import app.cash.redwood.treehouse.TreehouseApp
...
 import com.example.redwood.emojisearch.treehouse.HostApi
+import com.example.redwood.emojisearch.treehouse.emojiSearchSerializersModule
 import kotlinx.coroutines.flow.Flow
 
 class EmojiSearchAppSpec(
   override val manifestUrl: Flow<String>,
   private val hostApi: HostApi,
 ) : TreehouseApp.Spec<EmojiSearchPresenter>() {
   override val name = "emoji-search"
+  override val serializersModule = emojiSearchSerializersModule
```
veyndan, line 30: It took a while to notice that the `serializersModule` had to be set in _two_ places. I remembered that `Zipline.get()` took a `serializersModule`, but didn't realise that the `TreehouseApp.Spec` also needed one.

The stacktrace was very unhelpful, as it just said that there is no serializer attached for some type. This led to me fiddling with how the serializers are set up, without realising that I needed to attach _another_ `serializersModule` somewhere. I eventually figured it out by doing a global search of `EmptySerializersModule` and found out that I can override the `serializersModule` of `TreehouseApp.Spec`. So I ultimately figured it out by happenstance. 

Is there a way to merge the two ways of setting a `serializersModule`, so we only set the `serializersModule` in a single place? (In the case of Emoji Search, it needed to be set in _three_ places!)

swankjesse: I think the best thing is to define a serializersModule instance somewhere in your app’s commonMain, then to use that shared instance everywhere Treehouse is initialized.

We can’t really share the initialization code as it’s not in commonMain.

veyndan: > I think the best thing is to define a serializersModule instance somewhere in your app’s commonMain, then to use that shared instance everywhere Treehouse is initialized.

Yep, that's what I did with `emojiSearchSerializersModule`. The `serializersModule` is defined once and referenced where needed.

> We can’t really share the initialization code as it’s not in commonMain.

👍

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
```diff
@@ -18,75 +18,69 @@ package app.cash.redwood.treehouse
...
+    @Suppress("UNCHECKED_CAST")
+    when (value) {
+      is List<*> -> ListSerializer(PolymorphicSerializer(Any::class)) as SerializationStrategy<Any>
+      is MutableState<*> -> MutableStateSerializer as SerializationStrategy<Any>
+      else -> null
+    }
+  }
+  polymorphicDefaultDeserializer(Any::class) { className ->
+    when (className) {
+      "kotlin.collections.ArrayList" -> ListSerializer(PolymorphicSerializer(Any::class))
```
swankjesse, line 63: What happens to List subtypes other than ArrayList?

redwood-treehouse/src/commonTest/kotlin/app/cash/redwood/treehouse/StateSnapshotTest.kt
```diff
@@ -18,66 +18,55 @@ package app.cash.redwood.treehouse
...
+        "key3" to listOf(mutableStateOf("str")),
+        "key4" to listOf("str"),
+        "key5" to listOf(null),
+        "key6" to listOf(true),
+        "key7" to listOf(1.0),
+        "key8" to listOf(1.0f),
+        "key9" to listOf(listOf(1, "str")),
+      ),
+    )
+    val serialized = json.encodeToString(stateSnapshot)
```
swankjesse, line 50: Please also assert on the contents of this string? As-is if there’s a backwards-incompatible change in how a string is encoded we won’t get a failing test.

redwood-treehouse/src/commonTest/kotlin/app/cash/redwood/treehouse/StateSnapshotTest.kt
```diff
@@ -18,66 +18,55 @@ package app.cash.redwood.treehouse
...
+  fun stateSnapshotSerializeThenDeserialize() {
+    val json = Json {
+      useArrayPolymorphism = true
+      serializersModule = SaveableStateSerializersModule
+    }
+    val stateSnapshot = StateSnapshot(
+      mapOf(
+        "key1" to listOf(mutableStateOf(1)),
+        "key2" to listOf(1),
+        "key3" to listOf(mutableStateOf("str")),
```
swankjesse, line 42: Do mutable state values support all this polymorphism?

swankjesse: Mind adding a test for that?

swankjesse (approved):

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/StateSnapshot.kt
veyndan: It'll crash! I'm grouping other List subtypes as part of the "Add support for rest of built-ins serializers." TODO comment above this.

redwood-treehouse/src/commonTest/kotlin/app/cash/redwood/treehouse/StateSnapshotTest.kt
veyndan: It does and this is the test for that! The `Int` and `String` values are both resolved polymorphically.

redwood-treehouse/src/commonTest/kotlin/app/cash/redwood/treehouse/StateSnapshotTest.kt
veyndan: For future readers, I tried asserting on `encodeToString`, but the results are different depending on the target its run on (JS gives a different JSON string than the JVM). I'm gonna assert on the `JsonElement` instead.

PR #1606, opened 2023-10-17
Fix crash on scroll position restoration

### LINKED ISSUE #1604: Crash in emoji-search after scroll position preservation commit
I haven't properly reviewed the PR yet, but I'm noticing a crash in Emoji Search after these changes:

- Run Emoji Search using Android Views
- Type "medal" into the search box
- Click on one of the medal emojis
- Press the system back button
- The app crashes with the exception `app.cash.zipline.ZiplineException: IllegalArgumentException: Failed requirement.`

_Originally posted by @veyndan in https://github.com/cashapp/redwood/issues/1599#issuecomment-1766197162_

jingwei99:
Fixes: https://github.com/cashapp/redwood/issues/1604

### DIFF
--- a/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
+++ b/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
@@ -107,7 +107,7 @@ internal open class ViewLazyList private constructor(
             val viewsOnScreen = recyclerView.children.map { recyclerView.getChildViewHolder(it).bindingAdapterPosition }
             userHasScrolled = true // Prevent guest code from hijacking the scrollbar.
             onViewportChanged?.invoke(
-              viewsOnScreen.min(),
+              viewsOnScreen.min().coerceAtLeast(0),
               viewsOnScreen.max(),
             )
           }


redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -107,7 +107,7 @@ internal open class ViewLazyList private constructor(
             val viewsOnScreen = recyclerView.children.map { recyclerView.getChildViewHolder(it).bindingAdapterPosition }
             userHasScrolled = true // Prevent guest code from hijacking the scrollbar.
             onViewportChanged?.invoke(
-              viewsOnScreen.min(),
+              viewsOnScreen.min().coerceAtLeast(0),
```
jingwei99, line 110: `.min()` _sometimes_ returns -1 (which had been observed / is expected)

jingwei99:
I need another round of internal release when this is in.

JakeWharton (approved):

swankjesse (commented):
LGTM

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
JakeWharton: -1 is `NO_POSITION` which is what is returned for all positions between adapter invalidations and the next layout pass.

Also `min` and `max` throw if the sequence is empty, are we assuming the scroll callback will never be invoked on an empty list?

jingwei99: > Also min and max throw if the sequence is empty, are we assuming the scroll callback will never be invoked on an empty list?

ahhh 🤔, that's the case (not invoked on empty list) so far (this has only only been applied to emoji search). But it's not guaranteed the invocation wouldn't happen on an empty list (someone else might use it, or other teams might use it).  

> -1 is NO_POSITION which is what is returned for all positions between adapter invalidations and the next layout pass.

👍

PR #1608, opened 2023-10-18
Eliminate enum indirection to values

JakeWharton:


veyndan (approved):
I think this was necessary when we were using `@TestParameterInjector`. Makes sense to remove it now that we're no longer depending on it.

PR #1609, opened 2023-10-18
Eliminate allocation and boxing in `onScrolled` callback.

JakeWharton:
This callback is invoked on every frame when scrolling and therefore should not produce a lot of allocations. Previously it allocated multiple iterators/sequences and was subject to integer boxing and unboxing. Unrolling the functional operations into a traditional loop keeps data in primitives and never allocates an iterator or sequence.

### DIFF
--- a/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
+++ b/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
@@ -25,8 +25,9 @@ import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
 import android.widget.FrameLayout
 import androidx.annotation.ColorInt
-import androidx.core.view.children
 import androidx.core.view.doOnDetach
+import androidx.core.view.get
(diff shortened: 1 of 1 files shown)


redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -104,11 +105,21 @@ internal open class ViewLazyList private constructor(
       addOnScrollListener(
         object : RecyclerView.OnScrollListener() {
           override fun onScrolled(recyclerView: RecyclerView, dx: Int, dy: Int) {
-            val viewsOnScreen = recyclerView.children.map { recyclerView.getChildViewHolder(it).bindingAdapterPosition }
             userHasScrolled = true // Prevent guest code from hijacking the scrollbar.
+
+            var min = Int.MAX_VALUE
+            var max = Int.MIN_VALUE
+            for (i in recyclerView.indices) {
+              val position = recyclerView.getChildAdapterPosition(recyclerView[i])
```
JakeWharton, line 113: When this returns `NO_POSITION` we should probably just abandon all computation and wait for future scrolls post-layout pass. Did not want to change behavior in this refactor, though.

jingwei99 (approved):

PR #1613, opened 2023-10-18
Use the NSHomeDirectory on iOS

swankjesse:
The NSTemporaryDirectory had similar hazards to Android's cache directory, of being evicted externally.


JakeWharton (approved):
I assume each app has its own UID and thus home directory?

dnagler:
> I assume each app has its own UID and thus home directory?

[Yep](https://developer.apple.com/documentation/foundation/1413045-nshomedirectory):
> In iOS, the home directory is the application’s sandbox directory.

dnagler (approved):
Thanks!

PR #1619, opened 2023-10-20
LazyListScrollProcessor manages user and programmatic scrolls

jingwei99:


redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -105,22 +108,14 @@ internal open class ViewLazyList private constructor(
...
-              max = maxOf(max, position)
-            }
-            if (min == Int.MAX_VALUE || max == Int.MIN_VALUE) {
-              throw NoSuchElementException()
-            }
-            onViewportChanged?.invoke(
-              min.coerceAtLeast(0),
-              max,
-            )
+            if (scrollState == RecyclerView.SCROLL_STATE_IDLE) return
```
swankjesse, line 111: This is important! We were getting lots of scroll chatter as the LazyList loaded items

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyListState.kt
```diff
@@ -31,55 +31,43 @@ public fun rememberLazyListState(): LazyListState {
...
+    this.firstIndex = index
+    this.lastIndex = index + delta
   }
 
-  public fun onScrolled(firstVisibleItemIndex: Int) {
-    this.firstVisibleItemIndex = firstVisibleItemIndex
+  /** React to a user-initiated scroll. */
+  public fun onUserScroll(firstIndex: Int, lastIndex: Int) {
+    this.firstIndex = firstIndex
+    this.lastIndex = lastIndex
```
swankjesse, line 71: The big idea in this PR is to separate programmatic scrolls (data flows from guest to host) from interactive scrolls (data flows from host to guest).

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListScrollProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+
+  public fun onEndChanges() {
+    // Do nothing: we don't have deferred scrolls.
+    if (deferredProgrammaticScrollIndex == -1) return
+
+    // Do nothing: we don't do programmatic scrolls if the user has already scrolled manually.
+    if (userHasScrolled) return
+
+    // Do nothing: we can't scroll to this item because it hasn't loaded yet!
+    if (contentSize() <= deferredProgrammaticScrollIndex) return
```
swankjesse, line 47: This is also important  . . . we defer applying programmatic scrolls until there’s enough rows!

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyListState.kt
```diff
@@ -31,55 +31,43 @@ public fun rememberLazyListState(): LazyListState {
...
-    if (this.hasRestoredScrollPosition) return
-    this.hasRestoredScrollPosition = true
-
-    // Scroll to the target item.
-    if (restoredIndex != -1) {
-      scrollItemIndex = restoredIndex
-    }
+    val delta = (lastIndex - firstIndex)
+    this.firstIndex = index
+    this.lastIndex = index + delta
```
swankjesse, line 65: On a programmatic scroll, we set the loaded data window in anticipation!

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListScrollProcessor.kt
JakeWharton: I'm a bit worried about keeping the deferred value basically forever. Should we clear unconditionally after first non-zero size even if it's insufficient items (such as if data changed in background reducing rows)? This would depend on us loading the target range in a single operation rather than requiring multiple.

JakeWharton: If we have fewer rows when restoring (for whatever reason), and then later we gain some additional rows such that the index suddenly becomes valid, would it trigger a sudden scroll? Even if the user had done some manual scrolling?

JakeWharton: Okay I should have read one method down... 🤦 . I think these are covered!

JakeWharton (approved):

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -105,22 +108,14 @@ internal open class ViewLazyList private constructor(
...
+
+            val firstIndex = linearLayoutManager.findFirstVisibleItemPosition()
+            if (firstIndex == RecyclerView.NO_POSITION) return
+            val lastIndex = linearLayoutManager.findLastVisibleItemPosition()
```
jingwei99, line 115: `linearLayoutManager.findFirstVisibleItemPosition()` and `linearLayoutManager.findLastVisibleItemPosition()` work now!

redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
jingwei99: linearLayoutManager.findFirstVisibleItemPosition() and linearLayoutManager.findLastVisibleItemPosition() work now!

I'm guessing ti's because of `SCROLL_STATE_IDLE` (I didn't debug into it)

PR #1621, opened 2023-10-21
task 'browserRun' not found replace with jsBrowserRun

huanglizhuo:
update the demo apps README to use `jsBrowserRun`, otherwise it has below error report.

```
Cannot locate tasks that match 'samples:emoji-search:browser:browserRun' as task 'browserRun' not found in project ':samples:emoji-search:browser'. Some candidates are: 'jsBrowserRun'.
```

I am not familiar with gradle, but seems this is some change in kotlin multiplatform plugin?

### DIFF
--- a/samples/counter/README.md
+++ b/samples/counter/README.md
@@ -56,7 +56,7 @@ To simply view the sample app, open: https://cashapp.github.io/redwood/latest/co
 
 To build and modify the app locally, run this:
 ```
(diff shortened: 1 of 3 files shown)


JakeWharton (approved):
Thanks. It changed when we switched from the JS plugin to the multiplatform plugin for the browser modules.

PR #1624, opened 2023-10-23
Support animated programmatic scrolls

swankjesse:
In emoji search when the search terms is changed we scroll to the top of the list. This means we need to treat programmatic scrolls as something that can happen repeatedly, not just when we restore a lazy list.

This moves the feature that prevents programmatic scrolls from colliding with user scrolls from host code to guest code. The benefit is it no longer risks getting the loaded window out of sync with the true scroll window.

This also implements all scroll behavior for iOS.

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -26,6 +26,7 @@ import app.cash.redwood.layout.api.Constraint
 import app.cash.redwood.layout.api.CrossAxisAlignment
 import app.cash.redwood.lazylayout.api.ScrollItemIndex
 import app.cash.redwood.lazylayout.widget.LazyList
+import app.cash.redwood.lazylayout.widget.LazyListScrollProcessor
 import app.cash.redwood.lazylayout.widget.LazyListUpdateProcessor
 import app.cash.redwood.lazylayout.widget.LazyListUpdateProcessor.Binding
 import app.cash.redwood.lazylayout.widget.RefreshableLazyList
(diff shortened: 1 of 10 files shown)


swankjesse:
For manual testing, I hacked emoji search to scroll to the target index, rather than filtering the list.

```kotlin
  LaunchedEffect(searchTerm) {
    val scrollPosition = searchTerm.text.toIntOrNull()
    if (scrollPosition != null) {
      lazyListState.programmaticScroll(firstIndex = scrollPosition, animated = true)
    }
  }
```

https://github.com/cashapp/redwood/assets/133019/fb0fb3c6-eca9-48c8-8143-a1bf30391c74

https://github.com/cashapp/redwood/assets/133019/5c223444-3da8-4e7d-8c8c-11153eb44aba

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -136,15 +150,18 @@ internal open class UIViewLazyList(
...
     object : NSObject(), UITableViewDelegateProtocol {
       override fun scrollViewDidScroll(scrollView: UIScrollView) {
+        if (isDoingProgrammaticScroll) return // Only notify of user scrolls.
+
         val visibleIndexPaths = tableView.indexPathsForVisibleRows ?: return
+        if (visibleIndexPaths.isEmpty()) return
+
+        val firstIndex = visibleIndexPaths.minOf { (it as NSIndexPath).item.toInt() }
+        val lastIndex = visibleIndexPaths.maxOf { (it as NSIndexPath).item.toInt() }
+        scrollProcessor.onUserScroll(firstIndex, lastIndex)
```
JakeWharton, line 160: Does this need de-duping logic like Android? Or was that happening inside the actual processor?

JakeWharton (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: Bingo. It’s in the scroll processor where the logic can be shared.

PR #1627, opened 2023-10-23
Add DSL helpers to enable metrics and reports

### LINKED ISSUE #1616: Add DSL function to more easily facilitate dumping Compose compiler metrics
This requires adding a flag to the compiler which we could handle automatically.

Details on what it all means here: https://github.com/androidx/androidx/blob/androidx-main/compose/compiler/design/compiler-metrics.md

JakeWharton:
Closes #1616

### DIFF
--- a/redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodComposeExtension.kt
+++ b/redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodComposeExtension.kt
@@ -15,12 +15,10 @@
  */
 package app.cash.redwood.gradle
 
-import javax.inject.Inject
-import org.gradle.api.model.ObjectFactory
+import org.gradle.api.Action
 import org.gradle.api.provider.Property
 
-public abstract class RedwoodComposeExtension
-@Inject constructor(objectFactory: ObjectFactory) {
+public interface RedwoodComposeExtension {
   /**
    * The version of the JetBrains Compose compiler to use, or a Maven coordinate triple of
    * the custom Compose compiler to use.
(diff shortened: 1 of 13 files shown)


JakeWharton:
Actually this needs some work. We need to capture the Kotin compilation task name in the directory or else main/test or multiplatform will overwrite each other.

JakeWharton:
Screenshot 2023-10-23 at 8 20 52 PM

redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodComposeExtension.kt
```diff
@@ -39,7 +36,35 @@ public abstract class RedwoodComposeExtension
...
+   * directory. See
+   * [the compiler documentation](https://github.com/androidx/androidx/blob/androidx-main/compose/compiler/design/compiler-metrics.md#reports-breakdown)
+   * for more information about the contents.
+   *
+   * **NOTE:** This should only be enabled during investigation as it breaks the use of
+   * Gradle's build cache for this project's Kotlin compilation tasks.
+   *
+   * @see enableReports
+   */
+  public fun enableMetrics()
```
swankjesse: Worth incorporating the scary consequence into the function name?

```
enableMetrics_ThisBreaksGradleBuildCaching_()
```

I might inadvertently check in code that calls enableMetrics(), but my code reviewer would catch that!

swankjesse (approved):

redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodComposeExtension.kt
JakeWharton: It wouldn't actually be the end of the world if you did so. So maybe we can find something that's still cautionary but not as in-your-face. Will sleep on it.

JakeWharton: What about wrapping them into something else like
```groovy
redwood {
  dangerZone {
    enableMetrics()
  }
}
```
Is that enough to signal you should learn more about why it's dangerous?

swankjesse: Ooooh I like that a lot

JakeWharton: Done. Also added tests.

Issue #1594, opened 2023-10-14
Unify the wrapper nodes and tree manipulation between protocol and native Redwood


JakeWharton:
When we're doing Redwood Compose talks to the applier which then manipulates the nodes. When we're doing protocol Redwood the `ProtocolBridge` has its own nodes and its manipulates the tree directly. Both bottom out in widgets and the native UI toolkit, and the protocol is just an output of the applier interaction. We should be able to unify this code more than it already is so that more can be shared.

Issue #1084, opened 2023-05-16
Ability to decorate Widgets with functionality


chrisbanes:
There's a few use cases where we need to 'decorate' widgets with functionality, specific to the host platform. The API I keep coming back to is something similar to Compose UI's `Modifier`s, but with implementations bound for each platform.

Some examples:

### Pull to Refresh

Due to how the host platforms differ, we've had to bake 'pull to refresh' into `LazyColumn` and friends (#1082). That's not great as we're coupling two things that really shouldn't be coupled. The API is also gross. Ideally we'd have some way to decorate a `LazyColumn` instead:

``` kotlin
LazyColumn(
    modifier = Modifier.pullToRefresh(refreshing, onRefresh = { ... })
) {
    items(foo)
}
```

### Animations

This one is more hazy, as it's mostly just stuff in the back of my head. We need a way to run animations on widgets, and it's looking likely that our API structure will be similar to [React Native](https://reactnative.dev/docs/animations)'s, with a specific 'animated widget'. Something like:

``` kotlin
var animations by remember { mutableStateListOf() }

AnimatedWidget(
    animations = animations,
    onAnimatedEnd = { animations.remove(it) }
) {
    Text(
        text = "Foo",
        onClick = { animations += Animation(transformX = -100f) }
    )
}
```

Now we don't necessarily want `AnimatedWidget` to produce a host-side UI node. It will likely just be there to decorate the child.

JakeWharton:
Can the animated one be a modifier, too? A modifier designation (either another annotation or an interface implemented on the schema type) that changes it to participate in the view tree and probably in the factory with a callback to attach to the native view would be something that would be pretty straightforward.

I don't think we should do both forms if we can avoid it. Nodes in the Compose tree which don't show up in the native UI toolkit tree are possible, but likely more difficult.

chrisbanes:
> Can the animated one be a modifier, too?

Yeah, technically there's nothing wrong here, but the API may look a little weird. I'll sketch out some pseudo-code and see how it looks.

JakeWharton:
Other examples that keep coming up:
- Stable ID
- Accessibility (description, focus info)
- Background color

[...]

PR #1640, opened 2023-10-26
Move protocol modifier deserialization into factory

JakeWharton:
This is in preparation of adding index tracking to the protocol nodes to match support in the native widget nodes. `ProtocolNode` became an abstract class because of this, and common functionality to all subtypes moved into the class.

Additionally, create a split between the protocol factory which users need to reference and the generated protocol factory APIs that are unstable and used internally. This simplifies how these types and API signatures which use them look to users.

Looks big, but has a naming change so it's less big than you think...

Work towards #1084 and #1594.


redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ChangeListenerTest.kt
```diff
@@ -56,7 +56,7 @@ class ProtocolChangeListenerTest : AbstractChangeListenerTest() {
     snapshot: () -> T,
   ): TestRedwoodComposition<T> {
     val composeBridge = TestSchemaProtocolBridge.create()
-    val widgetBridge = ProtocolBridge(MutableListChildren(), TestSchemaProtocolNodeFactory(factories)) {
+    val widgetBridge = ProtocolBridge(MutableListChildren(), TestSchemaProtocolFactory(factories)) {
```
JakeWharton, line 59: The protocol traditionally used "compose" and "widget" naming, but in practice it's "guest" and "host", respectively. I'd like to do a rename of modules and types in a future change.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
```diff
@@ -90,8 +93,11 @@ public class ProtocolBridge<W : Any>(
           }
         }
         is ModifierChange -> {
+          val modifier = change.elements.fold<_, Modifier>(Modifier) { modifier, element ->
+            modifier.then(factory.createModifier(element))
+          }
```
JakeWharton: My first use of [underscore type operator](https://kotlinlang.org/docs/whatsnew17.html#underscore-operator-for-type-arguments) and I love it.

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolNode.kt
```diff
@@ -28,19 +28,27 @@ import app.cash.redwood.widget.Widget
  * @suppress
  */
 @RedwoodCodegenApi
-public interface ProtocolNode<W : Any> {
-  public val widget: Widget<W>
+public abstract class ProtocolNode<W : Any> {
+  public abstract val widget: Widget<W>
+
+  private var container: Widget.Children<W>? = null
```
JakeWharton: As the commit message details, soon this type will host `private val index: Int` so as to be able to remove and re-attach its `widget.view` within `container` at the correct position if the decorator modifiers change the instance.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/sharedHelpers.kt
```diff
@@ -224,3 +222,15 @@ private fun Deprecation.Level.toMemberName(): MemberName {
     },
   )
 }
+
+internal fun <T : Annotatable.Builder<T>> T.optIn(vararg names: ClassName): T = apply {
```
JakeWharton, line 226: Bearing the fruits of https://github.com/square/kotlinpoet/issues/1553

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
swankjesse: NEAT

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolNode.kt
swankjesse: Who’s responsible for shifting 'em when an element is removed in the middle?

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/widgetGeneration.kt
```diff
@@ -109,11 +105,7 @@ internal fun generateWidgetFactory(schema: Schema): FileSpec {
     .addType(
       TypeSpec.interfaceBuilder(widgetFactoryType)
         .addTypeVariable(typeVariableW)
-        .addAnnotation(
-          AnnotationSpec.builder(Stdlib.OptIn)
-            .addMember("%T::class", Stdlib.ExperimentalObjCName)
-            .build(),
-        )
+        .optIn(Stdlib.ExperimentalObjCName)
```
swankjesse, line 108: Reads so much better

swankjesse (approved):

redwood-protocol-widget/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolNode.kt
JakeWharton: For native Redwood we have a ChildrenNode which holds all of its WidgetNode children. For protocol Redwood we don't currently have that concept, and will need one.

Currently each ProtocolNode returns a Widget.Children when asked for a children container and the ProtocolBridge interacts with that directly. Next I'll instead return some wrapper around a Widget.Children that maintains the list of its ProtocolNode children and the Widgets into the Widget.Children. This way, when we do a tree operation, well ask this new type and it can update all the indices as well as perform the operation.

PR #1641, opened 2023-10-26
Rename protocol modules to 'guest' and 'host'

JakeWharton:
From 'compose' and 'widget', respectively.

Type renames will come in the future.


swankjesse (approved):
Oh hell yes

PR #1643, opened 2023-10-27
Test TreehouseAppContent and its internal state machines

swankjesse:


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
```diff
@@ -74,9 +73,9 @@ internal class TreehouseAppContent<A : AppService>(
   private val dispatchers: TreehouseDispatchers,
   private val appScope: CoroutineScope,
   private val eventPublisher: EventPublisher,
-  private val source: TreehouseContentSource<A>,
   private val codeListener: CodeListener,
-) : Content {
+  private val source: TreehouseContentSource<A>,
```
veyndan, line 77: [nitpick] Was the reordering deliberate?

redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeDispatchers.kt
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+import kotlinx.coroutines.CoroutineDispatcher
+
+class FakeDispatchers(
+  override val ui: CoroutineDispatcher,
+  override val zipline: CoroutineDispatcher,
```
veyndan, line 22: As we're not using these values, perhaps remove the constructor and just override these values in the class body?

veyndan (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
swankjesse: Yep! As a rule I try to order variables by their scope; most-singleton things first followed by least-singleton things. In this case the codeListener has a longer lifetime than the source.

(I don’t impose this on other people)

redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeDispatchers.kt
swankjesse: This one’s tricky. They’re `override` values from the interface.

PR #1646, opened 2023-10-27
Rename protocol-guest package name to match artifact

JakeWharton:


swankjesse (approved):

swankjesse (approved):
Manually confirmed this package doesn’t show up in any `.zipline-api.toml` files. I gotta get that `@ZiplineId` function working at some point!

JakeWharton:
Yep, thanks. Only the shared protocol types should. These are the impls sitting behind the scenes producing the shared values so it'd be a bad bug to leak them across.

Issue #1539, opened 2023-09-29
Treehouse Deprecation Dashboard


swankjesse:
This is an evergreen issue of in-flight deprecations.

## Background

We’ve got to worry about 2 kinds of compatibility.

 - **Regular API compatibility.** These work like any open source project: we avoid them when we can, and we make major version bumps when we can’t. (And until we hit 1.0, every release is allowed to make API changes.)

 - **Cross-Version Compatibility.** This is where a host application built with one Redwood version X and a guest application built with version Y need to interoperate smoothly.

## Policy

I propose we support the following:

**Hosts can run guest code built in the past 4 months, and for the next 16 months**

This puts this obligation on app owners that are always up-to-date on Redwood:

 * If you’re deploying new guest code, you can target apps up to 16 months old. (Probably closer to 12 months in practice because Redwood-upgrade day is not equal to release day.)
 * If you’re deploying new host code, you can only run apps built in the last 4 months.

## Deprecation Tracking

 * ~**2023-09-19** ZiplineTreehouseUi.start() changed from 3 arguments to 4. [PR link](https://github.com/cashapp/redwood/pull/1489)~
   * Removed in #2477 
 * ~**2023-10-31** ZiplineTreehouseUi.start() changed from 4 arguments to 1. [PR link](https://github.com/cashapp/redwood/pull/1647)~
   * Removed in #2477 
 * ~**2024-04-04** ChildrenChange.removedIds is no longer being sent to host [PR link](https://github.com/cashapp/redwood/pull/1936)~
   * Removed in #2470 
 * ~**2024-04-09** LazyList.placeholders throws when children are removed [PR link](https://github.com/cashapp/redwood/pull/1948)~
   * Removed in #2470 
 * **2024-12-12** `Remove.count` is not longer being sent to host. [PR link](https://github.com/cashapp/redwood/pull/2503)

Issue #1458, opened 2023-09-06
Uncaught Exceptions


swankjesse:
I made a button that throws an exception:

```kotlin
Button(
  text = "Die",
  onClick = {
    throw RuntimeException("boom")
  }
)
```

It causes a severe failure: the host program exits immediately.

```
FATAL EXCEPTION: Treehouse
Process: com.squareup.cash.beta.debug, PID: 16870
app.cash.zipline.ZiplineException: RuntimeException: boom
	at captureStack (runtime/coreRuntime.kt:12)
	at RuntimeException_init_$Create$_0 (kotlin-kotlin-stdlib-js-ir.js)
	at PlaygroundHome$Show$composable$lambda$lambda$lambda (src/commonMain/kotlin/com/squareup/cash/treehouse/playground/PlaygroundHome.kt)
	at <anonymous> (arcade/compose-protocol/build/generated/redwood/app/cash/arcade/compose/arcade/ProtocolButton.kt:107)
	at <anonymous> (arcade/compose-protocol/build/generated/redwood/app/cash/arcade/compose/ArcadeProtocolBridge.kt:30)
	at <anonymous> (Users/colinwhite/redwood/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt)
	at <anonymous> (Users/colinwhite/redwood/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineTreehouseUi.kt)
	at <anonymous> (redwood-redwood-treehouse.js)
	at <anonymous> (zipline/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/InboundService.kt:1391)
	at <anonymous> (zipline/zipline/src/commonMain/kotlin/app/cash/zipline/internal/bridge/Endpoint.kt:99)
	at <anonymous> (zipline-root-zipline.js)
	at <anonymous> (zipline/zipline/src/jsMain/kotlin/app/cash/zipline/GlobalBridge.kt)
	at <anonymous> (zipline-root-zipline.js)
	at app.cash.redwood.treehouse.ZiplineTreehouseUi$Companion$Adapter$GeneratedOutboundService.sendEvent(ZiplineTreehouseUi.kt:29)
	at app.cash.redwood.treehouse.ViewContentCodeBinding$sendEvent$1.invokeSuspend(TreehouseAppContent.kt:319)
[...]

PR #1647, opened 2023-10-27
New ZiplineTreehouseUi.Host bridges all host functions

swankjesse:
At the moment we've got 4 host functions offered to each UI. In a follow-up I'd like to introduce a 5th. I'd like the process of adding a 5th to not require yet another start() overload, so I'm introducing this type to make additional features easier to add later.

### DIFF
--- a/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
+++ b/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
@@ -135,7 +135,9 @@ private fun ZiplineTreehouseUi.Host.asOnBackPressedDispatcher() = object : OnBac
 }
 
 private fun OnBackPressedCallback.asService() = object : OnBackPressedCallbackService {
-  override var isEnabled: Boolean by this@asService::isEnabled
+  override var isEnabled: Boolean
+    get() = this@asService.isEnabled
+    set(value) = this@asService.isEnabled = value
 
   override fun handleOnBackPressed() {
     this@asService.handleOnBackPressed()
--- a/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
+++ b/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
@@ -17,6 +17,7 @@ package app.cash.redwood.treehouse
 
 import androidx.compose.runtime.saveable.SaveableStateRegistry
 import app.cash.redwood.compose.RedwoodComposition
+import app.cash.redwood.protocol.Change
 import app.cash.redwood.protocol.EventSink
 import app.cash.redwood.protocol.compose.ProtocolBridge
 import app.cash.redwood.protocol.compose.ProtocolRedwoodComposition
@@ -57,35 +58,41 @@ private class RedwoodZiplineTreehouseUi(
 
   private lateinit var saveableStateRegistry: SaveableStateRegistry
 
-  @Deprecated(
-    "Use `start` method that takes in an `OnBackPressedDispatcherService` instead.",
-    ReplaceWith("start(changesSink, TODO(), uiConfigurations, stateSnapshot)"),
-  )
+  @Suppress("OVERRIDE_DEPRECATION")
   override fun start(
     changesSink: ChangesSinkService,
     uiConfigurations: StateFlow<UiConfiguration>,
     stateSnapshot: StateSnapshot?,
   ) {
-    val onBackPressedDispatcher = object : OnBackPressedDispatcherService {
-      override fun addCallback(
-        onBackPressedCallback: OnBackPressedCallbackService,
-      ): CancellableService {
-        return object : CancellableService {
-          override fun cancel() = Unit
-        }
-      }
-    }
-    start(changesSink, onBackPressedDispatcher, uiConfigurations, stateSnapshot)
+    start(changesSink, NullOnBackPressedDispatcherService, uiConfigurations, stateSnapshot)
   }
 
+  @Suppress("OVERRIDE_DEPRECATION")
   override fun start(
     changesSink: ChangesSinkService,
     onBackPressedDispatcher: OnBackPressedDispatcherService,
     uiConfigurations: StateFlow<UiConfiguration>,
     stateSnapshot: StateSnapshot?,
   ) {
+    val host = object : ZiplineTreehouseUi.Host {
+      override val uiConfigurations = uiConfigurations
+      override val stateSnapshot = stateSnapshot
+
+      override fun sendChanges(changes: List<Change>) {
+        changesSink.sendChanges(changes)
+      }
+
+      override fun addOnBackPressedCallback(
+        callback: OnBackPressedCallbackService,
+      ): CancellableService = onBackPressedDispatcher.addCallback(callback)
+    }
+
+    start(host)
+  }
+
+  override fun start(host: ZiplineTreehouseUi.Host) {
     this.saveableStateRegistry = SaveableStateRegistry(
-      restoredValues = stateSnapshot?.content,
+      restoredValues = host.stateSnapshot?.content,
       // Note: values will only be restored by SaveableStateRegistry if `canBeSaved` returns true.
       // With current serialization mechanism of stateSnapshot, this field is always true, an update
       // to lambda of this field might be needed when serialization mechanism of stateSnapshot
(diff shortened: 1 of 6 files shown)


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
```diff
@@ -388,50 +394,44 @@ private class ViewContentCodeBinding<A : AppService>(
...
+      }
       try {
-        treehouseUi.start(
-          changesSink = this@ViewContentCodeBinding,
-          onBackPressedDispatcher = onBackPressedDispatcherService,
-          uiConfigurations = uiConfigurationFlow,
-          stateSnapshot = restoredState,
-        )
+        treehouseUi.start(this@ViewContentCodeBinding)
       } catch (e: ZiplineApiMismatchException) {
```
swankjesse, line 402: We’ve got 3 versions of this function:
  3 args
  4 args
  Host

If Host isn’t found this falls back to the 3-args form, losing support for BackPressed for some cases. I don’t believe that feature is widely used so I view this shortcut as being economical with code.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineTreehouseUi.kt
```diff
@@ -46,4 +49,12 @@ public interface ZiplineTreehouseUi : ZiplineService, EventSink {
...
 
   public fun snapshotState(): StateSnapshot? = null
+
+  /** Access to the host that presents this UI. */
+  public interface Host : ZiplineService, ChangesSinkService {
+    public val uiConfigurations: StateFlow<UiConfiguration>
+    public val stateSnapshot: StateSnapshot?
+
+    public fun addOnBackPressedCallback(callback: OnBackPressedCallbackService): CancellableService
+  }
```
swankjesse, line 59: Strictly speaking introducing this Host abstraction is a bit less efficient than passing the 4 parameters inline. This is because we need an extra 2 host-guest bridges for the two value parameters (uiConfigurations and stateSnapshot) that were originally passed by value into `start()`. I think this is a good tradeoff for simplicity.

swankjesse:

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
veyndan: I'm assuming "losing support for BackPressed for some cases" means "losing support for `BackHandler` for some cases."

I'm in the midst of adding native back support to our internal flows project. It sounds frustrating from a user perspective if the custom back override only occurs sometimes.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/OnBackPressedDispatcherService.kt
```diff
@@ -22,5 +22,5 @@ import kotlin.native.ObjCName
 /** Redwood's [OnBackPressedDispatcher] but implementing [ZiplineService]. */
 @ObjCName("OnBackPressedDispatcherService", exact = true)
 public interface OnBackPressedDispatcherService : ZiplineService {
-  public fun addCallback(onBackPressedCallback: OnBackPressedCallbackService): CancellableService
+  public fun addCallback(callback: OnBackPressedCallbackService): CancellableService
```
veyndan: Revert as the names should match those in `OnBackPressedDispatcher`. These names were taken from Compose UI's, and I'd rather not diverge without reason.

```suggestion
  public fun addCallback(onBackPressedCallback: OnBackPressedCallbackService): CancellableService
```

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineTreehouseUi.kt
```diff
@@ -46,4 +49,12 @@ public interface ZiplineTreehouseUi : ZiplineService, EventSink {
   )
 
   public fun snapshotState(): StateSnapshot? = null
+
+  /** Access to the host that presents this UI. */
+  public interface Host : ZiplineService, ChangesSinkService {
+    public val uiConfigurations: StateFlow<UiConfiguration>
+    public val stateSnapshot: StateSnapshot?
+
+    public fun addOnBackPressedCallback(callback: OnBackPressedCallbackService): CancellableService
```
veyndan, line 58: Can we pass the `OnBackPressedDispatcherService` here? The fact that `OnBackPressedDispatcherService` has only one function is happenstance, and if we were to add more functions to `OnBackPressedDispatcherService`, I'd prefer to not have to also update `Host`.

swankjesse: Great question!

The Host interface is a Redwood-internal implementation detail so we can reduce the number of ZiplineService instances we need to bridge, and also the number we need to close.

This is one of those things where we’ve gradually added more things as we add more capabilities, and we’re overdue to flatten them all both for forwards-compatibility and for performance.

Rather than having one ZiplineService-style adapter for every feature we bridge (changes sink, save, back, and soon exceptions), this flattens them all into a single interface to reduce the runtime setup. When we need a new capability, we don’t need a new overload and a new catch clause to fall back on.

Instead of adding new BackHandler features to BackHandlerService to bridge them, we will instead add them to this Host interface. It doesn’t limit our ability to introduce new features for back handling in the future.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/OnBackPressedDispatcherService.kt
swankjesse: Can fix.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
swankjesse: The situation where we lose back handling is when a user’s host app is on a release that has the 4-parameter overload and they guest app has this new interface.

I don’t expect that to impact any of Cash App’s production users. I didn’t see any uses of BackHandler in our production code?

If I’m mistaken it’s easy to restore that functionality! I just need another pair of try/catch clauses to support falling back to that intermediate case.

veyndan: Gotcha, we don't have any usage of `BackHandler` within production code right now, but we will very soon (like within a week or so).

Just to double check, if we:

1. Merge this PR
2. Do an internal release containing these changes
3. Track the new release in all our repos
4. If we then use and ship usages of `BackHandler`, the `BackHandler` will work in ALL cases (i.e., as before)?

I just want to make sure that we don't get into a weird state where the `BackHandler` is invoked inconsistently for users where it should work!

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/ZiplineTreehouseUi.kt
veyndan: That makes sense. Thanks for clarifying!

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseAppContent.kt
swankjesse: Yep, exactly.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/OnBackPressedDispatcherService.kt
swankjesse: Done.

veyndan (approved):
Update https://github.com/cashapp/redwood/issues/1539?

swankjesse:
Good call. Done.

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
```diff
@@ -119,24 +126,25 @@ private class RedwoodZiplineTreehouseUi(
...
 
-private fun OnBackPressedCallback.asService(): OnBackPressedCallbackService {
-  return object : OnBackPressedCallbackService {
-    override var isEnabled: Boolean
-      get() = this@asService.isEnabled
-      set(value) {
-        this@asService.isEnabled = value
-      }
+private fun OnBackPressedCallback.asService() = object : OnBackPressedCallbackService {
+  override var isEnabled: Boolean by this@asService::isEnabled
```
JakeWharton: This is dramatically more expensive than it seems because it requires a `KProperty` to be created and used to read/write to the property. It's more verbose, but this would be much more efficient:
```suggestion
  override var isEnabled: Boolean
    get() = this@asService.isEnabled
    set(value) { this@asService.isEnabled = value }
```

JakeWharton (approved):

PR #1648, opened 2023-10-27
Add unit tests for OnBackPressedDispatcher

swankjesse:


swankjesse:
# ⚠️ stacked PR!

swankjesse:

veyndan (approved):

JakeWharton (approved):
Are we testing real codepaths here? There's a lot of fake infrastructure in play and it's hard to understand what real behavior we're trying to test.

swankjesse:
The `TreehouseAppContent` is the real codepath this is exercising. It’s at the center of 3 different lifecycles (view attach, code load, content), and so it’s been very scary to change. With this I’m hoping to make it a little bit better.

I also see some options to promote other things to be under test, perhaps by moving the boundaries of what’s an interface and what’s not.

JakeWharton:
It's probably easier to see in the IDE where I can click around more easily. It's hard to see exact boundaries in the diff.

PR #1651, opened 2023-10-29
CodeListener.onUncaughtException

swankjesse:


swankjesse:
https://github.com/cashapp/redwood/assets/133019/79e0448c-51c7-4a03-86ea-0dba6983c6fa

swankjesse:
# ⚠️ stacked PR!

swankjesse:

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -183,22 +186,41 @@ public class TreehouseApp<A : AppService> private constructor(
...
-  ) : CodeHost<A> {
-    override var session: ZiplineCodeSession<A>? = null
+  private class TreehouseDispatchersWithExceptionHandler(
+    val delegate: TreehouseDispatchers,
+    exceptionHandler: CoroutineExceptionHandler,
+  ) : TreehouseDispatchers by delegate {
+    override val zipline = delegate.zipline + exceptionHandler
+  }
 
+  private inner class ZiplineCodeHost<A : AppService> : CodeHost<A> {
```
veyndan, line 196: Was their a requirement to make this an `inner class`?

(My over dislike of `inner class` may have motivated this comment 😅)

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseDispatchers.kt
```diff
@@ -15,21 +15,24 @@
...
  */
 @ObjCName("TreehouseDispatchers", exact = true)
 public interface TreehouseDispatchers {
-  public val ui: CoroutineDispatcher
-  public val zipline: CoroutineDispatcher
+  /** Must contain a non-null [kotlinx.coroutines.CoroutineDispatcher]. */
+  public val ui: CoroutineContext
+
+  /** Must contain a non-null [kotlinx.coroutines.CoroutineDispatcher]. */
+  public val zipline: CoroutineContext
```
veyndan, line 35: These comments feel weird as we're saying that implementations should be a `CoroutineDispatcher`, when the type is a `CoroutineContext`. Is it really a requirement that a `CoroutineDispatcher` instance be passed?

veyndan (commented):
That video looks pretty sweet

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
swankjesse: It reduces the amount of boilerplate to share state with the enclosing class.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseDispatchers.kt
swankjesse: Ooooh good call out. Lemme fix those. (I had temporarily made that change, and then rolled it back.)

swankjesse: Actually, as of this PR I haven’t rolled it back, but that happens in https://github.com/cashapp/redwood/pull/1655

I was trying to find a good way to include a CoroutineExceptionHandler, but I found it was easier to put that in the CoroutineScope instead of the CoroutineDispatcher. These CoroutineContext APIs are very dynamic so I tried it both ways and I like the other way better!

veyndan (approved):

samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchActivity.kt
```diff
@@ -78,11 +81,22 @@ class EmojiSearchActivity : ComponentActivity() {
...
     treehouseLayout = TreehouseLayout(this, widgetSystem, onBackPressedDispatcher).apply {
-      treehouseContentSource.bindWhenReady(this, treehouseApp)
+      treehouseContentSource.bindWhenReady(this, treehouseApp, codeListener)
     }
     setContentView(treehouseLayout)
   }
 
+  private val codeListener: CodeListener = object : CodeListener() {
+    override fun onUncaughtException(view: TreehouseView<*>, e: Throwable) {
+      val treehouseLayout = view as TreehouseLayout
```
JakeWharton:
```suggestion
```
`treehouseLayout` is available on the instance. I would strongly prefer code listener implementations close over their precise `TreehouseView` subtype rather than doing unsafe casts. In real code this means that if you change UI toolkits (from views to Compose UI), the code below fails to compile, whereas with the unsafe cast it compiles but crashes at runtime.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -125,10 +124,14 @@ public class TreehouseApp<A : AppService> private constructor(
    * Continuously polls for updated code, and emits a new [LoadResult] instance when new code is
    * found.
    */
+  @OptIn(ExperimentalStdlibApi::class)
```
JakeWharton, line 127: What is this for? Whatever it is, it needs added to the README's list of unstable API usage.

JakeWharton (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
swankjesse: It’s for this:

```
    val dispatcher = dispatchers.zipline[CoroutineDispatcher.Key]
      ?: error("expected TreehouseDispatchers.zipline to include a CoroutineDispatcher")
```

But I get rid of this in a follow-up on the stack because I found a better way!

JakeWharton: I also forgot what repo I'm in. Zipline has the readme table. We can tolerate instability for now anyway.

PR #1653, opened 2023-10-30
Handle uncaught exceptions from async calls

swankjesse:
When the guest throws an uncaught exception, a guest-side CoroutineExceptionHandler now forwards it to the host via AppLifecycle.handleUncaughtException(). That now triggers a graceful break of all UIs.

### DIFF
--- a/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
+++ b/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
@@ -370,4 +370,20 @@ public open class EventListener {
     name: String,
   ) {
   }
+
+  /**
+   * Invoked when [app] has thrown an uncaught exception.
+   *
+   * This indicates an unrecoverable software bug. Development implementations should report the
+   * exception to the developer. Production implementations should post the exception to a bug
+   * tracking service.
+   *
+   * When a Treehouse app fails its current [Zipline] instance is canceled so no further code will
+   * execute. A new [Zipline] will start when new code available, or when the app is restarted.
+   */
+  public open fun uncaughtException(
+    app: TreehouseApp<*>,
+    exception: Throwable,
+  ) {
+  }
 }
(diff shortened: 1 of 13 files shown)


swankjesse:
https://github.com/cashapp/redwood/assets/133019/a33c22d7-e1b3-4bc6-a727-6777083c8ab1

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -370,4 +370,20 @@ public open class EventListener {
...
+   * Invoked when [app] has thrown an uncaught exception.
+   *
+   * This indicates an unrecoverable software bug. Development implementations should report the
+   * exception to the developer. Production implementations should post the exception to a bug
+   * tracking service.
+   *
+   * When a Treehouse app fails its current [Zipline] instance is canceled so no further code will
+   * execute. A new [Zipline] will start when new code available, or when the app is restarted.
+   */
+  public open fun uncaughtException(
```
swankjesse, line 384: For follow-up, I’d like more Zipline-specific context here. The ZiplineManifest has the release version, and that’ll be super valuable for reporting upstream.

swankjesse:

swankjesse:
# ⚠️ stacked PR!

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppLifecycle.kt
```diff
@@ -30,8 +30,16 @@ public interface AppLifecycle : ZiplineService {
   public interface Host : ZiplineService {
     public fun requestFrame()
 
+    /** Notify the host that an event was unrecognized and will be ignored. */
     public fun onUnknownEvent(widgetTag: WidgetTag, tag: EventTag)
 
+    /**
+     * Notify the host that an event was received for a node that either never existed or that no
+     * longer exists. That event will be ignored.
```
JakeWharton:
```suggestion
     * Notify the host that an event was received for a node that no longer exists.
     * That event will be ignored.
```
A node that never existed should never happen if our code is correct, whereas a no-longer-existing node is quite plausible.

JakeWharton (approved):

PR #1657, opened 2023-10-31
Show an error screen on iOS

swankjesse:


swankjesse:
https://github.com/cashapp/redwood/assets/133019/8fcc5a6a-e860-4b71-a38a-5269c8f590cd

swankjesse:
https://github.com/cashapp/redwood/assets/133019/58867edc-f6f9-4cb5-8be0-06473cb107f0

swankjesse:
Part of: https://github.com/cashapp/redwood/issues/1458

swankjesse:
# ⚠️ stacked PR!

samples/emoji-search/ios-uikit/EmojiSearchApp/EmojiSearchViewController.swift
```diff
@@ -71,6 +71,23 @@ class EmojiSearchViewController : UIViewController, EmojiSearchEventListener {
     }
 }
 
+class EmojiSearchCodeListener : CodeListener {
+    override func onUncaughtException(view: TreehouseView, exception: KotlinThrowable) {
+        let treehouseLayout = view as! TreehouseUIView
```
JakeWharton: Can you pass in the `TreehouseUIView` instance through the constructor rather than unsafe cast? Same justification as the other PR.

I want to remove the `TreehouseView` instance from the code listener at some point...

JakeWharton (approved):

samples/emoji-search/ios-uikit/EmojiSearchApp/EmojiSearchViewController.swift
swankjesse: Fixed!

I think we’re probably more ready to lose that parameter now than we’ve ever been. Personally I like it because the function calls are all scoped to that instance. It’s not just that there’s code loading, there’s code loading _for that view_. I also think it’s both true that a ClassCastException could happen if you switch from ComposeUI to Views, but also that it’ll be very weird to have both in the same build path at the same time.

PR #1659, opened 2023-10-31
Two bits of cleanup In UIViewLazyList

dnagler:
- Removing call to set contentView.translatesAutoresizingMaskIntoConstraints to false, because this is not supported by iOS, and yields the following warning:

"Changing the translatesAutoresizingMaskIntoConstraints property of the contentView of a UITableViewCell is not supported and will result in undefined behavior, as this property is managed by the owning UITableViewCell"

- Adding call to removeAllSubviews within prepareForReuse

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -239,7 +239,6 @@ internal class LazyListContainerCell(
       removeAllSubviews()
       if (value != null) {
         contentView.addSubview(value.value)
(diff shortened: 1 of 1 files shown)


JakeWharton (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -273,6 +272,7 @@ internal class LazyListContainerCell(
 
   override fun prepareForReuse() {
     super.prepareForReuse()
+    removeAllSubviews()
```
underscoretang, line 275: this sounds right for what we have. wishful thinking would be that these views stay put and redwood could just modify the values of them to actually reuse the cell

underscoretang (approved):

swankjesse (approved):

PR #1664, opened 2023-11-01
Scope all events to a code session

swankjesse:
This gives us a code version to attribute failures to.

This is a backwards-incompatible change; we need to pass an EventListener.Factory instead of an EventListener when creating a TreehouseApp.

This drops the appStart() and appCanceled() events. These are roughly approximated by the scoped codeLoadStart() and codeUnloaded() functions.

### DIFF
--- a/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
+++ b/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
@@ -47,9 +47,6 @@ public class TreehouseApp<A : AppService> private constructor(
 ) {
   private val codeHost = ZiplineCodeHost<A>()
 
-  // TODO: remove after https://github.com/cashapp/zipline/pull/1159
-  private val ziplineToPublisherHack = mutableMapOf<Zipline, RealEventPublisher>()
-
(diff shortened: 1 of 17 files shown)


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -46,9 +47,10 @@ public class TreehouseApp<A : AppService> private constructor(
 ) {
   private val codeHost = ZiplineCodeHost<A>()
 
-  public val dispatchers: TreehouseDispatchers = factory.dispatchers
+  // TODO: remove after https://github.com/cashapp/zipline/pull/1159
+  private val ziplineToPublisherHack = mutableMapOf<Zipline, RealEventPublisher>()
```
swankjesse: I’m massively cheating here 'cause I want to correlate the Zipline-layer EventListener with the Treehouse-layer EventListener, but I don’t control the creation of the Zipline-layer one.

That PR has the fix!

swankjesse: I’d like to hold this until I can rebase on that upgrade.

JakeWharton (approved):

swankjesse:

Issue #1636, opened 2023-10-25
Pipe app lifecycle into Redwood and Treehouse composition


JakeWharton:
For when the render goes off-screen (Activity pause, backgrounded view controller, etc.).

Two goals:
* Stop doing composition work (i.e., [`collectAsStateWithLifecycle`](https://developer.android.com/reference/kotlin/androidx/lifecycle/compose/package-summary#(kotlinx.coroutines.flow.Flow).collectAsStateWithLifecycle(kotlin.Any,androidx.lifecycle.Lifecycle,androidx.lifecycle.Lifecycle.State,kotlin.coroutines.CoroutineContext)) and [`LifecycleResumeEffect`](https://developer.android.com/reference/kotlin/androidx/lifecycle/compose/package-summary#LifecycleResumeEffect(kotlin.Any,androidx.lifecycle.LifecycleOwner,kotlin.Function1)))
* Stop ticking the frame clock so composition pauses

This might need to be added to the tester, too. We'll see whether simply deferring calls to `awaitSnapshot()` is enough to replicate this behavior.

JakeWharton:
Hopefully we can get away with a single state so that it's `collectAsStateWhileVisible()` and `VisibleEffect`!

PR #1666, opened 2023-11-02
WIP - RootView Visibility

underscoretang:
All just placeholders for discussion for now.

What we want is for treehouse apps to be able to subscribe to changes in the visibility of the root view from the operating system.

Here's a made up use case
```kotlin
@Composble
fun Show() {
  val isVisible = UIConfiguration.current.isVisible.observeAsState(initial = true)

  if (isVisible) {
      pollForMoreData(interval: 10)
  }

  Text(someData)
}
```

`isVisible` is a placeholder name. We need to decide what this should be called so that all platforms have an clear idea of what it means.

These are some ideas of what this could mean on each:

| isVisible | iOS | Android | Web |
| --- | --- | --- | --- |
| true | view.window != nil | view.visibility == VISIBLE | window.load |
| false | view.window == nil | view.visibility == GONE | INVISIBLE | window.unload |

### DIFF
--- a/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/UiConfiguration.kt
+++ b/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/UiConfiguration.kt
@@ -37,6 +37,19 @@ public class UiConfiguration(
    * raw pixels, if needed.
    */
   public val density: Double = 1.0,
+
+  /**
+   * TODO: The visibility of the root view on the screen.
+   * The use case for this is so that a treehouse app can react
+   * to their view going on and off the screen. This means
+   * slightly different things on every platform, but the goal is 
+   * to find common ground.
+   * Some ideas:
+   * iOS - view.window != nil or viewDidAppear vs viewDidDisappear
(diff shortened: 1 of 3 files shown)


JakeWharton:
I have the design for this started in https://github.com/cashapp/redwood/issues/1636. I haven't thought about what mechanism the signal comes in through. `UiConfiguration` might work well.

underscoretang:
@JakeWharton, let us know if Ari and I can help with this one. We'd love to learn more about how to add features into redwood and get this in in time for our use case

JakeWharton:
Wiring this value up as this PR proposed would be a sufficiently good start. The design inside of Compose code is mostly separate from the design of how the value gets into Compose.

JakeWharton:
Note that on Android the view's visibility is not the correct signal. Views within an activity hierarchy that are on the backstack are still declared as visible, they're just not rendering. There is no view-level signal. We'd need to extract the `Lifecycle` from the view tree which originates from the enclosing activity and monitor it for start/stop lifecycle changes.

For the web we probably want to use [the page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to read the current visibility state and then monitor it for changes.

PR #1667, opened 2023-11-02
Make a CodeHost state machine

### LINKED ISSUE #1660: Make TreehouseApp more state machiney
Right now there’s a lot of implicit states in `TreehouseApp` and its current code. I’d like to make this more structured to support code crashes and restarts.

These are our states:

 * `Idle`
 * `Starting`: the ZiplineLoader is collecting code updates, waiting for a Zipline
 * `Running`: the ZiplineLoader is collecting code updates, and a Zipline is ready
 * `Crashed`: the ZiplineLoader is collecting code updates, but the most recent Zipline failed

I should be able to avoid `Stopping` and `Crashing` states by doing the corresponding cleanup asynchronously. Ie. when the app crashes it’ll ‘immediately’ transition to `Crashed` even though the old Zipline instance is still being torn down.

When the Zipline instance is updated via a hot-reload, it stays in the `Running` state without transition.

This is similar to [Guava’s service](https://github.com/google/guava/wiki/ServiceExplained). But unlike that interface I’d like to make `TreehouseApp` support multiple starts and stops. That way the consuming application doesn’t need to build its own lifecycles on top.

These states won’t necessarily appear in the public API anywhere, but functions to transition between them will:

 * start() - transition to `Starting` unless it’s `Starting` or `Running`.
 * stop() - transition to `Idle` immediately
 * restart() - transition to `Starting` unless it’s already there.

There’s no function to transition to `Crashed`; you have to do that by running bad code.

swankjesse:
This simplifies much of the internal complexity of TreehouseApp. We can also delete FakeCodeHost because the single implementation is suitable for unit testing.

Closes: https://github.com/cashapp/redwood/issues/1660

### DIFF
--- a/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeHost.kt
+++ b/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeHost.kt
@@ -15,16 +15,188 @@
  */
 package app.cash.redwood.treehouse
 
-/** Manages loading and hot-reloading a series of code sessions. */
-internal interface CodeHost<A : AppService> {
-  val stateStore: StateStore
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.job
+import kotlinx.coroutines.launch
 
-  /** Only accessed on [TreehouseDispatchers.ui]. */
-  val session: CodeSession<A>?
+/**
+ * Manages loading and hot-reloading a series of code sessions.
+ *
+ * The code host has 4 states:
+ *
+ *  * `Idle`
+ *  * `Starting`: collect code updates, and wait for an `Zipline` to load.
+ *  * `Running`: collect code updates, and a `Zipline` is running.
+ *  * `Crashed`: collect code updates, but the most recent `Zipline` failed.
+ *
+ * Transitions between states always occur on the UI dispatcher. These functions initiate state
+ * transitions:
+ *
+ *  * `start()` - transition to `Starting` unless it’s `Starting` or `Running`.
+ *  * `stop()` - transition to `Idle` immediately
+ *  * `restart()` - transition to `Starting` unless it’s already `Starting`.
+ *
+ * Other state transitions also occur:
+ *
+ *  * From `Starting` to `Running` when a `Zipline` finishes loading.
+ *  * From `Running` to `Crashed` when a `Zipline` fails.
+ *  * From `Running` to `Running` when the `Zipline` is replaced by a hot-reload.
+ */
+internal abstract class CodeHost<A : AppService>(
+  private val dispatchers: TreehouseDispatchers,
+  private val appScope: CoroutineScope,
+  private val frameClockFactory: FrameClock.Factory,
+  val stateStore: StateStore,
+) {
+  /** Contents that this app is currently responsible for. */
+  private val listeners = mutableListOf<Listener<A>>()
 
-  fun addListener(listener: Listener<A>)
+  private var state: State<A> = State.Idle()
 
-  fun removeListener(listener: Listener<A>)
+  private val codeSessionListener = object : CodeSession.Listener<A> {
+    override fun onUncaughtException(codeSession: CodeSession<A>, exception: Throwable) {
+    }
+
+    override fun onCancel(codeSession: CodeSession<A>) {
+      dispatchers.checkUi()
+
+      codeSession.removeListener(this)
+
+      // If a code session is canceled while we're still listening to it, it must have crashed.
+      val previous = state
+      if (previous is State.Running) {
+        state = State.Crashed(previous.codeUpdatesScope)
+      }
+    }
+  }
+
+  val codeSession: CodeSession<A>?
+    get() = state.codeSession
+
+  /** Returns a flow that emits a new [CodeSession] each time we should load fresh code. */
+  abstract fun codeUpdatesFlow() : Flow<CodeSession<A>>
+
+  fun start() {
+    dispatchers.checkUi()
+
+    val previous = state
+
+    if (previous is State.Starting || previous is State.Running) return // Nothing to do.
+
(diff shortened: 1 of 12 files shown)


swankjesse:
I’d like to add tests before reviewing this.

redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeCodeSession.kt
```diff
@@ -18,22 +18,32 @@ package app.cash.redwood.treehouse
 import app.cash.redwood.treehouse.CodeSession.Listener
 import app.cash.redwood.treehouse.CodeSession.ServiceScope
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.job
+import kotlinx.coroutines.launch
 import kotlinx.serialization.json.Json
 
 internal class FakeCodeSession(
```
swankjesse, line 27: I think I can refactor the ZiplineCodeSession state management to share a lot more code with this fake.

redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeOnBackPressedDispatcher.kt
```diff
@@ -19,17 +19,27 @@ import app.cash.redwood.ui.Cancellable
...
 
   override fun addCallback(onBackPressedCallback: OnBackPressedCallback): Cancellable {
+    val id = nextCallbackId++
     mutableCallbacks += onBackPressedCallback
 
     return object : Cancellable {
+      var canceled = false
+
       override fun cancel() {
+        if (canceled) return
```
swankjesse, line 39: I don’t like that these are being canceled multiple times. It’s not a bug, but it’s inelegant.

redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/TreehouseAppContentTest.kt
```diff
@@ -20,35 +20,59 @@ import assertk.assertThat
...
+    frameClockFactory = FakeFrameClock.Factory,
+  )
   private val codeListener = FakeCodeListener(eventLog)
   private val uiConfiguration = UiConfiguration()
 
+  @BeforeTest
+  fun setUp() {
+    runBlocking {
+      codeHost.start()
+      eventLog.takeEvent("codeHost.collectCodeUpdates()")
```
swankjesse: I added an event for this so I can tell when we’re receiving new code and when we’re not. In this test it’s not that interesting!

redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/CodeHostTest.kt
```diff
@@ -0,0 +1,244 @@
...
+    eventLog.takeEvent("codeSessionA.app.uis[0].start()")
+
+    codeHost.stop()
+    eventLog.takeEvent("codeSessionA.app.uis[0].close()")
+    eventLog.takeEvent("codeSessionA.cancel()")
+
+    content.unbind()
+  }
+
+  /** CodeHost doesn't have to stay resident forever. */
```
swankjesse, line 96: This is probably the most exciting thing about the PR. We can now shut down a Zipline completely and start it up again. That’ll be potentially great for saving memory.

swankjesse:
# ⚠️ stacked PR!

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -62,7 +71,7 @@ public class TreehouseApp<A : AppService> private constructor(
    * instance may be replaced if new code is loaded.
    */
   public val zipline: Zipline?
-    get() = codeHost.session?.zipline
+    get() = (codeHost.codeSession as? ZiplineCodeSession)?.zipline
```
veyndan, line 74: Is `codeSession` not being a `ZiplineCodeSession` valid?

If not, perhaps we can make `CodeSession` itself a type parameter?

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeHost.kt
```diff
@@ -15,16 +15,184 @@
...
+    ) : State<A>()
+
+    class Starting<A : AppService>(
+      override val codeUpdatesScope: CoroutineScope,
+    ) : State<A>()
+
+    class Crashed<A : AppService>(
+      override val codeUpdatesScope: CoroutineScope,
+    ) : State<A>()
+  }
```
veyndan, line 195: Personally, I would've loved to see the states hoisted inside the respective states, in order to get the true benefits of making a state machine (i.e., it being easy to follow how one state gets to the next, a thing which is still difficult in this PR).

Apologies if that was a planned follow up!

veyndan (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeHost.kt
swankjesse: Yeah lemme explore that in follow up.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
swankjesse: Lemme explore that in follow up!

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeHost.kt
swankjesse: A month later . . . we looked into this, and didn’t find something I loved!

@veyndan if you’d like to run with this, please do!

PR #1670, opened 2023-11-07
Add tests for `BackHandler` composable

veyndan:


redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/BackHandlerTest.kt
```diff
@@ -0,0 +1,204 @@
...
+      assertThat(awaitSnapshot()).contains(TextValue(text = "0"))
+      onBackPressedDispatcher.onBackPressed()
+      assertNoSnapshot()
+    }
+  }
+}
+
+private suspend fun TestRedwoodComposition<List<WidgetValue>>.assertNoSnapshot() {
+  assertFailure { awaitSnapshot() }.isInstanceOf<TimeoutCancellationException>()
+}
```
JakeWharton, line 191: I wonder if we can provide a proper API for this rather than relying on the timeout.

redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/BackHandlerTest.kt
```diff
@@ -0,0 +1,204 @@
...
+    onBackPressedCallbacks += onBackPressedCallback
+    return object : Cancellable {
+      override fun cancel() {
+        onBackPressedCallbacks -= onBackPressedCallback
+      }
+    }
+  }
+
+  fun onBackPressed() {
+    val callbackToNotify = onBackPressedCallbacks.lastOrNull { it.isEnabled }
```
JakeWharton: Because this fake implements the behavior that chooses the last callback to invoke, the tests which put multiple back handlers into the composition (at least when both are enabled) are exercising this behavior and not any actual library behavior.

colinrtwhite (approved):

JakeWharton (approved):

redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/BackHandlerTest.kt
JakeWharton: Doesn't seem like it

redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/BackHandlerTest.kt
veyndan: Good point. Just updated the PR to inline this function and add the desired assertions.

Issue #1672, opened 2023-11-08
`BackHandler` doesn't propagate `enabled` changing


veyndan:
For example, when running the following composable function:

```kotlin
@Composable
fun Foo() {
  var enabled by remember { mutableStateOf(true) }

  println("enabled = $enabled")
  BackHandler(enabled) { println("Back pressed") }

  LaunchedEffect(Unit) {
    delay(5.seconds)
    enabled = false
  }
}
```

We get the following output:

```
13:30:21   enabled = true
13:30:23   Back pressed
13:30:26   enabled = false
13:30:28   Back pressed
```

Having the initial value be `false` causes "Back pressed" to never be printed.

PR #1673, opened 2023-11-08
Propagate `BackHandler#enabled` changing in Redwood

veyndan:
This only fixes the bug for Redwood. I'll make a follow up PR(s) to fix it for Treehouse.

Refs #1672

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,6 +3,7 @@ kotlin = "1.9.20"
 kotlinx-coroutines = "1.7.3"
 kotlinx-serialization = "1.6.0"
 androidx-activity = "1.8.0"
+androidx-compose-ui = "1.5.4"
 jbCompose = "1.5.10"
 lint = "31.1.1"
 zipline = "1.5.0"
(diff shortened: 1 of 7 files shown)


redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/OnBackPressedCallback.kt
```diff
@@ -17,6 +17,12 @@ package app.cash.redwood.ui
 
 public abstract class OnBackPressedCallback(enabled: Boolean) {
   public var isEnabled: Boolean = enabled
+    set(value) {
+      field = value
+      enabledChangedCallback?.invoke()
+    }
+
+  public var enabledChangedCallback: (() -> Unit)? = null
```
veyndan, line 25: This mechanism was inspired by Android's implementation:

https://github.com/androidx/androidx/blob/dd1fc315eabedbcd024dd5ae219961c7236031a1/activity/activity/src/main/java/androidx/activity/OnBackPressedCallback.kt#L51-L60

gradle/libs.versions.toml
```diff
@@ -37,6 +37,8 @@ androidx-activity = { module = "androidx.activity:activity", version.ref = "andr
 androidx-activity-compose = { module = "androidx.activity:activity-compose", version.ref = "androidx.activity" }
 androidx-appCompat = { module = "androidx.appcompat:appcompat", version = "1.6.1" }
 androidx-compose-compiler = "androidx.compose.compiler:compiler:1.5.4"
+androidx-compose-ui-test-junit4 = "androidx.compose.ui:ui-test-junit4:1.5.4"
+androidx-compose-ui-test-manifest = "androidx.compose.ui:ui-test-manifest:1.5.4"
```
JakeWharton: Please pull out a version entry for these as they are grouped: https://developer.android.com/jetpack/androidx/releases/compose-ui

JakeWharton (approved):

PR #1674, opened 2023-11-08
Make CodeSession a base class

swankjesse:
This splits the Zipline dependency from the lifecycle management. It lets us test the same lifecycle code for the FakeCodeSession.

Also rename CodeSession.cancel() to CodeSession.stop().


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeHost.kt
```diff
@@ -60,12 +60,12 @@ internal abstract class CodeHost<A : AppService>(
     override fun onUncaughtException(codeSession: CodeSession<A>, exception: Throwable) {
     }
 
-    override fun onCancel(codeSession: CodeSession<A>) {
+    override fun onStop(codeSession: CodeSession<A>) {
```
veyndan, line 63: I think the rename could've been a separate PR! I feel like it obfuscates the core of the PR (i.e., making `CodeSession` a base class).

veyndan (approved):

PR #1675, opened 2023-11-09
Hoist state changing of `screen`

veyndan:
Per [the documentation](https://developer.android.com/jetpack/compose/state#state-hoisting), It's bad practice to pass `MutableState`'s down as parameters. Per their guidance, the parameter has been split between two: `screen` and `onScreenChange`.

### DIFF
--- a/test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
+++ b/test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
@@ -16,10 +16,11 @@
 package com.example.redwood.testing.presenter
 
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.MutableState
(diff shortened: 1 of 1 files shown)


test-app/presenter/src/commonMain/kotlin/com/example/redwood/testing/presenter/TestApp.kt
```diff
@@ -44,12 +45,11 @@ class TestContext(
 
 @Composable
 fun TestApp(context: TestContext) {
-  val screenKeyState = rememberSaveable { mutableStateOf<String?>(null) }
-  val screenKey = screenKeyState.value
+  var screenKey by rememberSaveable { mutableStateOf<String?>(null) }
```
veyndan, line 48: Using property delegation to reduce the risk of this happening again.

JakeWharton (approved):

PR #1677, opened 2023-11-09
Propagate `BackHandler#enabled` changing in Treehouse

### LINKED ISSUE #1672: `BackHandler` doesn't propagate `enabled` changing

veyndan:
Closes https://github.com/cashapp/redwood/issues/1672.

### DIFF
--- a/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
+++ b/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
@@ -83,8 +83,8 @@ private class RedwoodZiplineTreehouseUi(
       }
 
       override fun addOnBackPressedCallback(
-        callback: OnBackPressedCallbackService,
-      ): CancellableService = onBackPressedDispatcher.addCallback(callback)
+        onBackPressedCallbackService: OnBackPressedCallbackService,
+      ): CancellableService = onBackPressedDispatcher.addCallback(onBackPressedCallbackService)
     }
 
     start(host)
--- a/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
+++ b/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
@@ -27,6 +27,7 @@ import app.cash.redwood.ui.OnBackPressedDispatcher
 import app.cash.redwood.ui.UiConfiguration
 import app.cash.zipline.ZiplineScope
 import app.cash.zipline.ZiplineScoped
+import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.plus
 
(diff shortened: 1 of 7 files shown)


redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
```diff
@@ -141,9 +144,22 @@ private fun OnBackPressedCallback.asService() = object : OnBackPressedCallbackSe
       this@asService.isEnabled = value
     }
 
+  private val _enabledChangedEvents: Channel<Unit> = Channel()
```
JakeWharton: I think this should probably be `CONFLATED` so that `trySend` always succeeds. As is it will fail, but that's okay because we basically are getting conflated semantics. But if we make the channel conflated it's a bit more self-documenting as to the fact that we're okay with this behavior.

redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeZiplineTreehouseUi.kt
```diff
@@ -64,6 +67,8 @@ class FakeZiplineTreehouseUi(
     val result = host.addOnBackPressedCallback(object : OnBackPressedCallbackService {
       override var isEnabled = isEnabled
 
+      override val enabledChangedEvents: Flow<Unit> = emptyFlow()
+
```
veyndan: I want to actually test the toggling of `isEnabled` and make sure everything propagates correctly. However, I don't want to copy [the production code logic](https://github.com/cashapp/redwood/blob/d6bf86806629727b5f7f0f80b045a8c5faf433dd/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt#L140-L163) to this fake, as I want to be testing the production code and not a fake.

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/OnBackPressedCallbackService.kt
```diff
@@ -18,11 +18,14 @@ package app.cash.redwood.treehouse
...
 import app.cash.zipline.ZiplineService
 import kotlin.native.ObjCName
+import kotlinx.coroutines.flow.Flow
 
 /** Redwood's [OnBackPressedCallback] but implementing [ZiplineService]. */
 @ObjCName("OnBackPressedCallbackService", exact = true)
 public interface OnBackPressedCallbackService : ZiplineService {
   public var isEnabled: Boolean
 
+  public val enabledChangedEvents: Flow<Unit>
```
JakeWharton: Is there a reason not to go with `StateFlow<Boolean>`?

JakeWharton (approved):

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/OnBackPressedCallbackService.kt
veyndan: Nope! I was just blinded by getting `OnBackPressedCallbackService` to mimic `OnBackPressedCallback` that I ended up using a `Channel`. Converted to a `StateFlow` and its looking a lot better!

PR #1678, opened 2023-11-13
Rename some functions in EventListener

swankjesse:
The 'on' prefix is apporpriate for Redwood's ProtocolMismatchHandler, which must make policy decisions. In Treehouse we've made those policy decisions already (ignore mismatches!) and are merely relaying our decisions as events for observability.

### DIFF
--- a/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
+++ b/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
@@ -84,48 +84,61 @@ public open class EventListener {
   }
 
   /**
-   * Invoked on a request to create an unknown widget [kind].
+   * Invoked on a request to create an unknown widget with [tag]. This is a schema mismatch and the
+   * widget is ignored.
    */
-  public open fun onUnknownWidget(
+  public open fun unknownWidget(
     tag: WidgetTag,
   ) {
   }
 
   /**
-   * Invoked on a request to create an unknown modifier [tag].
+   * Invoked on a request to create an unknown modifier with [tag]. This is a schema mismatch and
+   * the modifier is ignored.
    */
-  public open fun onUnknownModifier(
+  public open fun unknownModifier(
     tag: ModifierTag,
   ) {
   }
 
   /**
-   * Invoked on a request to manipulate unknown children [tag] for the specified widget [kind].
+   * Invoked on a request to manipulate unknown children with [tag] for a widget with [widgetTag].
+   * This is a schema mismatch and the child nodes are ignored.
    */
-  public open fun onUnknownChildren(
+  public open fun unknownChildren(
     widgetTag: WidgetTag,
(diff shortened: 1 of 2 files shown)


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -84,48 +84,61 @@ public open class EventListener {
...
-  public open fun onUnknownModifier(
+  public open fun unknownModifier(
     tag: ModifierTag,
   ) {
   }
 
   /**
-   * Invoked on a request to manipulate unknown children [tag] for the specified widget [kind].
+   * Invoked on a request to manipulate unknown children with [tag] for a widget with [widgetTag].
+   * This is a schema mismatch and the child nodes are ignored.
```
swankjesse, line 106: @JakeWharton is doing nothing here safe?

If it isn’t, maybe we make a synthetic parent node? Ideally we don’t get further events for child nodes of the unknown child. My policy preference is to ignore the unknown children!

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
```diff
@@ -84,48 +84,61 @@ public open class EventListener {
   }
 
   /**
-   * Invoked on a request to create an unknown widget [kind].
+   * Invoked on a request to create an unknown widget with [tag]. This is a schema mismatch and the
+   * widget is ignored.
```
swankjesse, line 88: @JakeWharton is doing nothing here safe?

JakeWharton: Hmm I didn't write these so I'm not quite sure how this gets invoked. From what I would guess, the decision was already made at the Redwood layer to return `null` (i.e., ignore the widget) and here you can either log or escalate that to a throw.

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
JakeWharton: There is no way to do that today. There is no "create" step for children so this event would happen for any child create/move/removes.

Generally we don't have anything in place to remember an ignore of anything. So if you ignore creation of a widget with a certain ID you'll still get failures for every property change on that widget because the ID was never put into the map.

JakeWharton (approved):

redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/EventListener.kt
swankjesse: Yeah, this may require some work. Tracking here: https://github.com/cashapp/redwood/issues/1679

PR #1680, opened 2023-11-14
Replace deprecated `buildDir` method

veyndan:
https://docs.gradle.org/current/userguide/upgrading_version_8.html#project_builddir


JakeWharton (approved):
So convoluted, Gradle!

PR #1687, opened 2023-11-17
Test layout content changes

swankjesse:
We were missing tests for items being removed from Column, etc.

### DIFF
--- a/redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testDynamicElementUpdates[LTR]_abcde.png
+++ b/redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testDynamicElementUpdates[LTR]_abcde.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:84c70ea7306c335b4f5ef372d5e49b4ce71e4800660f3a9e40613681ff572948
+size 5307
--- a/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
+++ b/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
@@ -415,6 +415,28 @@ abstract class AbstractFlexContainerTest<T : Any> {
     verifySnapshot(column)
   }
 
+  @Test
+  fun testDynamicElementUpdates() {
+    val container = flexContainer(FlexDirection.Column)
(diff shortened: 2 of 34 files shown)


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -429,6 +451,8 @@ interface TestFlexContainer<T : Any> : Widget<T>, ChangeListener {
   fun mainAxisAlignment(mainAxisAlignment: MainAxisAlignment)
   fun margin(margin: Margin)
   fun add(widget: Widget<T>)
+  fun addAt(index: Int, widget: Widget<T>)
```
swankjesse, line 454: The first thing I tried was exposing the `Widget.Children` list in this facet. That doesn’t work so well 'cause LazyList doesn’t know its own count until `onEndChanges()` is called.

redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testDynamicElementUpdates[LTR]_abcde.png
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
Comment (JakeWharton) [inline redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testDynamicElementUpdates[LTR]_abcde.png:1]: Why do these have lowercase suffixes? Did you start with lowercase and then switch to uppercase on a case-insensitive file system? The iOS ones have uppercase suffixes like I would expect.

Makes me wonder if we can force CI onto a case sensitive FS on Mac...

JakeWharton (approved):
Can't see the images on mobile app, but I'll trust they're right.

redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testDynamicElementUpdates[LTR]_abcde.png
Comment (swankjesse) [reply redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testDynamicElementUpdates[LTR]_abcde.png:1]: Good catch, yeah that’s super weird. I started uppercase and I’m using a case-sensitive file system. I wonder if Paparazzi does this?!

Comment (swankjesse) [reply redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testDynamicElementUpdates[LTR]_abcde.png:1]: Confirmed.
https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/Snapshot.kt#L36

PR #1688, opened 2023-11-17
 Fix column content changes

swankjesse:
We had a bug where changes to a column's content wouldn't trigger all the necessary layout updates.

### DIFF
--- a/redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
+++ b/redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
@@ -21,10 +21,6 @@ internal class YogaLayout(
 ) : ViewGroup(context) {
   val rootNode = Node()
 
-  init {
-    applyLayoutParams(rootNode, layoutParams)
-  }
-
   private fun applyLayout(node: Node, xOffset: Float, yOffset: Float) {
     val view = node.view
     if (view != null && view !== this) {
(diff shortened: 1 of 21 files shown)


redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_testDynamicContainerSize[LTR]_single.png
Comment (swankjesse) [inline redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewFlexContainerTest_testDynamicContainerSize[LTR]_single.png:1]: This one was broken before the code change. It showed the `B` at 50%, not at the bottom.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
```diff
@@ -71,6 +67,11 @@ internal class YogaLayout(
   }
 
   private fun calculateLayout(widthSpec: Int, heightSpec: Int) {
+    rootNode.requestedWidth = Size.Undefined
+    rootNode.requestedMaxWidth = Size.Undefined
+    rootNode.requestedHeight = Size.Undefined
+    rootNode.requestedMaxHeight = Size.Undefined
+
```
veyndan, line 74: Is there any downside to not specifying the requested sizes?

redwood-lazylayout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiLazyListTest_testDynamicContainerSize[LTR]_both.png
Comment (veyndan) [inline redwood-lazylayout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiLazyListTest_testDynamicContainerSize[LTR]_both.png:1]: We should probably create an Issue to have these work with Compose UI so we don't forget about it

veyndan (approved):

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
swankjesse: This just clears them out so that we can set the ones we want below. Otherwise it ends up accumulating both exact and max values when it’s repeatedly measured.

redwood-lazylayout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiLazyListTest_testDynamicContainerSize[LTR]_both.png
Comment (swankjesse) [reply redwood-lazylayout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiLazyListTest_testDynamicContainerSize[LTR]_both.png:1]: I can do better than that. I can fix the problem!
https://github.com/cashapp/redwood/pull/1691

Issue #1696, opened 2023-11-22
Android View-based Emoji Search sample crashes because "no more placeholders!"


dellisd:
Stacktrace:

```
FATAL EXCEPTION: main
Process: com.example.redwood.emojisearch.android.views, PID: 26278
java.lang.IllegalStateException: no more placeholders!
	at app.cash.redwood.lazylayout.widget.LazyListUpdateProcessor.takePlaceholder(LazyListUpdateProcessor.kt:317)
	at app.cash.redwood.lazylayout.widget.LazyListUpdateProcessor.requireBindingAtIndex$createBinding(LazyListUpdateProcessor.kt:305)
	at app.cash.redwood.lazylayout.widget.LazyListUpdateProcessor.requireBindingAtIndex(LazyListUpdateProcessor.kt:311)
	at app.cash.redwood.lazylayout.widget.LazyListUpdateProcessor.bind(LazyListUpdateProcessor.kt:291)
	at app.cash.redwood.lazylayout.view.ViewLazyList$LazyContentItemListAdapter.onBindViewHolder(ViewLazyList.kt:255)
	at app.cash.redwood.lazylayout.view.ViewLazyList$LazyContentItemListAdapter.onBindViewHolder(ViewLazyList.kt:225)
	at androidx.recyclerview.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:7747)
	at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:7847)
	at androidx.recyclerview.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:6646)
	at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:6917)
	at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6757)
	at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6753)
	at androidx.recyclerview.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:2362)
	at androidx.recyclerview.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1662)
	at androidx.recyclerview.widget.LinearLayoutManager.fill(LinearLayoutManager.java:1622)
[...]

PR #1697, opened 2023-11-23
Add UI tests for Emoji Search Android frontends

JakeWharton:
The view-based sample is currently crashing, which is why it's ignored.

Refs #537. Refs #1696.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -134,6 +134,12 @@ jobs:
 
       - run: ./gradlew -p samples/emoji-search build
 
+      - uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 29
+          emulator-boot-timeout: 20000
+          script: ./gradlew -p samples/emoji-search connectedCheck
+
       - name: Build Emoji Search iOS (UIKit)
         run: xcodebuild -project samples/emoji-search/ios-uikit/EmojiSearchApp.xcodeproj -scheme EmojiSearchApp -destination 'platform=iOS Simulator,name=iPhone 12,OS=latest'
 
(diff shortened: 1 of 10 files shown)


samples/emoji-search/android-tests/src/main/kotlin/com/example/redwood/emojisearch/android/tests/AbstractEmojiSearchUiTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+  @Test fun searchTrees() {
+    awaitText("0. +1")
+    search.text = "tree"
+    awaitText("301. christmas_tree")
+    search.clear()
+    awaitText("0. +1")
+  }
+
+  private fun awaitText(value: String, duration: Duration = 1.seconds) {
```
JakeWharton:
```suggestion
  private fun awaitText(value: String, duration: Duration = 4.seconds) {
```

.github/workflows/build.yaml
```diff
@@ -134,6 +134,12 @@ jobs:
 
       - run: ./gradlew -p samples/emoji-search build
 
+      - uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 29
+          emulator-boot-timeout: 20000
```
dellisd, line 140: Have we ever really figured out why the emulator boot would timeout in SQLDelight PR checks?
Hopefully this doesn't introduce flakiness.

dellisd (approved):

.github/workflows/build.yaml
JakeWharton: We have two other jobs that already have emulators. They don't fail that often. Maybe one out of every 30 PRs. Not great, but not terrible either.

PR #1706, opened 2023-11-29
Fix bug replacing of root widget on UIKit

Alex009:
Hello!

We find bug on iOS with UIKit and redwood. Case:
create uikit with root `UIStackView` - just like in CounterApp sample:
```
let container = UIStackView()
container.axis = .horizontal
container.alignment = .fill
container.distribution = .fillEqually
container.translatesAutoresizingMaskIntoConstraints = false

view.addSubview(container)
container.leadingAnchor.constraint(equalTo: view.leadingAnchor).isActive = true
container.topAnchor.constraint(equalTo: view.topAnchor).isActive = true
container.widthAnchor.constraint(equalTo: view.widthAnchor).isActive = true
container.heightAnchor.constraint(equalTo: view.heightAnchor).isActive = true

self.delegate = CounterViewControllerDelegate(root: container)
```

in delegate use `RedwoodUIView(root)`:
```
val composition = RedwoodComposition(
  scope = scope,
  view = RedwoodUIView(root),
  provider = SchemaWidgetFactories(
    Schema = IosWidgetFactory,
    RedwoodLayout = UIViewRedwoodLayoutWidgetFactory(),
  ),
)
```

but in Composable of our screen we should change root view by some condition (for example we change loading state to successful):
```
@Composable
fun Counter(value: Int = 0) {
  var count by rememberSaveable { mutableStateOf(value) }

  if (count == 3) {
    Text("this is end")
    return
  }
  
  Column(
    width = Constraint.Fill,
    height = Constraint.Fill,
    horizontalAlignment = CrossAxisAlignment.Center,
  ) {
    Button("-1", onClick = { count-- })
    Text("Count: $count")
    Button("+1", onClick = { count++ })
  }
}
```

but istead of change root uiview we see that no changes.
reason - UIStackView add new child in arrangedSubviews (at 0 index) but remove child from subviews (that have different order). so when new UILabel added this label will be removed, instead of remove ColumnUIView

### DIFF
--- a/redwood-widget-testing/src/commonMain/kotlin/app/cash/redwood/widget/testing/AbstractWidgetChildrenTest.kt
+++ b/redwood-widget-testing/src/commonMain/kotlin/app/cash/redwood/widget/testing/AbstractWidgetChildrenTest.kt
@@ -140,6 +140,16 @@ public abstract class AbstractWidgetChildrenTest<W : Any> {
     assertThat(names()).containsExactly("one", "two", "three")
   }
 
+  @Test public fun replace() {
+    // Models what happens when a conditional flips from the second branch to the first.
+    // The new item will be added and then the old item will be removed.
+    // From https://github.com/cashapp/redwood/pull/1706.
+    children.insert(0, widget("one"))
+    children.insert(0, widget("new one"))
+    children.remove(1, 1)
+    assertThat(names()).containsExactly("new one")
+  }
+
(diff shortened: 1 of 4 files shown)


JakeWharton:
Thanks! I'd like to get a test case in for this somewhere. Probably another subclass of the abstract children test.

JakeWharton (approved):

PR #1709, opened 2023-11-29
AndroidX Compose compiler 1.5.5

JakeWharton:


colinrtwhite (approved):
Do we still need the androidx compiler or could we migrate fully to Jetbrains?

PR #1713, opened 2023-11-30
Make `TreehouseAppContent#preload` idempotent

veyndan:
Both `TreehouseAppContent#bind` and `TreehouseAppContent#unbind` are idempotent. We should make `TreehouseAppContent#preload` idempotent for consistency.


JakeWharton (approved):

JakeWharton:
I was going to ask what happens if the UI config changes during preload but the next PR addresses that.

PR #1718, opened 2023-12-01
Upgrade to Zipline 1.7.0

swankjesse:
The '@Contextual' is both unfortunate ane necessary. There's no automatic way for Zipline to customize encoding of built-in types, and the built-in encoding of longs loses precision.

More details in this Zipline PR:
https://github.com/cashapp/zipline/pull/1171

### DIFF
--- a/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppLifecycle.kt
+++ b/redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppLifecycle.kt
@@ -20,12 +20,13 @@ import app.cash.redwood.protocol.Id
 import app.cash.redwood.protocol.WidgetTag
 import app.cash.zipline.ZiplineService
 import kotlin.native.ObjCName
+import kotlinx.serialization.Contextual
 
 @ObjCName("AppLifecycle", exact = true)
 public interface AppLifecycle : ZiplineService {
   public fun start(host: Host)
 
-  public fun sendFrame(timeNanos: Long)
+  public fun sendFrame(timeNanos: @Contextual Long)
 
   /** Platform features to the guest application. */
   public interface Host : ZiplineService {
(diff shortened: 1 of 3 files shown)


redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppLifecycle.kt
```diff
@@ -20,12 +20,13 @@ import app.cash.redwood.protocol.Id
...
 import app.cash.zipline.ZiplineService
 import kotlin.native.ObjCName
+import kotlinx.serialization.Contextual
 
 @ObjCName("AppLifecycle", exact = true)
 public interface AppLifecycle : ZiplineService {
   public fun start(host: Host)
 
-  public fun sendFrame(timeNanos: Long)
+  public fun sendFrame(timeNanos: @Contextual Long)
```
veyndan, line 29: This is a side point, but I feel like we've been overusing `@Contextual`. We very explicitly know what serializer we want to use here, and should probably opt for something as outlined at https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#specifying-serializer-globally-using-typealias. Doing `@Contextual` just pushes all unregistered types to throw a runtime exception, which doesn't make sense if we know what serializer we always want for a specific case.

veyndan (approved):

redwood-treehouse/src/commonMain/kotlin/app/cash/redwood/treehouse/AppLifecycle.kt
swankjesse: Great idea.

swankjesse: https://github.com/cashapp/redwood/issues/1723

PR #1724, opened 2023-12-05
Fix the flex property on iOS not being respected.

colinrtwhite:
Follow up to https://github.com/cashapp/redwood/pull/1699. This PR fixes `Modifier.flex` on iOS.

The root cause was `scrollEnabled` defaulting to `true` and the `YogaLayout`'s first layout pass would occur before it was set to `false` by Redwood. That first pass would set each child's frame such that it would affect the later measurement passes. `flex` relies on `flexBasis`, which has no effect when the main axis size is undefined (like when measuring a row with scroll enabled).

This PR also includes some minor formatting changes.


redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -31,10 +31,11 @@ internal class YogaUIView(
   var height = Constraint.Wrap
 
   init {
-    // TODO: Support Scroll Indicators
+    // TODO: Support scroll indicators.
+    scrollEnabled = false
```
colinrtwhite, line 35: This is the one-line fix.

JakeWharton (approved):

PR #1728, opened 2023-12-06
Make `TreehouseApp`'s `ziplineFlow` a property

veyndan:


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
veyndan, line 1: If you want a better diff here, you can step through the last two commits.

veyndan:
I didn't fully test this, and it doesn't appear to work. I'll close it out and rethink it.

PR #1729, opened 2023-12-06
Add a test to confirm there's no unnecessary recompositions

swankjesse:
Also expose the size of the preload window in
LazyListState. It should be possible for the
application layer to change this as necessary.


redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/LazyListState.kt
```diff
@@ -59,6 +61,12 @@ public open class LazyListState {
   public var lastIndex: Int by mutableStateOf(0)
     private set
 
+  /** How many items to load in anticipation of scrolling up. */
+  public var preloadBeforeItemCount: Int by mutableStateOf(DEFAULT_PRELOAD_ITEM_COUNT)
```
veyndan, line 65: Nice

veyndan (approved):

dellisd (approved):
👏

PR #1730, opened 2023-12-06
Support setting the background color for widgets in flex container tests.

colinrtwhite:
Using this property in a subsequent test. The screenshot changes are due to better visibility of overlapping row/column widgets.


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/utils.common.kt
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@file:Suppress("ktlint:standard:function-naming", "ktlint:standard:property-naming")
+
+/*
```
JakeWharton: This file has two headers

JakeWharton (approved):

PR #1735, opened 2023-12-08
Install a global uncaught exception handler

swankjesse:


redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/StandardAppLifecycle.kt
```diff
@@ -60,9 +61,14 @@ public class StandardAppLifecycle(
   internal val coroutineScope = CoroutineScope(coroutineExceptionHandler)
 
   override fun start(host: Host) {
+    check(!started) { "already started" }
```
jingwei99, line 64: idea to serve as an input: no-op / return directly if the app already started vs. throw an exception, it makes it potentially less painful when downstream is implemented with flickering state (like session manager 😅)

swankjesse: I like to fail fast by default, and relax as necessary later. It reduces the number of different states this thing can be in.

In this case I don’t want to think about what happens if I get two different `Host` instances. That would suck.

jingwei99: > I don’t want to think about what happens if I get two different Host instances. That would suck

that's really true! (so downstream is responsible to manage the/their life cycle correctly)

jingwei99 (approved):

PR #1736, opened 2023-12-08
Remove @Ignore since the fix is in for #1696

jingwei99:
Follow up item from https://github.com/cashapp/redwood/issues/1696#issuecomment-1823617752

Fix: https://github.com/cashapp/redwood/commit/63ddcf041f16cb510188bb2f64cb8cb9848de3ce


JakeWharton (approved):
Import needs removed

Issue #322, opened 2022-08-24
FakeTreehouseApp


swankjesse:
Let’s make a fake for TreehouseApp that doesn’t use Zipline. 

Perhaps it just has a `var app: T` that represents that value passed into `TreehouseView.Content..get()`.

That should make it easier to write tests that use TreehouseViews without having a dev server or production server nearby.

PR #1737, opened 2023-12-11
Make TreehouseApp an interface

jingwei99:
This makes it easier to mock `TreehouseApp` for testing in downstream repo(s).

### DIFF
--- a/redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/treehouseAppFactoryAndroid.kt
+++ b/redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/treehouseAppFactoryAndroid.kt
@@ -34,7 +34,7 @@ public fun TreehouseAppFactory(
   cacheMaxSizeInBytes: Long = 50L * 1024L * 1024L,
   concurrentDownloads: Int = 8,
   stateStore: StateStore = MemoryStateStore(),
-): TreehouseApp.Factory = TreehouseApp.Factory(
+): TreehouseApp.Factory = RealTreehouseApp.Factory(
   platform = AndroidTreehousePlatform(context),
   dispatchers = AndroidTreehouseDispatchers(),
   eventListenerFactory = eventListenerFactory,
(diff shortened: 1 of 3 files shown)


JakeWharton:
I'm pretty opposed to the idea that someone could effectively mock this interface and provide suitable behavior. Mocking service types that you don't own is a pretty strong indicator of a broken test that is encoding implementation details in the test ([change detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html)) rather than observing visible effects.

I'd be interested in what is being tested and what the goal of the test actually is. That should tell us whether we need to provide a fake (#322) or whether the test should be running against a real implementation instead.

jingwei99:
> I'm pretty opposed to the idea that someone could effectively mock this interface and provide suitable behavior. Mocking service types that you don't own is a pretty strong indicator of a broken test that is encoding implementation details in the test ([change detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html)) rather than observing visible effects.
> 
> I'd be interested in what is being tested and what the goal of the test actually is. That should tell us whether we need to provide a fake (#322) or whether the test should be running against a real implementation instead.

I ran into a case where a class depends on (has an injected) `TreehouseApp` needs to be tested in `cash-android`. In that use case, a fake will do / would be sufficient.

JakeWharton:
I'm also skeptical as to whether a fake makes sense for this type in general (we've gone a long time without one!).

I'd be curious what the test is doing and what it expects from the fake as opposed to a real implementation on the JVM. But I guess I'll have to wait and someone else can weigh in on a near term solution to unblock.

jingwei99:
> I'm also skeptical as to whether a fake makes sense for this type in general (we've gone a long time without one!).
> 
> I'd be curious what the test is doing and what it expects from the fake as opposed to a real implementation on the JVM. But I guess I'll have to wait and someone else can weigh in on a near term solution to unblock.

I left a DM to you (I know you are not checking DMs while you are out 🥲). I think for now a fake TreehouseApp will do, will close this PR and try to add a fake.

colinrtwhite:
@jingwei99 Probably a good discussion topic for today's treehouse meeting later today!

jingwei99:
chatted with @colinrtwhite during the meeting today, since this not urgent*, I'll close this PR, and we'll chat about it when @JakeWharton is back. 

not urgent: I've found a walk-around for the need to have a fake, so this is more of a follow-up to be proactive to not to get to the same place again, other than something blocking and needs an urgent change.

PR #1739, opened 2023-12-12
Add basic layout system documentation.

colinrtwhite:
Going to follow up with visual documentation for modifiers.

### DIFF
--- a/redwood-layout-schema/README.md
+++ b/redwood-layout-schema/README.md
@@ -13,7 +13,7 @@ Internally the layout system uses a common layout engine written in Kotlin Multi
 
 ### Row
 
-Lays widgets out along the X axis (horizontally). 
+Lays widgets out along the X axis (horizontally).
 
 ```kotlin
 Row {
@@ -29,7 +29,7 @@ Row {
 
 ### Column
 
-Lays widgets out along the Y axis (vertically). 
+Lays widgets out along the Y axis (vertically).
 
 ```kotlin
 Column {
@@ -65,7 +65,7 @@ Box {
 ```
 
 <p style="text-align: center;">
-    <img src="docs_images/box1.png">
+    <img src="docs_images/box1.png" width="250" height="250">
 </p>
 
 ### Spacer
--- a/redwood-layout-schema/README.md
+++ b/redwood-layout-schema/README.md
@@ -24,7 +24,7 @@ Row {
 ```
 
 <p style="text-align: center;">
-    <img src="../docs_images/row1.png">
+    <img src="docs_images/row1.png">
 </p>
 
 ### Column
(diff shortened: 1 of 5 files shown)


redwood-layout-schema/README.md
```diff
@@ -0,0 +1,85 @@
+# Redwood Layout
+
+Redwood's layout system provides common layout primitives to support arranging and laying out widgets. It provides widgets for:
+
+- `Row`: Lays wigets out along the X axis (horizontally).
+- `Column`: Lays widgets out along the Y axis (vertically).
+- `Box`: Lays widgets out along the Z axis (on top of eachother).
+- `Spacer`: Adds space between widgets.
+
+Internally the layout system uses a common layout engine written in Kotlin Multiplatform to support consistent rendering across platforms. The system provides widget bindings for for Android Views (`redwood-layout-view`), iOS UiKit (`redwood-layout-uiview`), and Compose UI (`redwood-layout-composeui`).
```
JakeWharton: No DOM mention?

redwood-layout-schema/README.md
```diff
@@ -0,0 +1,85 @@
...
+</p>
+
+### Spacer
+
+Adds space between widgets.
+
+```kotlin
+Column {
+    Text("Top")
+    Spacer(height = 24.dp)
```
JakeWharton: Indent seems to be 4 spaces throughout. Should be 2, but I would think the editorconfig set that correctly...

redwood-layout-schema/docs_images/box1.png
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
JakeWharton, line 1: These are in LFS. Do we want to change our rules so only screenshot tests go into LFS? Or are we okay sending all pngs?

JakeWharton (approved):

redwood-layout-schema/README.md
colinrtwhite: Ah yep - wrote this up in VS Code so the editorconfig didn't apply.

redwood-layout-schema/docs_images/box1.png
colinrtwhite: Hmm I don't have a strong opinion. I imagine the docs are going to either be viewed on github.com, which correctly renders the images inline, or on the published docs site, which will host the images. Is there a downside to storing these with LFS?

JakeWharton: Mostly just philosophical. It breaks the concept of a self-contained repo. The screenshot test goldens can be refreshed by running the tests. These are more like part of the codebase.

colinrtwhite: @JakeWharton Good point, I'll clean up the code that generates these screenshots and add it to the repo.

colinrtwhite: Good point, I'll clean up the code that generates these screenshots and add it to the repo.

PR #1755, opened 2024-01-02
Support the Emoji Search sample on desktop.

colinrtwhite:
It's hooked up directly the presenter on desktop similar to the counter sample.

https://github.com/cashapp/redwood/assets/5580160/882dadfb-9d17-451e-90c6-7aed0af443a0


samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchApplication.kt
```diff
@@ -0,0 +1,34 @@
...
+import coil3.SingletonImageLoader
+import coil3.annotation.ExperimentalCoilApi
+import coil3.fetch.NetworkFetcher
+
+class EmojiSearchApplication : Application(), SingletonImageLoader.Factory {
+  @OptIn(ExperimentalCoilApi::class)
+  override fun newImageLoader(context: Context): ImageLoader {
+    return ImageLoader.Builder(context)
+      .components {
+        add(NetworkFetcher.Factory())
```
colinrtwhite: Network URL support currently has to be added manually as it was pulled out of `coil-core`. Ktor finds imported engines at runtime, which could also be done by Coil, but I'm not sure it's a good idea.

JakeWharton:
> It's hooked up directly the presenter on desktop similar to the counter sample.

I'll look more into the PR later, but why was this done? The emoji sample is meant to be the Treehouse one. Is there something we can refactor towards to ensure it remains so that isn't available today?

gradle/libs.versions.toml
```diff
@@ -7,8 +7,9 @@ androidx-compose-ui = "1.5.4"
 jbCompose = "1.5.11"
 lint = "31.2.0"
 zipline = "1.7.0"
-coil = "2.5.0"
+coil = "3.0.0-alpha01"
```
JakeWharton: Probably worth landing this by itself first

samples/emoji-search/composeui/build.gradle
```diff
@@ -0,0 +1,79 @@
...
+  desktop {
+    application {
+      mainClass = "com.example.redwood.emojisearch.composeui.Main"
+    }
+  }
+}
+
+kotlin {
+  androidTarget()
+  jvm("desktop")
```
JakeWharton: Not sure how I feel about these being in the same module. In the (admittedly very small) audience of those interested in Redwood and Treehouse, I expect the overwhelming majority to care about Android and not JVM Desktop. That's why I kept the counter frontends as separate modules so that you don't get lost in the multiplatform stuff around it. I would prefer to do the same here, unless there's some other strong reason to merge them.

samples/emoji-search/composeui/src/commonMain/kotlin/com/example/redwood/emojisearch/composeui/ComposeUiText.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Square, Inc.
+ * Copyright (C) 2024 Square, Inc.
```
JakeWharton: All these dates need reverted. An annoying behavior of the move action of IDEA.

JakeWharton (commented):
Excited for this

samples/emoji-search/composeui/build.gradle
colinrtwhite: Sounds good I'll separate them 👍🏻

colinrtwhite:
> > It's hooked up directly the presenter on desktop similar to the counter sample.
> 
> I'll look more into the PR later, but why was this done? The emoji sample is meant to be the Treehouse one. Is there something we can refactor towards to ensure it remains so that isn't available today?

@JakeWharton I think we need a JVM version of `FrameClock` to use Zipline + Treehouse. Do you know if we've implemented something similar elsewhere?

colinrtwhite:
> > It's hooked up directly the presenter on desktop similar to the counter sample.
> 
> I'll look more into the PR later, but why was this done? The emoji sample is meant to be the Treehouse one. Is there something we can refactor towards to ensure it remains so that isn't available today?

@JakeWharton I think we need a JVM instance of `FrameClock` to support Zipline + Treehouse. Do you think it's possible to write a common implementation that wraps Compose Multiplatform's `MonotonicFrameClock`? Or should I go another route?

JakeWharton:
> @JakeWharton I think we need a JVM instance of `FrameClock` to support Zipline + Treehouse. Do you think it's possible to write a common implementation that wraps Compose Multiplatform's `MonotonicFrameClock`? Or should I go another route?

Almost certainly. It would go in a 'treehouse-host-composeui' module (if we don't already have one). It should really happen automatically as part of using `TreehouseContent`, although I honestly can't remember if we built that yet or not.

If you refactor to keep the Android and JVM Desktop frontends in separate modules, I'd be okay with merging the direct usage and working towards getting it working on Treehouse over the coming weeks (or after I get back).

samples/emoji-search/composeui/src/commonMain/kotlin/com/example/redwood/emojisearch/composeui/RealHostApi.kt
```diff
@@ -67,8 +63,6 @@ class RealHostApi(
   }
 
   override fun openUrl(url: String) {
-    val intent = Intent(Intent.ACTION_VIEW)
-    intent.setData(Uri.parse(url))
-    startActivity(context, intent, null)
+    openUrl.invoke(url)
```
jingwei99: 👍

jingwei99 (approved):
this is really cool!

(and I saw [the pending comment](https://github.com/cashapp/redwood/pull/1755/files#r1439651088) 👍 )

JakeWharton:
Screenshot 2024-02-22 at 2 48 47 PM

PR #1767, opened 2024-01-10
Don't crash if we exhaust placeholders

swankjesse:
Instead just show an empty placeholder that's sized by the first placeholder in the pool.

Also change the placeholder pool size from 30 down to 20. It was 30 because we didn't have a solution to this problem, now we do.


jingwei99 (approved):

veyndan (commented):
I'm happy with this as an interim solution, but I think we should keep #1500 open. We have (or had?) a mechanism on Android, that whenever the number of views are exhausted, we **re-request** more placeholders. In my mind, this is still the north star, and is what #1500 outlines.

swankjesse:
Works for me. BTW, we can implement that exclusively in guest code.

PR #1768, opened 2024-01-10
Update to coil 3.0.0-alpha02.

colinrtwhite:


samples/emoji-search/android-composeui/src/main/kotlin/com/example/redwood/emojisearch/android/composeui/EmojiSearchApplication.kt
```diff
@@ -1,32 +0,0 @@
...
-import coil3.ImageLoader
-import coil3.PlatformContext
-import coil3.SingletonImageLoader
-import coil3.fetch.NetworkFetcher
-
-class EmojiSearchApplication : Application(), SingletonImageLoader.Factory {
-  override fun newImageLoader(context: PlatformContext): ImageLoader {
-    return ImageLoader.Builder(context)
-      .components {
-        add(NetworkFetcher.Factory())
```
colinrtwhite, line 28: As of alpha02 Coil adds this by default if it's on the classpath. It uses the same mechanism that Ktor uses to detect engines. Can be opted out of using `ImageLoader.Builder.serviceLocatorEnabled(false)`.

JakeWharton: Nice. In a real app I would load it manually still, but this is fine for the samples.

JakeWharton (approved):

samples/emoji-search/android-composeui/src/main/kotlin/com/example/redwood/emojisearch/android/composeui/EmojiSearchApplication.kt
colinrtwhite: Yep, I was unsure about adding this feature, but it's there if folks want it. Adding components manually always takes precedence over automatically loaded components.

PR #1773, opened 2024-01-16
Document the lazy layout widget definitions

veyndan:
Unlike #1771, which documents the public API of `LazyColumn`/`LazyRow`, this documents the internal widget API, so that the guest-host bridge for lazy list can be better understood. Some properties are identical to the already-documented public facing API, so I've been deliberately light on the documentation here. However, some properties are unique (namely `itemsBefore`, `itemsAfter`, and `items`) so I wanted to document them a bit better.

If anything seems unclear or worthy of expansion, please lmk!

### DIFF
--- a/redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
+++ b/redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
@@ -23,83 +23,26 @@ import app.cash.redwood.schema.Property
 import app.cash.redwood.schema.Widget
 import app.cash.redwood.ui.Margin
 
+/**
+ * The documentation for [LazyList] is a subset of the documentation for [RefreshableLazyList]. In
+ * order to avoid documentation duplication, see [LazyList]. The documentation should be unified
+ * once https://github.com/cashapp/redwood/issues/1084 is implemented and [RefreshableLazyList] is
+ * deprecated and removed.
+ *
+ * @see LazyList
+ */
 @Widget(1)
 public data class LazyList(
-  /**
-   * The layout orientation of the list.
-   */
   @Property(1) val isVertical: Boolean,
-
-  /**
-   * Invoked when the user has scrolled the list, such that the `firstVisibleItemIndex` or the
-   * `lastVisibleItemIndex` has changed. When the user performs a fling, [onViewportChanged] will be
-   * invoked multiple times.
-   *
-   * The `firstVisibleItemIndex` is the index of the first partially visible item within the
-   * viewport. The `lastVisibleItemIndex` is the index of the last partially visible item within the
-   * viewport.
-   */
   @Property(2) val onViewportChanged: (firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) -> Unit,
-
-  /**
-   * The number of un-emitted items before the [items] window.
-   *
-   * @see [items]
-   */
   @Property(3) val itemsBefore: Int,
-
-  /**
-   * The number of un-emitted items after the [items] window.
-   *
-   * @see [items]
-   */
   @Property(4) val itemsAfter: Int,
-
-  /**
-   * Sets whether the list's width will wrap its contents ([Constraint.Wrap]) or match the width of
-   * its parent ([Constraint.Fill]).
-   */
   @Property(5) val width: Constraint,
-
-  /**
-   * Sets whether the list's height will wrap its contents ([Constraint.Wrap]) or match the height
-   * of its parent ([Constraint.Fill]).
-   */
   @Property(6) val height: Constraint,
-
-  /**
-   * Applies margin (space) around the list.
-   */
   @Property(7) val margin: Margin,
-
-  /**
-   * If [isVertical], sets the default horizontal alignment for items in this list. Else, sets the
-   * default vertical alignment for items in this list.
-   */
   @Property(8) val crossAxisAlignment: CrossAxisAlignment,
-
-  /**
-   * The last [ScrollItemIndex] programmatically requested by the user.
-   */
   @Property(9) val scrollItemIndex: ScrollItemIndex,
-
-  /**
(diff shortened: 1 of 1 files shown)


redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
```diff
@@ -25,33 +25,175 @@ import app.cash.redwood.ui.Margin
 
 @Widget(1)
 public data class LazyList(
+  /**
+   * The layout orientation of the list.
```
dellisd:
```suggestion
   * Whether the list should be vertically oriented.
```

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
```diff
@@ -25,33 +25,175 @@ import app.cash.redwood.ui.Margin
...
+   * invoked multiple times.
+   *
+   * The `firstVisibleItemIndex` is the index of the first partially visible item within the
+   * viewport. The `lastVisibleItemIndex` is the index of the last partially visible item within the
+   * viewport.
+   */
   @Property(2) val onViewportChanged: (firstVisibleItemIndex: Int, lastVisibleItemIndex: Int) -> Unit,
+
+  /**
+   * The number of un-emitted items before the [items] window.
```
dellisd: I don't think it's clear enough what it means to be "emitted" in this context, especially since the term isn't used anywhere else in these doc comments.

dellisd (approved):
Looks good! Just a couple of comments.

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
```diff
@@ -25,33 +25,175 @@ import app.cash.redwood.ui.Margin
...
   @Property(7) val margin: Margin,
+
+  /**
+   * If [isVertical], sets the default horizontal alignment for items in this list. Else, sets the
+   * default vertical alignment for items in this list.
+   */
   @Property(8) val crossAxisAlignment: CrossAxisAlignment,
+
+  /**
+   * The last [ScrollItemIndex] programmatically requested by the user.
```
swankjesse: Call out that we expect to scroll until this item is at the top of the viewport?

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
```diff
@@ -25,33 +25,175 @@ import app.cash.redwood.ui.Margin
...
+   */
   @Property(9) val scrollItemIndex: ScrollItemIndex,
+
+  /**
+   * A block which describes the content of each placeholder item.
```
swankjesse: Don’t just give me generic docs, tell me what the contract is!

Each child in this collection is an element of the placeholder pool
Placeholders are used interchangeably
Policy will decide how large the placeholder pool is

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
```diff
@@ -25,33 +25,175 @@ import app.cash.redwood.ui.Margin
...
+   * `LazyListScope` specify what the list theoretically consists of. This property specifies what
+   * the list practically consists of, as a function of the current view port. This difference is
+   * what distinguishes the `LazyRow` and `LazyColumn` widgets from their non-lazy counterparts
+   * (`Row` and `Column`).
+   */
   @Children(2) val items: () -> Unit,
 )
 
 @Widget(2)
 public data class RefreshableLazyList(
```
swankjesse, line 50: The amount of code duplication between these two is unfortunate. Doubling up on that duplication by duplicating the docs makes it worse? Would you consider only documenting the refreshable-specific docs?

swankjesse (approved):

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
veyndan: Good point. I'll just document `RefreshableLazyList` and chuck and `@see RefreshableLazyList` on the the `LazyList`.

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
veyndan: This property is better documented in `LazyDsl.kt`. Perhaps I can put an `@see` here to make that clearer.

> Policy will decide how large the placeholder pool is

The placeholder pool size is currently hardcoded to 20. We'll have a policy once https://github.com/cashapp/redwood/issues/1500 is implemented, but in the meantime, we distinctly don't have a policy.

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
veyndan: Good shout! The relationship between the `items` window and emitted items is completely absent. Will add that in.

redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
veyndan: This line was copy-pasted from the [Compose UI docs](https://github.com/androidx/androidx/blob/540188236dea66e1565548ae50da4f6b121b4aba/compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/lazy/LazyList.kt#L60), but I agree that it isn't worded very well. Will use your suggestion!

PR #1775, opened 2024-01-18
Update Kotlin, Compose, and Zipline

JakeWharton:


JakeWharton:
This is an earlier commit from #1755 which previously went green (before the Zipline 1.8 release).

swankjesse:
I got the other one green!
https://github.com/cashapp/redwood/pull/1752

PR #1778, opened 2024-01-19
Scope EventListener.Factory to TreehouseApp

swankjesse:
Previously this was scoped to TreehouseApp.Factory, so we couldn't share a Zipline cache without also sharing an EventListener.Factory. This turns out to be quite annoying if we want to customize behavior in the event listener for each app we launch.


dellisd (approved):
LGTM!

PR #1786, opened 2024-01-26
Fix parsing data objects schema modifiers.

colinrtwhite:
Fixes a `NullPointerException` that's thrown when parsing a `data object`.


JakeWharton (approved):
Yep

PR #1796, opened 2024-02-06
Expose frameClock in StandardAppLifecycle

dellisd:
This makes it feasible to launch compositions that aren't tied to the main UI.

### DIFF
--- a/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/StandardAppLifecycle.kt
+++ b/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/StandardAppLifecycle.kt
@@ -36,8 +36,12 @@ public class StandardAppLifecycle(
   private var started = false
   private lateinit var host: Host
 
-  private lateinit var broadcastFrameClock: BroadcastFrameClock
-  internal lateinit var frameClock: MonotonicFrameClock
+  private val broadcastFrameClock: BroadcastFrameClock = BroadcastFrameClock {
+    if (started) {
+      host.requestFrame()
+    }
+  }
+  public val frameClock: MonotonicFrameClock = broadcastFrameClock
 
   internal val mismatchHandler: ProtocolMismatchHandler = object : ProtocolMismatchHandler {
     override fun onUnknownEvent(widgetTag: WidgetTag, tag: EventTag) {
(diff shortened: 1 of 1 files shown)


redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/StandardAppLifecycle.kt
```diff
@@ -36,8 +36,12 @@ public class StandardAppLifecycle(
...
   private lateinit var host: Host
 
-  private lateinit var broadcastFrameClock: BroadcastFrameClock
-  internal lateinit var frameClock: MonotonicFrameClock
+  private val broadcastFrameClock: BroadcastFrameClock = BroadcastFrameClock {
+    if (started) {
+      host.requestFrame()
+    }
+  }
+  public val frameClock: MonotonicFrameClock = broadcastFrameClock
```
JakeWharton:
```suggestion
  public val frameClock: MonotonicFrameClock get() = broadcastFrameClock
```

JakeWharton (approved):
If they're not tied to the main UI, would you still want them limited by its frame rate? Or should it operate more like Molecule's immediate clock?

dellisd:
For my specific use case it doesn't really matter, but in our current Zipline environment I don't think we have any access to time equivalent to `window.performance.now()` that molecule uses. (QuickJs only very recently added `os.now()`)

I think it would still make sense to have access to a framerate-limited clock in any case.

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/StandardAppLifecycle.kt
```diff
@@ -36,8 +36,12 @@ public class StandardAppLifecycle(
   private var started = false
   private lateinit var host: Host
 
-  private lateinit var broadcastFrameClock: BroadcastFrameClock
-  internal lateinit var frameClock: MonotonicFrameClock
+  private val broadcastFrameClock: BroadcastFrameClock = BroadcastFrameClock {
+    if (started) {
```
swankjesse, line 40: FYI, I confirmed that the host code always calls start() before it does anything else so there’s no hazard of losing a frame here.

The only thing that could do it is the main function that binds the app, which exposes the app lifecycle. Like if you used a composition in your navigation system or something.

PR #1813, opened 2024-02-23
Some Emoji sample cleanups after adding desktop

JakeWharton:
- Fix a bad package name
- Share an `OkHttpClient` instance between HTTP client and image loader
- Explicitly pass in an image loader


colinrtwhite (approved):
We can also update our Coil dependency to `io.coil-kt.coil3:coil-compose-core` as that removes the singleton entirely.

PR #1817, opened 2024-02-26
Refactor generic types in Composition and Widget hierarchy classes

ahmed3elshaer:


colinrtwhite:
@ahmed3elshaer Looks like this PR only includes renames for the generic params. Since it only changes the style and could be considered bikeshedding I don't think we can merge this, sorry.

ahmed3elshaer:
@colinrtwhite It is still a draft PR reflecting on https://github.com/cashapp/redwood/issues/1586
It will be closed and will create multiple PRs scoped for every generic type change.

PR #1823, opened 2024-02-27
Fix blocking touch events to non-subviews below Row/Column/Box on iOS.

colinrtwhite:
Fixes a case where `Column`/`Row`/`Box` on iOS would consume all touch events within its bounds even if it's empty space. [More context](https://cash.slack.com/archives/C01MNSK38DB/p1708718623388729). For example:

```kotlin
Box(
    width = Constraint.Fill,
    height = Constraint.Fill,
    horizontalAlignment = CrossAxisAlignment.Center,
    verticalAlignment = CrossAxisAlignment.Center,
  ) {
    Button("-1", onClick = { count-- })
    Column(
        horizontalAlignment = CrossAxisAlignment.Center,
        modifier = Modifier.size(300.dp, 300.dp),
    ) {
        Spacer(height = 100.dp)
        Text("Count: $count")
        Button("+1", onClick = { count++ })
    }
}
```

In the above case the `-1` button wouldn't be clickable without this fix.

Android Views/Compose UI already have this touch handling behaviour by default.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Changed:
 - The render function of `ComposeWidgetChildren` has been renamed to `Render` to reflect Compose best practices (https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#naming-unit-composable-functions-as-entities).
 
 Fixed:
-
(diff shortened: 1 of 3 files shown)


dnagler (approved):
Thanks for the quick turnaround!

JakeWharton (approved):

sallyjdevitry:
Nice, thank you!

Issue #900, opened 2023-04-10
Box Layout


swankjesse:
I’d like a built-in layout with a widget in the background (a grey roundrect) and another widget in the foreground (a column with some controls).

Perhaps the API to do this is:

```kotlin
Box {
  RoundRect(
    layoutModifier = LayoutModifier
       .horizontalAlignment(STRETCH)
       .verticalAlignment(STRETCH)
       .margin(Margin(horizontal = 8, vertical = 8)),
    color = ColorCode("#cccccc"),
    radius = 10
  )
  Column { ... }
}
```

Note that in this example, the background element’s width and height are both stretched to match the foreground element. It should also be possible to reverse these.

Also note that `RoundRect` is assumed to be defined by the application’s schema.

 * [x] Android views
   * https://github.com/cashapp/redwood/pull/1585
 * [x] Compose UI
   * https://github.com/cashapp/redwood/pull/1839
 * [x] iOS UIKit
   * https://github.com/cashapp/redwood/pull/1543
 * [ ] DOM
 * Screenshot tests
   * [x] Alignment tests
      * https://github.com/cashapp/redwood/pull/1830
   * [x] RTL tests
      * https://github.com/cashapp/redwood/pull/1837

squarejesse:
Needs HTML support!

PR #1830, opened 2024-02-29
Add screenshot tests for Box.

colinrtwhite:
Adds screenshot tests for `Box` and also includes a couple fixes for issues I noticed while writing the tests (can pull these into separate PRs if preferred):

- Fixes `ViewBox` to wrap by default to match `UIViewBox` and other layout widgets.
- Adds missing support for the `Size` modifier to both `View`s and `UIView`s.
- Adds missing `Margin` support to `UIViewBox`.
- Adds some missing calls to `setNeedsLayout` in `UIViewBox` when a widget parameter changes.

There's still one outstanding issue where `UIViewBox` doesn't fill in the tests. Will try to figure it out tomorrow.

Refs #900 

---


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import app.cash.redwood.layout.widget.Box
+import app.cash.redwood.ui.Dp
+import app.cash.redwood.ui.dp
+import kotlin.test.Test
+
+abstract class AbstractBoxTest<T : Any> {
+
+  abstract fun Box(block: Box<T>.() -> Unit = {}): Box<T>
+
+  abstract fun Color(block: Color<T>.() -> Unit = {}): Color<T>
```
JakeWharton: Is there any advantage to doing this over calling `.apply` at the call sites below? I'm guessing each implementation is doing something like `.apply(block)`.

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
```diff
@@ -0,0 +1,85 @@
...
+  fun Color(color: Int, width: Dp, height: Dp) = Color {
+    color(color)
+    width(width)
+    height(height)
+  }
+
+  abstract fun verifySnapshot(value: T)
+
+  @Test
+  open fun testDefaults() {
```
JakeWharton: Huh do the other test classes use `open`? I don't recall this being needed, but maybe it is.

redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewBoxTestHost/testFillWithChildren.1.png
JakeWharton, line 1: Can we give the box a background color? In this case it should fill the whole view but in the wrap case it shouldn't be otherwise visible, right? Right now they look the same.

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
```diff
@@ -0,0 +1,85 @@
...
+  open fun testFill() {
+    val widget = Box {
+      width(Constraint.Fill)
+      height(Constraint.Fill)
+    }
+    verifySnapshot(widget.value)
+  }
+
+  @Test
+  open fun testWrapWithChildren() {
```
JakeWharton: For this and the test below, can you make each test body a helper function and then make four test cases for the four alignments (top-left, top-right, bottom-left, bottom-right)? I can also do as a follow-up.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewBox.kt
```diff
@@ -38,83 +40,88 @@ import app.cash.redwood.widget.ViewGroupChildren
...
-    updateLayoutParams {
-      val layoutParams = this as MarginLayoutParams
+    updateLayoutParams<MarginLayoutParams> {
       with(density) {
-        layoutParams.setMargins(
-          margin.start.toPxInt(),
-          margin.top.toPxInt(),
-          margin.end.toPxInt(),
-          margin.bottom.toPxInt(),
+        updateMargins(
```
JakeWharton: Should be `updateMarginsRelative` with `start` and `end`, I think.

Going to file an issue to add RTL layout to our tests (not that we do a good job of even exercising margins in tests today).

redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewBoxTestHost/testFillWithChildren.1.png
JakeWharton: Just got to the view one which has a bg color!

JakeWharton (approved):

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewBox.kt
JakeWharton: Looks like we have them for flex containers already. So I'll just note it on the Box issue.

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
colinrtwhite: That's left over debug code so I could override + run in the IDE. Will update!

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
colinrtwhite: Yep not much benefit - I'll update and move `apply` to the callsites.

redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewBoxTestHost/testFillWithChildren.1.png
colinrtwhite: Yep it already has a background colour [here](https://github.com/cashapp/redwood/pull/1830/files#diff-9672fac5913cb114da7a1074d4611c8fcc160715beee71e0955d878317de47aaR31), but it's not filling properly at the moment in the tests. Taking a look

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
colinrtwhite: There's also `CENTER` so we'd end up with 2 x 9 tests. I can add them manually, but is there a better way to parameterize these tests?

JakeWharton: Nope!

We gotta build a parameterized Kotlin compiler plugin so they work on non-JVM.

colinrtwhite:
@JakeWharton Seems like there are a couple rendering issues in the `UIView` tests. Reverting my changes to `UIViewBox` produces the same results. Going to merge this and do a follow PR if that's ok.

JakeWharton:
Yep sounds good. Tests for the win!

PR #1838, opened 2024-03-01
Do not crash applying Box margins before view attach

JakeWharton:
Refs #900

---

### DIFF
--- a/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/utils.common.kt
+++ b/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/utils.common.kt
@@ -13,11 +13,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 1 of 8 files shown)


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/utils.common.kt
```diff
@@ -26,7 +37,7 @@ const val Red: Int = 0xffff0000.toInt()
 const val Green: Int = 0xff00ff00.toInt()
 const val Blue: Int = 0xff0000ff.toInt()
 
-fun Color(
+fun argb(
```
JakeWharton, line 40: The widget factory function is named `Color` and I was getting confused. Figured I'd appease ktlint while I was at it.

colinrtwhite (approved):

JakeWharton:
Looks like margins entirely not functioning on iOS. Going to file that for follow-up.

PR #1839, opened 2024-03-01
Add Compose UI `Box` implementation

JakeWharton:
Refs #900 

Not working yet:
- Margins on children
- Stretch alignment

But I'm saving those for follow-ups because anything is better than nothing.

---

### DIFF
--- a/redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
+++ b/redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
@@ -1,0 +1,122 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.layout.composeui
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.wrapContentHeight
+import androidx.compose.foundation.layout.wrapContentWidth
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.BiasAlignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import app.cash.redwood.Modifier as RedwoodModifier
+import app.cash.redwood.layout.api.Constraint
+import app.cash.redwood.layout.api.CrossAxisAlignment
+import app.cash.redwood.layout.widget.Box
+import app.cash.redwood.ui.Margin
+import app.cash.redwood.widget.compose.ComposeWidgetChildren
+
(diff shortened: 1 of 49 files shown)


redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
```diff
@@ -0,0 +1,122 @@
...
+
+internal class ComposeUiBox(
+  private val backgroundColor: Int? = null,
+) : Box<@Composable () -> Unit> {
+  override val children = ComposeWidgetChildren()
+
+  private var width by mutableStateOf(Constraint.Wrap)
+  private var height by mutableStateOf(Constraint.Wrap)
+  private var margin by mutableStateOf(Margin.Zero)
+  private var alignment by mutableStateOf(BiasAlignment(-1f, -1f))
```
JakeWharton, line 48: The relationship between `Alignment` and `BiasAlignment` is super weird. I can't create a two-directional `Alignment` from `Alignment.Horizontal` + `Alignment.Vertical`. I can't get the raw float values because they're exposed as those types and not the implementation types `BiasAlignment.Horizontal` and `BiasAlignment.Vertical` which are actually public (but whose `bias` float is private anyway). Super weird shit.

Going to file a bug about it.

redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
```diff
@@ -0,0 +1,122 @@
...
+    this.margin = margin
+  }
+
+  override fun horizontalAlignment(horizontalAlignment: CrossAxisAlignment) {
+    alignment = BiasAlignment(
+      horizontalBias = when (horizontalAlignment) {
+        CrossAxisAlignment.Start -> -1f
+        CrossAxisAlignment.Center -> 0f
+        CrossAxisAlignment.End -> 1f
+        CrossAxisAlignment.Stretch -> TODO()
```
JakeWharton, line 103: No idea how to do this. I think we need `BoxWithConstraints` or just a straight-up `Layout`. Future problems.

redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
```diff
@@ -0,0 +1,122 @@
...
+import androidx.compose.ui.graphics.Color
+import app.cash.redwood.Modifier as RedwoodModifier
+import app.cash.redwood.layout.api.Constraint
+import app.cash.redwood.layout.api.CrossAxisAlignment
+import app.cash.redwood.layout.widget.Box
+import app.cash.redwood.ui.Margin
+import app.cash.redwood.widget.compose.ComposeWidgetChildren
+
+internal class ComposeUiBox(
+  private val backgroundColor: Int? = null,
```
colinrtwhite: We could use `0x00000000` (full transparency) as the default here to avoid boxing and skip setting the modifier if `backgroundColor == 0x00000000`.

redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
colinrtwhite: We can partly accomplish this using `BoxScope.matchParentSize` and wrapping each child composable in a `Layout`. However, I don't think we can't match on one dimension - only both.

I took a quick look at the implementation and it doesn't look like we can easily wrap the `Modifier` to only wrap one dimension unfortunately. I think we probably do need a custom `Layout` implementation where we copy `Box`'s code and modify the `MeasurePolicy`.

colinrtwhite (approved):
Thanks for adding this 🙏🏻

redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
JakeWharton: I thought the very same thing as soon as I got in my car. Will change.

redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
JakeWharton: https://issuetracker.google.com/issues/328086355
https://issuetracker.google.com/issues/328088992

PR #1843, opened 2024-03-04
Add `Box` stretch tests

JakeWharton:
Status:

* ✅ view
* ❌ compose
* ❌ uikit

Refs #900

---

### DIFF
--- a/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
+++ b/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
@@ -87,6 +87,13 @@ abstract class AbstractBoxTest<T : Any> {
   )
 
   @Test
+  fun testWrapWithChildren_Wrap_Stretch_Start() = testWithChildren(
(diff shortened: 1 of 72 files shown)


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
```diff
@@ -86,6 +86,13 @@ abstract class AbstractBoxTest<T : Any> {
     verticalAlignment = CrossAxisAlignment.Start,
   )
 
+  @Test
+  fun testWrapWithChildren_Wrap_Stretch_Start() = testWithChildren(
```
JakeWharton, line 90: I noticed that all these tests specify `Wrap` twice in the name. Below we use `Fill` but then `Wrap`.

Will rename in a follow-up.

JakeWharton: https://github.com/cashapp/redwood/pull/1844

colinrtwhite (approved):

PR #1853, opened 2024-03-08
Make LazyList background clear for UIKit

underscoretang:
- UITableView and it’s cells default to white which is different from how the lazy list behaves on other platforms. This allows the parent of the lazy list to define it’s background color and lazy list is only concerned about layout.

---


dnagler (approved):
Works for me!

JakeWharton (approved):
Can you add a line in the change log for this?

PR #1854, opened 2024-03-08
Partially implement Compose UI Box stretch support.

colinrtwhite:
This PR adds a custom `Box` layout that controls the children's size via a custom `MeasurePolicy`. This is necessary as we can't apply `Modifier`s to child composables to change their constraints. We'll need this to support our other Redwood `BoxScope` modifiers as well.

This PR avoids adding support for the case where `CrossAxisAlignment.Stretch` and `Constraint.Wrap` are applied along the same axis. This is a bit tough to implement in Compose UI since we can only call `Measureable.measure` at most once, which means measuring the child locks its size and it can't be expanded to fill the size of the largest item in the box.

I think we might be able to implement this with a combination of `Measurable.[min/max]Intrinsic[Width/Height]`, but I'll tackle that in a separate PR.

---

I didn't add a line item to the `CHANGELOG` since the Compose UI Box support hasn't been released in a public release yet.


redwood-layout-composeui/src/commonMain/kotlin/app/cash/redwood/layout/composeui/ComposeUiBox.kt
```diff
@@ -30,29 +29,113 @@ import androidx.compose.ui.Alignment
...
+import app.cash.redwood.ui.Density
 import app.cash.redwood.ui.Margin
 import app.cash.redwood.widget.compose.ComposeWidgetChildren
 
 internal class ComposeUiBox(
   private val backgroundColor: Int = 0,
 ) : Box<@Composable () -> Unit> {
-  override val children = ComposeWidgetChildren()
+  private var modifierTick by mutableStateOf(0)
+  override val children = ComposeWidgetChildren(onModifierUpdated = { modifierTick++ })
```
colinrtwhite, line 52: Maybe it's better to expose a `State` object from `ComposeWidgetChildren` instead of exposing a callback?

JakeWharton: Seems reasonable

JakeWharton (approved):

PR #1858, opened 2024-03-11
Change TreehouseUIKitView's size as necessary

swankjesse:
If the content grows or shrinks, this reflects that.

I really didn't want to delegate to UIStackView here because this layout isn't a column at all. But after fighting it for hours and hours I gave up because I couldn't figure out how to react to my child views' size changes. I suspect the fix may be to add a listener on the bounds of the child views; nothing in the docs recommends this however. I'd love to know how UIStackView implements it!

---

### DIFF
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
@@ -71,7 +71,8 @@ internal class UIViewFlexContainer(
   }
 
   override fun onEndChanges() {
-    value.setNeedsLayout()
+    value.invalidateIntrinsicContentSize() // Tell the enclosing view that our size changed.
+    value.setNeedsLayout() // Update layout of subviews.
   }
 }
 
--- a/redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
+++ b/redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
@@ -20,6 +20,10 @@ import app.cash.redwood.treehouse.TreehouseView.WidgetSystem
 import app.cash.redwood.widget.RedwoodUIView
 import kotlinx.cinterop.cValue
 import platform.CoreGraphics.CGRectZero
+import platform.UIKit.UILayoutConstraintAxisVertical
+import platform.UIKit.UIStackView
+import platform.UIKit.UIStackViewAlignmentFill
+import platform.UIKit.UIStackViewDistributionFillEqually
 import platform.UIKit.UITraitCollection
 import platform.UIKit.UIView
 
@@ -33,7 +37,7 @@ public typealias TreehouseUIKitView = TreehouseUIView
 @ObjCName("TreehouseUIView", exact = true)
 public class TreehouseUIView private constructor(
   override val widgetSystem: WidgetSystem<UIView>,
-  view: UIView,
+  view: RootUiView,
 ) : TreehouseView<UIView>, RedwoodUIView(view) {
   override var saveCallback: TreehouseView.SaveCallback? = null
   override var stateSnapshotId: StateSnapshot.Id = StateSnapshot.Id(null)
@@ -50,26 +54,41 @@ public class TreehouseUIView private constructor(
   public constructor(widgetSystem: WidgetSystem<UIView>) : this(widgetSystem, RootUiView())
 
   init {
-    (view as RootUiView).treehouseView = this
+    view.treehouseView = this
   }
 
   private fun superviewChanged() {
     readyForContentChangeListener?.onReadyForContentChanged(this)
   }
 
-  private class RootUiView : UIView(cValue { CGRectZero }) {
+  /**
+   * The root view is just a vertical stack.
+   *
+   * In practice we expect this to contain either zero child subviews (especially when
+   * newly-initialized) or one child subview, which will usually be a layout container.
+   *
+   * This could just as easily be a horizontal stack. A box would be even better, but there's no
+   * such built-in component and implementing it manually is difficult if we want to react to
+   * content resizes.
+   */
+  private class RootUiView : UIStackView(cValue { CGRectZero }) {
     lateinit var treehouseView: TreehouseUIView
 
+    init {
+      this.axis = UILayoutConstraintAxisVertical
+      this.alignment = UIStackViewAlignmentFill // Fill horizontal.
+      this.distribution = UIStackViewDistributionFillEqually // Fill vertical.
+    }
+
     override fun layoutSubviews() {
+      super.layoutSubviews()
+
       // Bounds likely changed. Report new size.
       treehouseView.updateUiConfiguration()
-
-      subviews.forEach {
-        (it as UIView).setFrame(bounds)
-      }
     }
 
     override fun didMoveToSuperview() {
+      super.didMoveToSuperview()
       treehouseView.superviewChanged()
     }
 
--- a/redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/LayoutTester.kt
+++ b/redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/LayoutTester.kt
@@ -1,0 +1,159 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.treehouse
+
+import app.cash.redwood.widget.UIViewChildren
+import platform.CoreGraphics.CGRectMake
+import platform.UIKit.UILayoutConstraintAxisVertical
+import platform.UIKit.UIStackView
+import platform.UIKit.UIStackViewAlignmentCenter
+import platform.UIKit.UIStackViewAlignmentFill
+import platform.UIKit.UIStackViewDistributionEqualSpacing
+import platform.UIKit.UIStackViewDistributionFillEqually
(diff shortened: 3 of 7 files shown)


redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
```diff
@@ -50,26 +54,41 @@ public class TreehouseUIView private constructor(
...
       // Bounds likely changed. Report new size.
       treehouseView.updateUiConfiguration()
-
-      subviews.forEach {
-        (it as UIView).setFrame(bounds)
-      }
     }
 
     override fun didMoveToSuperview() {
+      super.didMoveToSuperview()
```
swankjesse, line 91: I’m unsure if this is strictly necessary!

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -71,7 +71,8 @@ internal class UIViewFlexContainer(
   }
 
   override fun onEndChanges() {
-    value.setNeedsLayout()
+    value.invalidateIntrinsicContentSize() // Tell the enclosing view that our size changed.
+    value.setNeedsLayout() // Update layout of subviews.
```
swankjesse, line 75: Without this, the TreehouseUIView doesn’t change sizes when new items are added to a flex container.

dnagler: Usually, the way to [force a layout pass](https://developer.apple.com/documentation/uikit/uiview/1622482-layoutsubviews) is to call `setNeedsLayout()` followed by `layoutIfNeeded()`--does that not suffice here?

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
dnagler: it's the right thing to do in general for any `UIView` lifecycle hook

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
```diff
@@ -50,26 +54,41 @@ public class TreehouseUIView private constructor(
...
+   * The root view is just a vertical stack.
+   *
+   * In practice we expect this to contain either zero child subviews (especially when
+   * newly-initialized) or one child subview, which will usually be a layout container.
+   *
+   * This could just as easily be a horizontal stack. A box would be even better, but there's no
+   * such built-in component and implementing it manually is difficult if we want to react to
+   * content resizes.
+   */
+  private class RootUiView : UIStackView(cValue { CGRectZero }) {
```
dnagler, line 74: Using a `UIStackView` when we expect it to always have at most one subview doesn't feel ideal, but if it helps with resizing the child view, it works for me.

dnagler (approved):

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
seanho: UIKit Autolayout and SwiftUI depends on `intrinsicContentSize` for sizing changes, while UIKit frame layout relies on needsLayout, we probably need both to support all cases.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
seanho: Not sure if Treehouse is using autolayout or frame-based layout under the hood, we have definitely run into issue before with `UIStackView` and frame-based layout.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
seanho: Probably not needed in this case according to [documentation](https://developer.apple.com/documentation/uikit/uiview/1622433-didmovetosuperview?language=objc).

>The default implementation of this method does nothing. Subclasses can override it to perform additional actions whenever the superview changes.

redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/LayoutTester.kt
```diff
@@ -0,0 +1,159 @@
...
+import platform.UIKit.UIStackViewDistributionFillEqually
+import platform.UIKit.UIView
+
+/**
+ * This layout puts 3 children in a 120 x 120 vertical stackview.
+ *
+ * Each child has initial intrinsic dimensions of 60 x 20.
+ *
+ * ```
+ * .-------------------------.
```
seanho, line 34: Big fan of this ascii visualization! I've even [made a GitHub pages app](https://seanho.github.io/ascii-layout/) to generate these.

swankjesse: Ooooh fun!

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
JakeWharton: How can we test both scenarios?

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
dnagler: I believe we're mostly using frame-based layout within this repo--In the widget bindings defined `cash-ios`, we use paralayout, which is frame-based as well.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
```diff
@@ -50,26 +54,41 @@ public class TreehouseUIView private constructor(
...
-  private class RootUiView : UIView(cValue { CGRectZero }) {
+  /**
+   * The root view is just a vertical stack.
+   *
+   * In practice we expect this to contain either zero child subviews (especially when
+   * newly-initialized) or one child subview, which will usually be a layout container.
+   *
+   * This could just as easily be a horizontal stack. A box would be even better, but there's no
+   * such built-in component and implementing it manually is difficult if we want to react to
+   * content resizes.
```
JakeWharton, line 72: Isn't the behavior of the root view the same that we want out of a UIKit `Box` which we also are required to provide though? Does that mean our UIKit `Box` is broken when its set to wrap/wrap and one of its children changes size?

redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/TreehouseUIViewLayoutTest.kt
```diff
@@ -0,0 +1,70 @@
...
+      for (vertical in Constraint.entries) {
+        layoutIsPassThrough(Subject.Reference, horizontal, vertical)
+        layoutIsPassThrough(Subject.TreehouseView, horizontal, vertical)
+      }
+    }
+  }
+
+  private fun layoutIsPassThrough(subject: Subject, horizontal: Constraint, vertical: Constraint) {
+    val tester = LayoutTester(subject, horizontal, vertical)
+    assertThat(tester.subjectFrame())
```
JakeWharton:
```suggestion
    assertThat(tester.subjectFrame(), "$horizontal $vertical")
```
otherwise you won't know which variant failed from just the message and will have to actually step debug

redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/TreehouseUIViewLayoutTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        Rectangle(
+          x = horizontal.initialX,
+          y = vertical.initialY,
+          width = horizontal.initialWidth,
+          height = vertical.initialHeight,
+        ),
+      )
+
+    tester.shrinkSubject()
+    assertThat(tester.subjectFrame())
```
JakeWharton:
```suggestion
    assertThat(tester.subjectFrame(), "$horizontal $vertical")
```

redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/TreehouseUIViewLayoutTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        Rectangle(
+          x = horizontal.shrunkX,
+          y = vertical.shrunkY,
+          width = horizontal.shrunkWidth,
+          height = vertical.shrunkHeight,
+        ),
+      )
+
+    tester.growSubject()
+    assertThat(tester.subjectFrame())
```
JakeWharton:
```suggestion
    assertThat(tester.subjectFrame(), "$horizontal $vertical")
```

JakeWharton (approved):
I think we probably want to get some screenshot tests going of the root view's behavior across all platforms the same way we do with the built-in widget impls.

Also can you add an entry in the change log? I will no longer be reviewing commits or PRs when creating releases.

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
swankjesse: I can do a follow-up to test a flex layout inside an enclosing layout. It’ll look a lot like the test added to this PR, but it’ll use a flex layout as the child of the TreehouseUIView.

redwood-treehouse-host/src/iosMain/kotlin/app/cash/redwood/treehouse/TreehouseUIView.kt
swankjesse: Yep, it’s quite likely Box is broken in the same way that TreehouseUIView is broken.

Perhaps worth doing the hacky work to observe child size changes for that... sigh.

swankjesse:
https://github.com/cashapp/redwood/issues/1859

PR #1862, opened 2024-03-12
Supply the index and widget when modifier changes

JakeWharton:
This will enable replacing the underlying view in the tree if future interceptor modifiers change the instance.

Refs #1084

---

### DIFF
--- a/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
+++ b/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
@@ -185,9 +185,9 @@ public interface RedwoodApplier<W : Any> {
  */
 @Composable
 @RedwoodCodegenApi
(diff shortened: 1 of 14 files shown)


redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
```diff
@@ -185,9 +185,9 @@ public interface RedwoodApplier<W : Any> {
  */
 @Composable
 @RedwoodCodegenApi
-public inline fun <P : Widget.Provider<*>, W : Widget<*>> RedwoodComposeNode(
+public inline fun <P : Widget.Provider<V>, W : Widget<V>, V : Any> RedwoodComposeNode(
```
JakeWharton, line 188: These generics are brutal, but it's for generated code only so whatever. This change is entirely so that the singleton `SetModifiers` function (in next file) works correctly in the generated code. As a bonus, we lose a cast here!

colinrtwhite (approved):

JakeWharton:
Probably should have a change log entry since it's an API change

Issue #1840, opened 2024-03-04
Better row insert/remove/move animations in LazyColumn


swankjesse:
Right now we tell UITableView / RecyclerView that rows have been added or removed, but we don’t expose APIs to express which animations to use.

josereyes:
I believe we need to replace UIViewLazyList.kt's `insertRows(...)` and `deleteRows(...)` `UITableViewRowAnimationNone` with `UITableViewRowAnimationAutomatic`

This video shows the expected behavior https://youtu.be/-joSiASAnjA?t=184

PR #1866, opened 2024-03-12
Adding begin/endUpdates calls to insert/deleteRows hooks in UIViewLazyList

dnagler:
As per [Apple's docs](https://developer.apple.com/documentation/uikit/uitableview/1614908-beginupdates), this allows the insertion/deletion animations to occur simultaneously across multiple rows.

This is "step 0" of addressing issue #1840.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Fixed:
 - The `View` implementation of `Box` now applies start/end margins correctly in RTL, and does not crash if set before the native view was attached.
 - Fix the backgroundColor for `UIViewLazyList` to be transparent. This matches the behavior of the other `LazyList` platform implementations.
 - Fix `TreehouseUIView` to size itself according to the size of its subview.
+- In `UiViewLazyList`, added `beginUpdates`/`endUpdates` calls to `insertRows`/`deleteRows` to improve animations.
 
 
 ## [0.9.0] - 2024-02-28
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):
I suspect both inserts and deletes can happen in one batch. Do we need to add start and end change callbacks to this listener as a follow up?

apatronl (approved):

dnagler:
> I suspect both inserts and deletes can happen in one batch. Do we need to add start and end change callbacks to this listener as a follow up?

That's what I'm thinking--I've got a [separate branch](https://github.com/cashapp/redwood/compare/dylan/batch-lazylist-updates) going where I'm batching the calls within the `LazyListUpdateProcessor`, but (for now, at least) it's making animations worse.

dnagler:
FYI--I tried calling [performBatchUpdates](https://developer.apple.com/documentation/uikit/uitableview/2887515-performbatchupdates) instead of begin/endUpdates, as per the doc's suggestion, but it led to the following crash:

```
Thread 1: "Invalid batch updates detected: the number of sections and/or rows returned by the data source before and after performing the batch updates are inconsistent with the updates.\nData source before updates = { 1 section with row counts: [1] }\nData source after updates = { 1 section with row counts: [1] }\nUpdates = [\n\tInsert row (0 - 0)\n]\nTable view: <UITableView: 0x11f1a3600; frame = (0 0; 0 0); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x60000114c180>; backgroundColor = UIExtendedGrayColorSpace 0 0; layer = <CALayer: 0x6000005e0ee0>; contentOffset: {0, 0}; contentSize: {0, 44}; adjustedContentInset: {0, 0, 0, 0}; dataSource: <CashKt_kobjc2: 0x600000567220>>"
```

I could probably work around this by calling `tableView.reloadData()` within the same block, but that feels like a step in the wrong direction.

PR #1870, opened 2024-03-13
Split and rename generated "factories" and "provider"

JakeWharton:
In order to support global modifiers, we need make changes to these types in a way that does not work with their current setup. In anticipation of that, rename them to give distinct behavior.

🤞 this is the last change pulled out/ahead of #1084

---

### DIFF
--- a/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
+++ b/redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
@@ -39,6 +39,8 @@ import app.cash.redwood.ui.OnBackPressedDispatcher
 import app.cash.redwood.ui.UiConfiguration
 import app.cash.redwood.widget.RedwoodView
 import app.cash.redwood.widget.Widget
+import app.cash.redwood.widget.WidgetFactoryOwner
+import app.cash.redwood.widget.WidgetSystem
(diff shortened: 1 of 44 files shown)


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolGuestGeneration.kt
```diff
@@ -47,11 +47,11 @@ import com.squareup.kotlinpoet.UNIT
 import com.squareup.kotlinpoet.joinToCode
 
 /*
-class ExampleProtocolBridge(
+class ExampleProtocolBridge private constructor(
```
JakeWharton, line 50: Drive-by cleanup

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/RedwoodComposition.kt
```diff
@@ -58,7 +60,7 @@ public interface RedwoodComposition {
 public fun <W : Any> RedwoodComposition(
   scope: CoroutineScope,
   view: RedwoodView<W>,
-  provider: Widget.Provider<W>,
+  widgetSystem: WidgetSystem<W>,
```
swankjesse, line 63: YAASS

swankjesse (approved):

colinrtwhite (approved):

PR #1874, opened 2024-03-14
Implement DOM layout sizing modifiers

dellisd:
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 New:
 - Compose UI implementation for `Box`.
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):
There's a slight problem here.

The removal of one of these modifiers will not clear the DOM property. Probably safest to delete all of the properties we set at the start of the function.

Can be in a follow-up.

JakeWharton:
Filed https://github.com/cashapp/redwood/issues/1877 to test that case (not that we have screenshot tests working for HTML yet)

PR #1890, opened 2024-03-25
Add support for unscoped modifiers

### LINKED ISSUE #1541: Introducing a backgroundColor modifier
There's a very common need for applying a background color on a view that I think is a pretty universal need that might be worth promoting into redwood. 

This came out of working on `Box`. A familiar approach for debugging and building views (particularly around layout) is setting background colors for different views that you're trying to piece them together and position elements. This applies within redwood, but would also apply to folks building and composing a bunch of widgets together in a treehouse app as well. 

The representation of the color is up to debate but i'll just use hex for illustration

Having the ability to create a 
```kotlin
Row(
  width = Constraint.Fill,
  height = Constraint.Fill,
  horizontalAlignment = MainAxisAlignment.Center,
  verticalAlignment = CrossAxisAlignment.Center,
  modifier = Modifier.backgroundColor("#FF0000"),
) {
   Text("Hello:")
   Text("World")
}
```
would be a much more experience than:
```kotlin
Box(
  width = Constraint.Fill,
  height = Constraint.Fill,  
  horizontalAlignment = MainAxisAlignment.Center,
  verticalAlignment = CrossAxisAlignment.Center,
) {
  Rectangle( // Doesn't exist in redwood
    width = Constraint.Fill,
    height = Constraint.Fill,
    color = "#FF0000"
  )
  Row(
    width = Constraint.Fill,
    height = Constraint.Fill,
    horizontalAlignment = MainAxisAlignment.Center,
    verticalAlignment = CrossAxisAlignment.Center,
  ) {
    Text("Hello:")
    Text("World")
  }
}
```

There's a case to be made for `cornerRadius` as well, so I'd like to have a discussion of where these additional view attributes begin and end. Talked to @swankjesse about having some amount of common attributes that are available on web would be ideal.

JakeWharton:
Unlike scoped modifiers, which are expected to be interpreted by the parent (and provider of the scope), unscoped modifiers are processed globally by the factory. These can be used for side-effecting operations, such as setting an accessibility description or universally supporting setting a background color.

I'm going to say this closes #1541 because it enables that to exist, but I don't think we want to provide it. A comprehensive background modifier probably wants things like corner radius support and maybe more options.

Refs #1084.

Things coming later:
- Tests (I want to change the test codegen first)
- Returning a new native UI node (Diff already too big, plus this isn't not urgently needed)
- #1583

---

### DIFF
--- a/redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolBridge.kt
+++ b/redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolBridge.kt
@@ -34,14 +34,14 @@ public interface ProtocolBridge : EventSink {
    * The root of the widget tree onto which [widgetSystem]-produced widgets can be added. Changes to
    * this instance are recorded as changes to [Id.Root] and [ChildrenTag.Root].
    */
-  public val root: Widget.Children<Nothing>
+  public val root: Widget.Children<Unit>
 
   /**
    * The provider of factories of widgets which record property changes and whose children changes
    * are also recorded. You **must** attach returned widgets to [root] or the children of a widget
    * in the tree beneath [root] in order for it to be tracked.
    */
-  public val widgetSystem: WidgetSystem<Nothing>
+  public val widgetSystem: WidgetSystem<Unit>
 
   /**
    * Returns any changes to [root] or [widgetSystem]-produced widgets since the last time this
--- a/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolFactory.kt
+++ b/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolFactory.kt
@@ -19,6 +19,7 @@ import app.cash.redwood.Modifier
 import app.cash.redwood.RedwoodCodegenApi
 import app.cash.redwood.protocol.ModifierElement
 import app.cash.redwood.protocol.WidgetTag
+import app.cash.redwood.widget.WidgetSystem
 import kotlin.native.ObjCName
 
 /**
@@ -27,7 +28,9 @@ import kotlin.native.ObjCName
  * @see ProtocolBridge
  */
 @ObjCName("ProtocolFactory", exact = true)
-public interface ProtocolFactory<W : Any>
+public interface ProtocolFactory<W : Any> {
+  public val widgetSystem: WidgetSystem<W>
+}
 
 /**
  * [ProtocolFactory] but containing codegen APIs.
--- a/redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/widgetGeneration.kt
+++ b/redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/widgetGeneration.kt
@@ -40,12 +40,29 @@ class ExampleWidgetSystem<W : Any>(
   override val Example: ExampleWidgetFactory<W>,
   override val RedwoodLayout: RedwoodLayoutWidgetFactory<W>,
 ) : WidgetSystem<W>,
-    ExampleWidgetFactoryProvider<W>,
-    RedwoodLayoutWidgetFactoryProvider<W>
+    ExampleWidgetFactoryOwner<W>,
+    RedwoodLayoutWidgetFactoryOwner<W> {
+  override fun apply(value: W, element: Modifier.UnscopedElement) {
+    Example.apply(value, element)
+    RedwoodLayout.apply(value, element)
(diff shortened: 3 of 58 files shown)


redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolBridge.kt
```diff
@@ -34,14 +34,14 @@ public interface ProtocolBridge : EventSink {
    * The root of the widget tree onto which [widgetSystem]-produced widgets can be added. Changes to
    * this instance are recorded as changes to [Id.Root] and [ChildrenTag.Root].
    */
-  public val root: Widget.Children<Nothing>
+  public val root: Widget.Children<Unit>
```
JakeWharton, line 37: An annoying change!

Because we hand the native UI instance to the factory callback, we need to be able to dereference it from the widget (i.e., `widget.value`). We obviously have no real nodes when serializing the protocol on the guest-side, but because the Compose `Applier` is shared with non-Treehouse Redwood, we still hit the `widget.value` codepath. Thus, we need _something_ to be our instance, so might as well be `Unit`.

I would have broken this out into its own change but I've already pulled out like 10 prerequisite PRs and it's time for this to actually start landing real behavior.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolFactory.kt
```diff
@@ -52,4 +52,6 @@ public interface GeneratedProtocolFactory<W : Any> : ProtocolFactory<W> {
    * or result in the unit [`Modifier`][Modifier.Companion] being returned.
    */
   public fun createModifier(element: ModifierElement): Modifier
+
+  public fun visit(value: W, element: Modifier.UnscopedElement)
```
JakeWharton: Whoops needs docs

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolBridge.kt
JakeWharton: Actually looking at it now, I think we can pass the `Widget<T>` to the `WidgetSystem` and have it responsible for unwrapping the value. The protocol-based, guest-side implementation is empty, so that allows us to keep `Nothing`.

This is only possible because I dropped the ability to return a wrapping native UI node. So when that gets added back we almost certainly need to make this change again. Going to walk this back and make a smaller diff!

JakeWharton: Nope this won't work 😢 . The `WidgetSystem` subtype is the same for both Redwood and Treehouse guest code. It's only the factories that change. Oh well!

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/widgetGeneration.kt
```diff
@@ -40,12 +40,29 @@ class ExampleWidgetSystem<W : Any>(
...
   override val RedwoodLayout: RedwoodLayoutWidgetFactory<W>,
 ) : WidgetSystem<W>,
-    ExampleWidgetFactoryProvider<W>,
-    RedwoodLayoutWidgetFactoryProvider<W>
+    ExampleWidgetFactoryOwner<W>,
+    RedwoodLayoutWidgetFactoryOwner<W> {
+  override fun visit(value: W, element: Modifier.UnscopedElement) {
+    Example.visit(value, element)
+    RedwoodLayout.visit(value, element)
+  }
```
JakeWharton: I think a future design for this is to give each schema a sealed unscoped modifier type. This would allow doing:
```kotlin
when (element) {
  is ExampleUnscopedModifier -> Example.visit(value, element)
  is RedwoodLayoutUnscopedModifier -> RedwoodLayout.visit(value, element)
  else -> throw!
}
```
to get throwing behavior on unknown modifiers.

It also means the `when` in each factory owner becomes exhaustive and gains an implicit throw (plus the inability to implement its `ExampleUnscopedModifier` interface).

JakeWharton:
Need to add `@ObjCName` to generated modifier interfaces.

colinrtwhite (approved):
Nice! Looks good to me. Agreed we probably shouldn't provide ~any first party unscoped layout modifiers.~ a first party background colour modifier. It might make sense to add some unscoped modifiers (e.g. accessibility tags). Worth discussing.

JakeWharton:
True. Anything that doesn't put pixels on the screen is a good candidate for in here.

PR #1896, opened 2024-03-28
ReuseList, a helper for doing compose-side widget reuse

swankjesse:
---

### DIFF
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ReuseList.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ReuseList.kt
@@ -20,6 +20,14 @@ import androidx.compose.runtime.movableContentOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 
+/**
+ * Execute [render] for each item in [items], while minimizing the amount of work the applier needs
+ * to do when elements are added and removed.
+ *
+ * This uses Compose's [movableContentOf] to match inserted items to removed items, and will reuse
+ * element subtrees where possible. This will also track moves of elements within the list, so long
+ * as the key function matches.
+ */
 @Composable
 public fun <T> ReuseList(
   items: List<T>,
@@ -61,7 +69,7 @@ public fun <T> ReuseList(
       nextPool[index] = entry
     }
 
-    entry.Show(item)
+    entry.show(item)
   }
 
   // Get the next pool ready for use.
@@ -76,5 +84,5 @@ private class PoolEntry<T>(
   render: @Composable (T) -> Unit,
 ) {
   var free = false
-  val Show: @Composable (T) -> Unit = movableContentOf(render)
+  val show: @Composable (T) -> Unit = movableContentOf(render)
 }
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ReuseList.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ReuseList.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.lazylayout.compose
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.movableContentOf
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+
+@Composable
+public fun <T> ReuseList(
+  items: List<T>,
+  key: (T) -> Any,
+  render: @Composable (T) -> Unit,
+) {
+  val poolState = remember { mutableStateOf(listOf<PoolEntry<T>>()) }
+  val pool = poolState.value
+  var freeEntryCount = pool.size
+
+  val nextPool = arrayOfNulls<PoolEntry<T>>(items.size)
+
+  // Find matches for items in the pool.
+  for ((index, item) in items.withIndex()) {
+    val itemKey = key(item)
+    val entry = pool.firstOrNull { it.free && it.key == itemKey }
+    if (entry != null) {
+      nextPool[index] = entry
+      freeEntryCount--
+    }
+  }
+
+  // Render each item and prepare the next pool.
+  for ((index, item) in items.withIndex()) {
+    var entry = nextPool[index]
+
+    if (entry == null) {
+      if (freeEntryCount > 0) {
+        entry = pool.first { it.free }
+        freeEntryCount--
+        entry.free = false
+        entry.key = key(item)
+      } else {
+        entry = PoolEntry(
+          key = key(item),
+          render = render,
+        )
+      }
+      nextPool[index] = entry
+    }
+
+    entry.Show(item)
+  }
+
+  // Get the next pool ready for use.
+  for (reusableItem in nextPool) {
+    reusableItem!!.free = true
+  }
+  poolState.value = nextPool.map { it!! }
+}
+
+private class PoolEntry<T>(
+  var key: Any,
+  render: @Composable (T) -> Unit,
+) {
+  var free = false
+  val Show: @Composable (T) -> Unit = movableContentOf(render)
+}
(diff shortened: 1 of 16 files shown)


redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
```diff
@@ -97,11 +97,11 @@ public class ProtocolBridge<W : Any>(
 
         is ModifierChange -> {
           val node = node(id)
-          val value = node.widget.value
 
           val modifier = change.elements.fold<_, Modifier>(Modifier) { outer, element ->
             val inner = factory.createModifier(element)
             if (inner is Modifier.UnscopedElement) {
+              val value = node.widget.value
```
swankjesse, line 104: This fixes a crash in my test because WidgetValue’s `node.widget.value` isn’t available until all the properties are assigned.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ReuseList.kt
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Execute [render] for each item in [items], while minimizing the amount of work the applier needs
+ * to do when elements are added and removed.
+ *
+ * This uses Compose's [movableContentOf] to match inserted items to removed items, and will reuse
+ * element subtrees where possible. This will also track moves of elements within the list, so long
+ * as the key function matches.
+ */
+@Composable
+public fun <T> ReuseList(
```
swankjesse, line 32: This isn’t strictly a LazyLayout thing; could just as easily go in redwood-runtime. Or even in the individual guest applications. There’s nothing particularly Redwood-proprietary here.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolState.kt
```diff
@@ -36,6 +38,48 @@ public class ProtocolState {
     changes.add(change)
   }
 
+  /**
+   * Each time we add a node, confirm it's preceded by either a [Create] or a [Remove].
+   *
+   * If we're adding a node that's been removed, the node should have been detached instead. Rewrite
+   * the change log to match what it should have been.
+   */
+  public fun requireCreateOrDetach(id: Id, widget: ProtocolWidget) {
```
swankjesse, line 53: This fix is necessary to make it all work. It has the effect of replacing a remove with a detach

redwood-protocol-guest/src/jsMain/kotlin/app/cash/redwood/protocol/guest/PlatformList.kt
```diff
@@ -20,6 +20,12 @@ internal external class JsArray<E> {
   @JsName("length")
   val size: Int
 
+  @JsName("at")
+  operator fun get(index: Int): E
+
+  @JsName("set")
```
swankjesse: I aught to validate that this works before I merge this PR.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ReuseList.kt
```diff
@@ -0,0 +1,88 @@
...
+  val poolState = remember { mutableStateOf(listOf<PoolEntry<T>>()) }
+  val pool = poolState.value
+  var freeEntryCount = pool.size
+
+  val nextPool = arrayOfNulls<PoolEntry<T>>(items.size)
+
+  // Find matches for items in the pool.
+  for ((index, item) in items.withIndex()) {
+    val itemKey = key(item)
+    val entry = pool.firstOrNull { it.free && it.key == itemKey }
```
dnagler, line 79: Just to confirm: if there are multiple items with the same key, then this will return the same `entry` for all of them.
Is this okay/expected, or should we be setting `entry.free = false` as part of this loop?

redwood-protocol-guest/src/jsMain/kotlin/app/cash/redwood/protocol/guest/PlatformList.kt
JakeWharton: For `set` I used `splice`: https://github.com/JakeWharton/platform-collections/blob/trunk/src/jsMain/kotlin/com/jakewharton/platformcollections/PlatformList.kt#L56-L58

No clue why I did this. Does Kotlin even let you produce `subject[index] = value` in the JS? There's no normal function to use other than bracket notation, right? I didn't see a `set`.

For `get` I used `asDynamic()` plus regular indexing: https://github.com/JakeWharton/platform-collections/blob/trunk/src/jsMain/kotlin/com/jakewharton/platformcollections/PlatformList.kt#L28-L34

This saves a few bytes and looks more idiomatic in the JS. `at` is relatively new, but should be fine for browsers and QuickJS though.

Not sure if you want index checking. Brackets and `at` behave differently between -1 and -length. Both return undefined outside that and normal range. An explicit index check means you avoid the negative index behavior of `at`.

Probably want extensions for both operators funs and then at or brackets + splice on the actual type.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
JakeWharton: You'll get the same crash on anything using an unscoped modifier though. The fold is synchronous.

You mean the MutableXyz widgets which produce WidgetNodes as their value is the cause of the crash? We might have to change their view type to use Unit, then, and have a separate function to snapshot the values which walks the tree. We'll need to be able to also wrap this value in the future here, not just reference it.

redwood-protocol-guest/src/jsMain/kotlin/app/cash/redwood/protocol/guest/PlatformList.kt
swankjesse: Extensions worked.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ReuseList.kt
swankjesse: YES to setting entry.free in this loop. Good catch. I fixed that.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolRedwoodComposition.kt
```diff
@@ -48,21 +49,28 @@ public fun ProtocolRedwoodComposition(
...
     widgetSystem = bridge.widgetSystem,
+  )
+  return ProtocolRedwoodComposition(
+    composition,
+    widgetVersion
   ) {
-    bridge.getChangesOrNull()?.let(changesSink::sendChanges)
+    key(bridge.compositionsCount) {
+      bridge.getChangesOrNull()?.let(changesSink::sendChanges)
+    }
```
swankjesse, line 62: I needed to defer this side effect until after the 'late changes' callback happened, otherwise we'd get two batches of changes with the remove in the first, and the add in the second, and the add would fail because the element wouldn’t exist on the host anymore.

swankjesse: Unfortunately, it’s still broken with a similar root cause . . . the host app crashes because we get an add() that precedes the corresponding remove():

```
FATAL EXCEPTION: main
Process: com.example.redwood.emojisearch.android.views, PID: 11706
java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.
	at android.view.ViewGroup.addViewInner(ViewGroup.java:5247)
	at android.view.ViewGroup.addView(ViewGroup.java:5076)
	at android.view.ViewGroup.addView(ViewGroup.java:5016)
	at android.view.ViewGroup.addView(ViewGroup.java:4988)
	at app.cash.redwood.lazylayout.view.ViewLazyList$ViewHolder.setContent(ViewLazyList.kt:286)
	at app.cash.redwood.lazylayout.view.ViewLazyList$processor$1.setContent(ViewLazyList.kt:92)
	at app.cash.redwood.lazylayout.view.ViewLazyList$processor$1.setContent(ViewLazyList.kt:74)
	at app.cash.redwood.lazylayout.widget.LazyListUpdateProcessor$Binding.bind$redwood_lazylayout_widget(LazyListUpdateProcessor.kt:402)
	at app.cash.redwood.lazylayout.widget.LazyListUpdateProcessor.bind(LazyListUpdateProcessor.kt:302)
	at app.cash.redwood.lazylayout.view.ViewLazyList$LazyContentItemListAdapter.onBindViewHolder(ViewLazyList.kt:263)
	at app.cash.redwood.lazylayout.view.ViewLazyList$LazyContentItemListAdapter.onBindViewHolder(ViewLazyList.kt:233)
```

JakeWharton: I think this change moves the callback to fire earlier, not later. The old one fired on `Applier.onEndChanges` which is after recomposition has walked the whole function tree. You are now triggering it within the function tree.

swankjesse: Yeah, I need a `SideEffect {}` here, which should run after all the composition stuff.

PR #1897, opened 2024-03-28
Declare GeneratedProtocolFactory.childrenTags and generate it

swankjesse:


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolHostGeneration.kt
```diff
@@ -115,6 +129,37 @@ internal fun generateProtocolFactory(
...
+                    add("%T(%L) to %M(", WidgetTag, widget.tag, listOf)
+                    var first = true
+                    for (children in widget.traits
+                      .filterIsInstance<ProtocolChildren>()
+                      .sortedBy { it.tag }
+                    ) {
+                      if (!first) add(", ")
+                      first = false
+                      add("%T(%L)", ChildrenTag, children.tag)
+                    }
```
JakeWharton: If you want, this can be
```suggestion
                    add(
                      widget.traits
                        .filterIsInstance<ProtocolChildren>()
                        .sortedBy { it.tag }
                        .map { CodeBlock.of("%T(%L)", ChildrenTag, children.tag) }
                        .joinToCode()
                    )
```

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/WidgetProtocolGenerationTest.kt
```diff
@@ -67,4 +67,15 @@ class WidgetProtocolGenerationTest {
...
+  @Test fun `children tags are included in generated protocol factory`() {
+    val schema = ProtocolSchemaSet.parse(TestSchema::class)
+
+    val fileSpec = generateProtocolFactory(schema)
+    assertThat(fileSpec.toString()).all {
+      contains("WidgetTag(1) to listOf(ChildrenTag(1)),")
+      contains("WidgetTag(1_000_003) to listOf(),")
+      contains("WidgetTag(2_000_002) to listOf(ChildrenTag(1), ChildrenTag(2)),")
+    }
+  }
```
JakeWharton: I think I want to say just delete this? It'll get tested implicitly when something starts relying on it. I'm not sure why we have the test above this. That also seems like it could be black-box tested. The only one that can't be is the first one where we validate sorting.

JakeWharton (approved):

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolHostGeneration.kt
JakeWharton: Filed an issue to collapse the `map`+`joinToCode` into one line, too: https://github.com/square/kotlinpoet/issues/1873

redwood-tooling-codegen/src/test/kotlin/app/cash/redwood/tooling/codegen/WidgetProtocolGenerationTest.kt
swankjesse: Deleted.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolHostGeneration.kt
swankjesse: Done.

PR #1898, opened 2024-03-28
Expose tracked widgets on children

JakeWharton:
In practice, every non-specialized subtype was already maintaining this list. This will be needed for subtree traversal at the protocol layer in a future change.

---


colinrtwhite (approved):
Nice!

Issue #1901, opened 2024-03-28
Consider exposing Redwood version bi-directionally across Treehouse bridge


JakeWharton:
This will enable us to work around bugs in the host from the guest based on its version. Or vise versa.

- [x] Pipe host version to guest
  - #1929 
- [x] Pipe guest version to host
  - #1939
- [x] Work around memory leak in old hosts from guest
  - #1936

JakeWharton:
We need this to stop sending `removedIds` from the guest to the host.

JakeWharton:
Probably better to just do a protocol version instead

PR #1904, opened 2024-03-29
Implement host-side view recycling

swankjesse:
---


swankjesse (commented):
I’m leaving this out of the CHANGELOG because it’s not yet an intentional public API.

redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ChangeListenerTest.kt
```diff
@@ -91,13 +93,19 @@ abstract class AbstractChangeListenerTest {
     c.setContent {
       Button(text, onClick = null)
     }
-    assertThat(c.awaitSnapshot()).containsExactly("modifier Modifier", "text hi", "onClick false", "onEndChanges")
+    assertThat(c.awaitSnapshot()).containsExactlyInAnyOrder(
```
JakeWharton: This does not seem like a desirable effect. This test will now pass if `onEndChanges` is called before some property or the modifier is set. Why is the behavior changing between regular Redwood and protocol Redwood? That also seems like an indicator something is wrong.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
```diff
@@ -137,6 +169,225 @@ public class ProtocolBridge<W : Any>(
...
+      }
+    }
+
+    // Build a new changes list that omits the events we no longer need.
+    val result = mutableListOf<Change>()
+    for (change in changesAndNulls) {
+      if (change != null) result += change
+    }
+
+    return result
```
JakeWharton:
```suggestion
    return changesAndNulls.filterNotNull()
```

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
```diff
@@ -56,13 +57,35 @@ public class ProtocolBridge<W : Any>(
...
+        //  - no redundant modifier updates on newly-created children
+        //  - we don't apply modifiers until their properties are all set
+        //  - we can't change the relative orders of insert/move/remove without adjusting indexes
+        when (it) {
+          is Create -> 0
+          is PropertyChange -> 1
+          is ModifierChange -> 2
+          else -> 3
+        }
+      }
```
JakeWharton: This order is already guaranteed by the protocol with the exception of modifiers coming before properties. If you want to invert that order, we can change the generated Compose code to put the modifier update last in the list.

This will retain symmetric behavior between Redwood and protocol Redwood, and avoid the need to use the "any order" version of the assertion in the tests.

It doesn't solve the case of a newer host talking to an older guest, but only in that case can we add in the additional sorting (although it probably needs #1901) then.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
```diff
@@ -137,6 +169,225 @@ public class ProtocolBridge<W : Any>(
...
+   */
+  private fun applyReuse(changes: List<Change>): List<Change> {
+    val idToSubtree = mutableMapOf<Id, ReuseSubtree<W>>()
+
+    // Find nodes that have Modifier.reuse
+    for ((index, change) in changes.withIndex()) {
+      if (change !is ModifierChange) continue
+
+      val reuseModifierIndex = change.elements.indexOfFirst { it.tag.value == REUSE_MODIFIER_TAG }
+      if (reuseModifierIndex == -1) continue
```
JakeWharton: This also short-circuits
```suggestion
      val wantsReuse = change.elements.any { it.tag.value == REUSE_MODIFIER_TAG }
      if (!wantsReuse) continue
```

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
```diff
@@ -137,6 +169,225 @@ public class ProtocolBridge<W : Any>(
...
+
+      val candidateIndex = pool.indexOfFirst { it.reuse }
+      if (candidateIndex == -1) continue
+
+      if (changes.subList(0, index).none { it is Create && it.id == change.id }) continue
```
JakeWharton, line 183: I think this is accidentally quadratic? At least in the case of enough reuse. Since the protocol guarantees a create immediately precedes the property+modifier assignment, we can walk backwards from `index` until we see either a `Create` (and `continue`) or `change.id != it.id` (and stop looking). This should minimize the look-behind to a few changes at most.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
JakeWharton: That is to say, I think the plan is:
- Land this as-is but with the Compose generation fix and changing the tests back
- Add the conditional sorting in a follow-up

test-app/ios-uikit/TestApp/IosTestSchemaWidgetFactory.swift
```diff
@@ -51,6 +51,14 @@ class IosTestSchemaWidgetFactory: TestSchemaWidgetFactory {
     func BackgroundColor(value: Any, modifier: BackgroundColor) {
         (value as! UIView).backgroundColor = UIColor(argb: UInt(modifier.color))
     }
+
+    func Split() -> Split {
+        fatalError()
+    }
+
+    func Reuse(value: Any, modifier: Reuse) {
+        fatalError()
```
JakeWharton:
```suggestion
```

JakeWharton (approved):
I didn't review the tests. Brain hurts from looking at the algorithm.

What happened to `@ReuseProperty`?

redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ChangeListenerTest.kt
JakeWharton: I figured this out later, and proposed some changes to avoid needing to relax this

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
swankjesse: Since this only causes problems on the new unscoped modifiers, I can fix the Compose code now!

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
swankjesse: Much better. An earlier version wanted `reuseModifierIndex` but I don’t need that any more.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
swankjesse: Great call. I think there’s an entire follow-up PR where I take this kinda-brute-force approach and tune it for some big efficiency gains. If you don’t mind I have a few such changes in mind and I’d like to do them in a follow-up.

(This PR: it works. Next PR: it still works, and it’s a bit faster.)

swankjesse:
> What happened to @ReuseProperty?

I decided that I wanted to do reuse on `Row` and `Column` etc., and the unscoped modifiers were whispering in my ear.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/widget/ProtocolBridge.kt
swankjesse: Yay! Good idea. https://github.com/cashapp/redwood/commit/f2a3af9d4615917e4a8299ca4f8d70f52194c951

JakeWharton: Excellent. And yeah even better that we don't need to do it conditionally since old guests can't send it.

PR #1911, opened 2024-03-29
Change protocol-host package to match its name

### LINKED ISSUE #1903: Rename redwood-protocol-host package to `.host` (from `.widget`)
I renamed the protocol guest, but never followed up on the host. We are currently working on the protocol so let's do this once that work settles (but before the next release).

JakeWharton:
Closes #1903

---


swankjesse (approved):
If we have any ZiplineServices that use these types as parameters, we’ll need to manually tag em with @ZiplineId. (And we need to implement that feature. https://github.com/cashapp/zipline/issues/1048)

JakeWharton:
We don't. They're host-side only. The only thing we might refer to ourselves is the mismatch handler. Everything else should be implementation detail of treehouse-host.

PR #1913, opened 2024-03-30
Change widget reuse to use hashes to find a match in the pool

swankjesse:
The previous strategy assumed all nodes were symmetric.

### DIFF
--- a/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
+++ b/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
@@ -141,8 +141,11 @@ public class ProtocolBridge<W : Any>(
     // Apply deferred removes. Do this last so properties + modifiers always find their widget,
     // even if it's marked for death.
     for (removedId in removedIds) {
-      val removedNode = nodes.remove(removedId)
-      if (removedNode?.reuse == true) pool.addLast(removedNode)
+      val removedNode = nodes.remove(removedId) ?: continue
+      if (removedNode.reuse) {
+        removedNode.shapeHash = shapeHash(factory, removedNode)
+        pool.addLast(removedNode)
+      }
     }
 
     if (changedWidgets.isNotEmpty()) {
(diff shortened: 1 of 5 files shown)


redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
```diff
@@ -141,8 +141,11 @@ public class ProtocolBridge<W : Any>(
     // Apply deferred removes. Do this last so properties + modifiers always find their widget,
     // even if it's marked for death.
     for (removedId in removedIds) {
```
JakeWharton, line 143: I think this was an artifact of when you were sorting changes. If properties or modifiers ever came in after a remove an exception would be thrown because the map lookup by ID would fail.

I have removed this in #1920

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
```diff
@@ -168,77 +171,88 @@ public class ProtocolBridge<W : Any>(
...
-      val wantsReuse = change.elements.any { it.tag.value == REUSE_MODIFIER_TAG }
-      if (!wantsReuse) continue
-
-      val candidateIndex = pool.indexOfFirst { it.reuse }
-      if (candidateIndex == -1) continue
+      // Must have a reuse modifier.
+      if (change.elements.none { it.tag.value == REUSE_MODIFIER_TAG }) continue
 
+      // Must have a Create node that precedes it.
       if (changes.subList(0, index).none { it is Create && it.id == change.id }) continue
```
JakeWharton, line 185: Filed #1924 so we don't forget

JakeWharton (approved):

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
swankjesse: Awesome.

PR #1918, opened 2024-03-31
Remove widgets recursively on guest side

### LINKED ISSUE #1899: Protocol guest leaks child nodes when entire subtree removed
When a subtree is removed, we only remove the root item from the map. We need to walk the tree and remove all children.

### LINKED ISSUE #1900: Protocol host leaks child nodes when entire subtree removed
When a subtree is removed, we only remove the root item from the map. We need to walk the tree and remove all children.

JakeWharton:
Closes #1899.

~With #1901 we can send the entire subtree in the `removedIds` list to the host to fix #1900 for older hosts.~ We cannot. See #1920 for reason.

---

### DIFF
--- a/redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
+++ b/redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
@@ -20,6 +20,8 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
 import app.cash.redwood.compose.WidgetVersion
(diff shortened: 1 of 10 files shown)


redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
```diff
@@ -277,6 +282,38 @@ class ProtocolTest {
     )
   }
 
+  @Test fun entireSubtreeRemoved() = runTest {
```
swankjesse, line 285: Nice test

swankjesse (approved):
Nice fix. 

I need to add a recycling test to confirm I’m not relying on this load-bearing bug!

PR #1920, opened 2024-04-01
Remove widgets recursively on host side

### LINKED ISSUE #1900: Protocol host leaks child nodes when entire subtree removed

JakeWharton:
Closes #1900. We no longer rely on `removedIds` from the protocol. ~With #1901 we can conditionally use the `removedIds` to send an entire subtree's IDs from the guest-side to fix this bug on the host-side for older hosts. That's why this doesn't _yet_ close the issue.~ We cannot work around this for older hosts. The protocol validates that the `count == removedIds.size` on both sides, so we can't sneak more values in.

---

### DIFF
--- a/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
+++ b/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
@@ -64,15 +64,12 @@ public class ProtocolBridge<W : Any>(
     @Suppress("NAME_SHADOWING")
     val changes = applyReuse(changes)
 
-    val removedIds = mutableListOf<Id>()
-
     for (i in changes.indices) {
       val change = changes[i]
       val id = change.id
       when (change) {
         is Create -> {
-          val node = factory.createNode(change.tag) ?: continue
-          node.widgetTag = change.tag
+          val node = factory.createNode(id, change.tag) ?: continue
           val old = nodes.put(change.id, node)
           require(old == null) {
             "Insert attempted to replace existing widget with ID ${change.id.value}"
@@ -93,8 +90,12 @@ public class ProtocolBridge<W : Any>(
             }
 
             is Remove -> {
+              for (childIndex in change.index until change.index + change.count) {
+                val child = children.nodes[childIndex]
+                poolIfReusable(child)
+                child.visitIds(nodes::remove)
+              }
               children.remove(change.index, change.count)
-              removedIds += change.removedIds
             }
           }
 
@@ -138,19 +139,6 @@ public class ProtocolBridge<W : Any>(
       }
     }
 
-    // Apply deferred removes. Do this last so properties + modifiers always find their widget,
-    // even if it's marked for death.
-    for (removedId in removedIds) {
-      val removedNode = nodes.remove(removedId) ?: continue
-      if (removedNode.reuse) {
-        removedNode.shapeHash = shapeHash(factory, removedNode)
-        pool.addFirst(removedNode)
-        if (pool.size > POOL_SIZE) {
-          pool.removeLast() // Prune the least-recently added element.
-        }
-      }
-    }
-
     if (changedWidgets.isNotEmpty()) {
       for (widget in changedWidgets) {
         widget.onEndChanges()
@@ -161,6 +149,16 @@ public class ProtocolBridge<W : Any>(
 
   private fun node(id: Id): ProtocolNode<W> {
     return checkNotNull(nodes[id]) { "Unknown widget ID ${id.value}" }
+  }
+
+  private fun poolIfReusable(removedNode: ProtocolNode<W>) {
+    if (removedNode.reuse) {
+      removedNode.shapeHash = shapeHash(this.factory, removedNode)
+      pool.addFirst(removedNode)
+      if (pool.size > POOL_SIZE) {
+        pool.removeLast() // Prune the least-recently added element.
+      }
+    }
   }
 
   /**
@@ -342,7 +340,7 @@ public class ProtocolBridge<W : Any>(
 @OptIn(RedwoodCodegenApi::class)
 private class RootProtocolNode<W : Any>(
   children: Widget.Children<W>,
-) : ProtocolNode<W>(), Widget<W> {
+) : ProtocolNode<W>(Id.Root, UnknownWidgetTag), Widget<W> {
   private val children = ProtocolChildren(children)
 
   override fun apply(change: PropertyChange, eventSink: EventSink) {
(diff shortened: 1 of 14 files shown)


redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
```diff
@@ -138,13 +141,6 @@ public class ProtocolBridge<W : Any>(
       }
     }
 
-    // Apply deferred removes. Do this last so properties + modifiers always find their widget,
-    // even if it's marked for death.
-    for (removedId in removedIds) {
-      val removedNode = nodes.remove(removedId)
-      if (removedNode?.reuse == true) pool.addLast(removedNode)
-    }
```
JakeWharton: This needing to occur a separate step seems like a relic of the change list sorting. The protocol should already guarantee this, as removal from the map would fail any future lookups by ID.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
```diff
@@ -57,22 +57,26 @@ public class ProtocolBridge<W : Any>(
   )
   private val changedWidgets = mutableSetOf<ChangeListener>()
 
+  private val removeNodeById: (Id) -> Unit = { nodes.remove(it) }
+  private val poolIfReusable: (ProtocolNode<W>) -> Unit = { it ->
+    if (it.reuse) {
+      pool.addLast(it)
+    }
+  }
```
swankjesse: Is this JavaScript? What a strange way to declare a function

```suggestion
  private fun poolIfReusable(node: ProtocolNode<W>) {
    if (node.reuse) {
      pool.addLast(node)
    }
  }
```

swankjesse: (I assume you’ve got a reason for this, but as a reader of this code I want an explanation comment here where the weirdness is)

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolNode.kt
```diff
@@ -31,15 +32,17 @@ import kotlin.math.min
  * @suppress
  */
 @RedwoodCodegenApi
-public abstract class ProtocolNode<W : Any> {
+public abstract class ProtocolNode<W : Any>(
+  public val id: Id,
+  public val widgetTag: WidgetTag,
```
swankjesse, line 37: Nice

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolNode.kt
```diff
@@ -85,8 +91,19 @@ public class ProtocolChildren<W : Any>(
     }
   }
 
-  internal fun remove(index: Int, count: Int) {
+  /**
+   * @param subtreeIds Invoked with the ID of each removed widget and all widgets contained within
+   * its children, recursively.
+   * @param removedNode Invoked with each removed node on this instance (i.e., not recursively).
+   */
+  internal fun remove(index: Int, count: Int, subtreeIds: (Id) -> Unit, removedNode: (ProtocolNode<W>) -> Unit) {
```
swankjesse: This function signature is awkward . . . it’s designed to support exactly one callsite, but defined here and not at that callsite. I think it’d be simpler to move the behavior there?

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolNode.kt
```diff
@@ -113,6 +130,12 @@ public class ProtocolChildren<W : Any>(
 
     children.move(from, to, count)
   }
-}
 
-internal val UnknownWidgetTag: WidgetTag = WidgetTag(-1)
```
swankjesse, line 121: Nice

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolNode.kt
swankjesse: (I’m also a little bit anxious about the amount of functions and IDs being boxed to support this faux encapsulation?)

redwood-protocol-host/src/commonTest/kotlin/app/cash/redwood/protocol/host/ProtocolBridgeTest.kt
```diff
@@ -177,4 +185,63 @@ class ProtocolBridgeTest {
...
+    )
+
+    // Remove root TestRow.
+    bridge.sendChanges(
+      listOf(
+        Remove(Id.Root, ChildrenTag.Root, 0, 1, listOf(Id(1))),
+      ),
+    )
+
+    assertFailure {
```
swankjesse, line 237: LOVE IT

swankjesse (approved):

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
JakeWharton: If you declare it as a function then an additional class needs to be generated to map it back to a lambda type, and that class needs instantiated for each remove change. The current setup is allocation-free.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolNode.kt
JakeWharton: Yeah it's very much not great, but the entire protocol-based system is really a single unit that's all aware of each other. This "design" (or non-design, really) is in service of avoiding the need to allocate anything.

The other one I built removed the native views (`children.remove(..)`) and then returned the `nodes.sublist` to the caller for it to consume, but then to also clear, which was also weird ownership-wise. Anything else I built had a bunch of allocation.

In the guest, the relationship is inverted where the nodes hold a shared state. In the host, the state holds the nodes, so there's no way for the nodes to mutate the state directly like the guest does.

JakeWharton: I didn't realize `nodes` was internal and already used by your reuse stuff. That will allow moving it to the caller and avoiding lambdas altogether.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
JakeWharton: I rolled all this back and will address separately. There's also a problem still on the guest-side. The lambda type is `KFunction1` which still causes `Id` to box. We probably need a `fun interface IdVisitor { fun visitId(id: Id) }` to avoid that on both guest and hots. Future problem.

PR #1921, opened 2024-04-01
Change Gradle plugin IDs to guest/host

### LINKED ISSUE #1907: Rename Gradle plugin IDs for protocol generation
Should probably use host and guest. Need to keep old IDs around with a deprecation warning if they are used for a release cycle.

JakeWharton:
Keep the old ones around as deprecated for one release.

Closes #1907

---


swankjesse (approved):
These new names are awesome

PR #1929, opened 2024-04-03
Create `RedwoodVersion` type for sharing across host-guest bridge

JakeWharton:
In order to work around bugs in the future or change the protocol in a backwards-incompatible way, the version gives the other side of the bridge the ability to change itself to remain compatible.

For now, we only have the host version hooked up to the guest side. The guest version being exposed to the host will come in a follow-up.

Part of #1901.

---

### DIFF
--- a/redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/RedwoodVersion.kt
+++ b/redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/RedwoodVersion.kt
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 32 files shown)


JakeWharton:
~There's two places where I need to try/catch zipline exception and default to `RedwoodVersion("0.0.0")`...~

And ideally I would somehow test that, but I think that's probably a bit hard.

colinrtwhite (approved):

redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/RedwoodVersion.kt
```diff
@@ -0,0 +1,96 @@
...
+   * Compare two versions to see which is newer.
+   *
+   * Note: Comparing instances is not particularly efficient, so the result should be cached
+   * instead of comparing each time.
+   */
+  override fun compareTo(other: RedwoodVersion): Int {
+    val thisMatch = format.matchEntire(value)!!
+    val otherMatch = format.matchEntire(other.value)!!
+
+    // First three parts are digits (1-indexed because part 0 is whole match).
```
swankjesse, line 64: Nice comment

swankjesse (approved):

PR #1931, opened 2024-04-03
Optimize detecting whether a reused widget was created

### LINKED ISSUE #1924: Reuse detection is accidentally quadratic
```kotlin
if (changes.subList(0, index).none { it is Create && it.id == change.id }) continue
```
We can walk backwards while `it.id == change.id` to look for a `Create`. No further. This should reduce the search space to a few indices at worst.

JakeWharton:
Because the protocol guarantees that create/property/modifier changes are all grouped together in that order, we can simply track the last created widget we saw to determine if the target widget was created in this set of changes.

Closes #1924

---


colinrtwhite (approved):
Nice!

PR #1932, opened 2024-04-03
Fix coroutine scoping of Redwood Treehouse UI

dellisd:
Ensure that Zipline resources aren't closed until the UI's coroutine scope job completes.

This prevents a queued-up job from resuming(?) and crashing from a `ZiplineException` instead of gracefully handling a `CancellationException` when trying to collect a `ZiplineService`-backed flow.

---

### DIFF
--- a/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
+++ b/redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
@@ -27,8 +27,12 @@ import app.cash.redwood.ui.OnBackPressedDispatcher
 import app.cash.redwood.ui.UiConfiguration
 import app.cash.zipline.ZiplineScope
 import app.cash.zipline.ZiplineScoped
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.job
 import kotlinx.coroutines.plus
 
 /**
@@ -54,6 +58,11 @@ private class RedwoodZiplineTreehouseUi(
    * host configurations flow.
    */
   override val scope = (treehouseUi as? ZiplineScoped)?.scope ?: ZiplineScope()
+
+  private val coroutineScope = CoroutineScope(
+    appLifecycle.coroutineScope.coroutineContext +
+      Job(appLifecycle.coroutineScope.coroutineContext.job),
+  )
 
   private lateinit var composition: RedwoodComposition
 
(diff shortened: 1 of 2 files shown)


redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
```diff
@@ -121,9 +130,12 @@ private class RedwoodZiplineTreehouseUi(
   }
 
   override fun close() {
+    coroutineScope.coroutineContext.job.invokeOnCompletion {
```
swankjesse, line 133: Love it

swankjesse (approved):

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
```diff
@@ -55,6 +59,11 @@ private class RedwoodZiplineTreehouseUi(
    */
   override val scope = (treehouseUi as? ZiplineScoped)?.scope ?: ZiplineScope()
 
+  private val coroutineScope = CoroutineScope(
+    appLifecycle.coroutineScope.coroutineContext +
+      SupervisorJob(appLifecycle.coroutineScope.coroutineContext.job),
```
JakeWharton: Why this change? Doesn't this mean crashes in the composition won't correctly propagate?

JakeWharton:
Also if you're fixing something please add a change log entry to accompany it.

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
dellisd: As far as I can tell the unhandled exceptions still propagate up to the appLifecycle scope's job and are handled by its exception handler--but you're right, a supervisor job is probably not correct here.

We just need some per-UI scope that can be cancelled independently of the app's overall scope.

JakeWharton: Yeah sorry the "this" was meant solely to be about the `SupervisorJob`. The subscope makes sense 👍

redwood-treehouse-guest/src/commonMain/kotlin/app/cash/redwood/treehouse/treehouseCompose.kt
```diff
@@ -55,6 +58,11 @@ private class RedwoodZiplineTreehouseUi(
    */
   override val scope = (treehouseUi as? ZiplineScoped)?.scope ?: ZiplineScope()
 
+  private val coroutineScope = CoroutineScope(
+    appLifecycle.coroutineScope.coroutineContext +
+      appLifecycle.coroutineScope.coroutineContext.job,
+  )
```
JakeWharton:
```suggestion
  private val coroutineScope = CoroutineScope(
    appLifecycle.coroutineScope.coroutineContext +
      Job(appLifecycle.coroutineScope.coroutineContext.job),
  )
```
I think you probably still want to put in a new `Job` which gets canceled. Or does creating a new `CoroutineScope` do that automatically? If so, you probably don't need the `+ ...` part.

dellisd: Oh woops, that's right. A new Job is only created if there's no Job in the context. This scope's job should be a child of the parent scope's job, but this is giving both scopes the same job. Will fix!

PR #1935, opened 2024-04-04
In UIViewLazyList, don't remove subviews from hierarchy during prepareForReuse call

dnagler:
Now that view recycling is supported, we should not be pre-emptively blowing away widget views

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Fixed:
 - In `UIViewLazyList`, adding `beginUpdates`/`endUpdates` calls to `insertRows`/`deleteRows`, and wrapping changes in `UIView.performWithoutAnimation` blocks.
 - Fix memory leak in 'protocol-guest' and 'protocol-host' where child nodes beneath a removed node were incorrectly retained in an internal map indefinitely.
 - Ensure that Zipline services are not closed prematurely when disposing a Treehouse UI.
+- In `UIViewLazyList`, don't remove subviews from hierarchy during `prepareForReuse` call
 
 
 ## [0.9.0] - 2024-02-28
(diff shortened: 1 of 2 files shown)


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -296,7 +296,6 @@ internal class LazyListContainerCell(
 
   override fun prepareForReuse() {
     super.prepareForReuse()
-    removeAllSubviews()
```
JakeWharton, line 299: Will this break usage without reuse?

I think we can separately discuss making reuse completely automatic for the lazy layouts, but today we need to support both ways.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Fixed:
 - In `UIViewLazyList`, adding `beginUpdates`/`endUpdates` calls to `insertRows`/`deleteRows`, and wrapping changes in `UIView.performWithoutAnimation` blocks.
 - Fix memory leak in 'protocol-guest' and 'protocol-host' where child nodes beneath a removed node were incorrectly retained in an internal map indefinitely.
 - Ensure that Zipline services are not closed prematurely when disposing a Treehouse UI.
+- In UIViewLazyList, don't remove subviews from hierarchy during prepareForReuse call
```
JakeWharton:
```suggestion
- In `UIViewLazyList`, don't remove subviews from hierarchy during `prepareForReuse` call.
```

JakeWharton (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: This should be fine without recycling, since we're already fairly aggressively converted "loaded" cells to placeholders when they're prepared for reuse

swankjesse: I think this was unnecessary code!

If `binding` was non-null, `binding.unbind()` will immediately call removeAllSubviews() anyway.

If binding was null, then it didn’t have any subviews to begin with.

PR #1939, opened 2024-04-04
Expose guest Redwood version to host

### LINKED ISSUE #1901: Consider exposing Redwood version bi-directionally across Treehouse bridge

JakeWharton:
In order to ensure the guest is fully setup such that we can query this, we have to defer creating the host-side bridge until we have received some changes.

Closes #1901

---

### DIFF
--- a/redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeAppService.kt
+++ b/redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeAppService.kt
@@ -15,6 +15,9 @@
  */
 package app.cash.redwood.treehouse
 
+import app.cash.redwood.protocol.RedwoodVersion
+import app.cash.redwood.protocol.host.hostRedwoodVersion
+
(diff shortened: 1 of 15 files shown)


redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeAppService.kt
```diff
@@ -26,6 +29,10 @@ internal class FakeAppService private constructor(
     get() = mutableUis.toList()
 
   override val appLifecycle = object : AppLifecycle {
+    override val guestProtocolVersion: RedwoodVersion
+      // Use latest host version as the guest version to avoid any compatibility behavior.
```
swankjesse, line 33: Neat!

swankjesse (approved):

PR #1944, opened 2024-04-08
In UIViewLazyList's UITableView, adding special-case handling for programmatic scroll-to-top calls

dnagler:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ New:
 - Nothing yet!
 
 Changed:
-- Nothing yet!
+- In `UIViewLazyList`'s `UITableView`, adding special-case handling for programmatic scroll-to-top calls.
 
 Fixed:
 - Work around a problem with our memory-leak fix where our old LazyList code would crash when its placeholders were unexpectedly removed.
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -37,6 +37,8 @@ import kotlinx.cinterop.CValue
 import kotlinx.cinterop.ObjCClass
 import kotlinx.cinterop.convert
 import kotlinx.cinterop.readValue
+import kotlinx.cinterop.useContents
+import platform.CoreGraphics.CGPoint
 import platform.CoreGraphics.CGRect
 import platform.CoreGraphics.CGRectZero
 import platform.CoreGraphics.CGSize
@@ -56,17 +58,29 @@ import platform.UIKit.UITableViewDataSourceProtocol
 import platform.UIKit.UITableViewDelegateProtocol
 import platform.UIKit.UITableViewRowAnimationNone
 import platform.UIKit.UITableViewScrollPosition
+import platform.UIKit.UITableViewStyle
 import platform.UIKit.UIView
 import platform.UIKit.indexPathForItem
 import platform.UIKit.item
 import platform.darwin.NSInteger
 import platform.darwin.NSObject
 
-internal open class UIViewLazyList(
-  internal val tableView: UITableView = UITableView(
+internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {
+  internal val tableView: UITableView = object : UITableView(
     CGRectZero.readValue(),
-  ),
-) : LazyList<UIView>, ChangeListener {
+    UITableViewStyle.UITableViewStylePlain,
+  ) {
+    override fun setContentOffset(contentOffset: CValue<CGPoint>, animated: Boolean) {
+      // If the caller is requesting a contentOffset with y == 0,
+      // and the current contentOffset.y is not 0,
+      // assume that it's a programmatic scroll-to-top call.
+      if (contentOffset.useContents { y } == 0.0 && this.contentOffset.useContents { y } != 0.0) {
+        isDoingProgrammaticScroll = true
+      }
+      super.setContentOffset(contentOffset, animated)
+    }
+  }
+
   override var modifier: Modifier = Modifier
 
   override val value: UIView
@@ -183,6 +197,10 @@ internal open class UIViewLazyList(
       }
 
       override fun scrollViewDidEndScrollingAnimation(scrollView: UIScrollView) {
+        isDoingProgrammaticScroll = false
+      }
+
+      override fun scrollViewDidScrollToTop(scrollView: UIScrollView) {
         isDoingProgrammaticScroll = false
       }
     }


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -183,7 +197,13 @@ internal open class UIViewLazyList(
       }
 
       override fun scrollViewDidEndScrollingAnimation(scrollView: UIScrollView) {
+        val wasDoingProgrammaticScroll = isDoingProgrammaticScroll
         isDoingProgrammaticScroll = false
+        
+        // Notify the scrollProcessor as needed
+        if (wasDoingProgrammaticScroll) {
+          scrollViewDidScroll(scrollView)
```
dnagler: I know that this method [should not be called for programmatic scrolls](https://github.com/cashapp/redwood/blob/7925bec1d6b135be5975a1d54ee36b6a3381a3cd/redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListScrollProcessor.kt#L56), but without this call, the placeholders at the top of the scrollView remain in place after the scroll animation finishes, without being swapped out for rendered widgets.

_With_ this call in place, we see two animations: the scroll itself (up to the top of the scrollView), followed by a second animation where the "wall of placeholders" being shown collapses upward, to be replaced with rendered widgets.

Is there a better hook to call here? Ideally the placeholders would swap out for rendered widgets without the second animation  after the scroll finishes.

dnagler: cc @swankjesse

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Changed:
 - The 'app.cash.redwood.generator.compose.protocol' and 'app.cash.redwood.generator.widget.protocol' Gradle plugins are now deprecated and will be removed in the next release. Use 'app.cash.redwood.generator.protocol.guest' and 'app.cash.redwood.generator.protocol.host', respectively.
 - The 'redwood-tooling-codegen' CLI flags for protocol codegen have changed from `--compose-protocol` and `--widget-protocol` to `--protocol-guest` and `--protocol-host`, respectively.
 - Entrypoints to the protocol on the host-side and guest-side now require supplying the version of Redwood in use on the other side in order to ensure compatibility and work around any bugs in older versions. This uses a new `RedwoodVersion` type, and will be automatically wired if using our Treehouse artifacts.
+- In `UIViewLazyList`'s `UITableView`, adding special-case handling for programmatic scroll-to-top calls
```
JakeWharton: This needs moved up into the "Unreleased" section. Also, period at the end, please!

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -56,17 +58,29 @@ import platform.UIKit.UITableViewDataSourceProtocol
...
-  ),
-) : LazyList<UIView>, ChangeListener {
+    UITableViewStyle.UITableViewStylePlain,
+  ) {
+    override fun setContentOffset(contentOffset: CValue<CGPoint>, animated: Boolean) {
+      // If the caller is requesting a contentOffset with y == 0,
+      // and the current contentOffset.y is not 0,
+      // assume that it's a programmatic scroll-to-top call.
+      if (contentOffset.useContents { y } == 0.0 && this.contentOffset.useContents { y } != 0.0) {
+        isDoingProgrammaticScroll = animated
```
JakeWharton: Based on the comment I would have assumed this would be set to `true`. Or maybe to change the value passed to `super` to be `true`?

JakeWharton (approved):
Would prefer Jesse take a look still

underscoretang (approved):
Just to confirm before merging. Did you test this on the money app from regressions?

dnagler:
> Just to confirm before merging. Did you test this on the money app from regressions?

I'm not seeing anything happen at all when tapping on the tab bar icon for the Money tab--looks like scroll-to-top behavior hasn't been implemented on the client side yet?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: I did this to be consistent with [this line](https://github.com/cashapp/redwood/blob/7925bec1d6b135be5975a1d54ee36b6a3381a3cd/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt#L117), but yeah, I'd rather have it always be `true`. Will change.

CHANGELOG.md
dnagler: Fixed! Think this might have gotten caught in a rebase.

underscoretang:
Oh yeah, the tab button isn't hooked up (not sure if that's part of the plan, but does tapping to status bar still work?

It will scroll to top for us rn without doing anything extra

dnagler:
> Oh yeah, the tab button isn't hooked up (not sure if that's part of the plan, but does tapping to status bar still work?
> 
> It will scroll to top for us rn without doing anything extra

Yep, that still works as expected!

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: I played around with some alternatives, replacing the `scrollViewDidScroll(scrollView)` call with each of the following:

- `updateProcessor.onEndChanges()`
- `(scrollView as? UITableView)?.reloadData()`
- `scrollView.setNeedsDisplay()`

but all of them kept the "wall of placeholders" in place. I wish we could do something akin to the `lazyPagingItems.refresh()` call we're doing in the Treehouse layer, which repopulates the feed without animation.

dnagler: Marginal improvement: rather than keeping track of `wasDoingProgrammaticScroll`, it's better to just use the native [scrollViewDidScrollToTop](https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619382-scrollviewdidscrolltotop) hook (which is actually fired earlier than this hook).

This feels better, but the above concern (needing to get rid of the second animation) remains to be addressed.

dnagler:
> Would prefer Jesse take a look still

Screenshot 2024-04-09 at 9 23 15 PM
Chatted offline with Jesse--will pair with him on this tomorrow, but this change is good to land as-is, as an incremental improvement.

PR #1948, opened 2024-04-09
Work around a bug where LazyList.placeholders crash on remove

swankjesse:
Unfortunately this requires a change to generated code. But it only changes the behavior when talking to older apps that don't have the proper fix to the memory leak.

---


JakeWharton (approved):
Nice. Let's add to #1539?

dnagler (approved):
Thanks!

PR #1949, opened 2024-04-09
Do not query guest version from UI thread

JakeWharton:
I'm not sure how I ended up committing the previous version as I specifically was making changes specifically to avoid this, but I did. Instead, read the version once on startup when the app lifecycle is first bound and cache its value for later access on the UI thread.

---


dnagler (approved):
Confirmed via running Cash App locally--this fixes the crash on launch. Thanks for the quick turnaround here!

JakeWharton:
Nice. Should also fix CI problems for Android. Same root cause.

PR #1956, opened 2024-04-10
When performing a programmatic scroll-to-top in UIViewLazyList, alert the LazyListScrollProcessor so that it can update its viewport early

dnagler:
Follow-up to PR #1944


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -199,10 +200,6 @@ internal open class UIViewLazyList() : LazyList<UIView>, ChangeListener {
       override fun scrollViewDidEndScrollingAnimation(scrollView: UIScrollView) {
         isDoingProgrammaticScroll = false
       }
-
-      override fun scrollViewDidScrollToTop(scrollView: UIScrollView) {
```
dnagler, line 203: This hook is called well in advance of `scrollViewDidEndScrollingAnimation`; we want to continue ignoring `scrollViewDidScroll` events until the scroll is completed, now that the `scrollProcessor` already knows "where the puck is going" (to quote @swankjesse).

swankjesse (approved):

PR #1962, opened 2024-04-11
Use SparseList.addRange and removeRange

swankjesse:
These are potentially significantly more efficient than item-by-item updates, because they take advantage of the sparseness of the list.


dnagler (approved):
LGTM as long as the unit tests are still happy!

dnagler:
Tested E2E in Cash, performance feels good

PR #1969, opened 2024-04-15
Don't crash when loaded rows are removed.

swankjesse:
We had a bug where we asynchronously detached a LazyList item from its parent, but synchronously recycled it. We didn't see this occur in practice very much because the bug only triggered when the removed row was in the middle of the loaded range. Rows at the top or bottom of the loaded range were synchronously turned into placeholders when they are removed from the loaded range.

One tough limitation of this approach is the removed items lose their content immediately, which may interfere with delete animations. This isn't an urgent problem for us but it suggests that we should explore an API to put items into the pool asynchronously and not the moment they're logically removed.

---

### DIFF
--- a/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
+++ b/redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
@@ -282,6 +282,9 @@ internal open class ViewLazyList private constructor(
 
         val view = value?.value
         if (view != null) {
+          require(view.parent == null) {
+            "Received $view with unexpected parent ${view.parent}; modifier=${content?.modifier}"
+          }
           view.layoutParams = createLayoutParams()
           container.addView(view)
         }
--- a/redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
+++ b/redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
@@ -237,7 +237,8 @@ public abstract class LazyListUpdateProcessor<V : Any, W : Any> {
         is Edit.Remove -> {
           for (i in edit.index until edit.index + edit.count) {
             val index = itemsBefore.size + edit.index
-            loadedItems.removeAt(edit.index)
+            val removed = loadedItems.removeAt(edit.index)
+            removed.unbind()
 
             // Publish a structural change.
             deleteRows(index, 1)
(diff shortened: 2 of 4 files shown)


redwood-lazylayout-view/src/main/kotlin/app/cash/redwood/lazylayout/view/ViewLazyList.kt
```diff
@@ -282,6 +282,9 @@ internal open class ViewLazyList private constructor(
 
         val view = value?.value
         if (view != null) {
+          require(view.parent == null) {
+            "Received $view with unexpected parent ${view.parent}; modifier=${content?.modifier}"
```
swankjesse, line 286: Just trying to get a more actionable crash when the invariant is violated

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
```diff
@@ -237,7 +237,8 @@ public abstract class LazyListUpdateProcessor<V : Any, W : Any> {
         is Edit.Remove -> {
           for (i in edit.index until edit.index + edit.count) {
             val index = itemsBefore.size + edit.index
-            loadedItems.removeAt(edit.index)
+            val removed = loadedItems.removeAt(edit.index)
+            removed.unbind()
```
swankjesse, line 241: This is the fix. This will clear the view’s content immediately.

redwood-lazylayout-widget/src/commonTest/kotlin/app/cash/redwood/lazylayout/widget/FakeUpdateProcessor.kt
```diff
@@ -95,7 +96,8 @@ class FakeUpdateProcessor : LazyListUpdateProcessor<FakeUpdateProcessor.StringCe
         scrollWindowOffset--
       } else if (index < scrollWindowLimit) {
         // Delete a cell from the scroll window. We'll replenish the window below.
-        scrollWindowCells.removeAt(index - scrollWindowOffset).binding.unbind()
+        val removed = scrollWindowCells.removeAt(index - scrollWindowOffset)
+        require(!removed.binding.isPlaceholder && removed.content == null)
```
swankjesse, line 100: This was hiding the bug. . . the fake processor was unbinding synchronously anyway. With this change it now asserts that it’s unbound.

redwood-lazylayout-widget/src/commonTest/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessorTest.kt
```diff
@@ -529,9 +528,39 @@ class LazyListUpdateProcessorTest {
...
+    processor.items.insert(2, StringWidget("K"))
+    processor.items.insert(3, StringWidget("L"))
+    processor.onEndChanges()
+
+    processor.scrollTo(6, 8)
+    assertThat(processor.toString()).isEqualTo("[6...] . . I J K L . . [...6]")
+
+    processor.items.remove(2, 1) // 'K'.
+    processor.onEndChanges()
+    assertThat(processor.toString()).isEqualTo("[6...] . . I J L . . . [...5]")
```
swankjesse, line 546: This fails without the fix

JakeWharton (approved):
Test setup paying dividends

PR #1970, opened 2024-04-16
Work around a bug where RefreshableLazyList.placeholders crashes on remove

swankjesse:
[x] Changelog already has this fix, though we didn’t know it was incomplete at the time.


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolGuestGeneration.kt
```diff
@@ -540,7 +540,10 @@ internal fun generateProtocolWidget(
     .build()
 }
 
-private val lazyListTypeNames = listOf("app.cash.redwood.lazylayout", "LazyList")
+private val placeholderParentTypeNames = listOf(
```
dnagler, line 543: Sure enough, sorry for not catching this in #1948 !

dnagler (approved):

PR #1971, opened 2024-04-16
Debug string for WidgetValue

alec-manabat:
Adds a "pretty printed" string to WidgetValue. This makes tests easier to debug


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -278,24 +279,53 @@ internal fun generateMutableWidget(schema: Schema, widget: Widget): FileSpec {
...
     enabled,
   ).hashCode()
 
-  public override fun toString(): String =
+  override fun toString(): String =
     """ButtonValue(text=$text, enabled=$enabled)"""
+
+  override fun toDebugString(): String {
+    val debugString = """
+        |ButtonValue(
```
JakeWharton: Do you think we should keep the "Value` suffix in this rendering? It's useful when referring to it in code, but I think in this rendering we might just want to use the widget name only.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -278,24 +279,53 @@ internal fun generateMutableWidget(schema: Schema, widget: Widget): FileSpec {
...
+        |  text = $text
+        |)
+        """.trimMargin()
+    val childrenDebugString = childrenLists.joinToString(prefix = " {\n", postfix = "\n}", separator
+        = "\n") {
+      it.toDebugString().prependIndent("  ")
+    }
```
JakeWharton: I'm having trouble picturing how a multi-children widget renders. It looks like maybe all of the children are put into a single set of curly braces? Meaning you wouldn't be able to differentiate between them.

You might consider making widgets with multiple children render them more like properties.
```
TestRowValue {  // Single children, renders with trailing lambda-like syntax
  MultiChildrenValue(
    front = {
      Button(..)
      Button(..)
    }
    back = {
      Text(..)
      Text(..)
    }
}
```
of course we could render all nodes like this, but it introduces an additional layer of nesting for the single-children case which is a bit annoying.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
```diff
@@ -426,6 +465,37 @@ internal fun generateWidgetValue(schema: Schema, widget: Widget): FileSpec {
             )
             .build(),
         )
+        .addFunction(
+          FunSpec.builder("toDebugString")
+            .addModifiers(OVERRIDE)
+            .returns(String::class)
+            .addCode(
+              buildCodeBlock {
```
JakeWharton: Instead of calling `.addCode` and then `buildCodeBlock` and then things like `addStatement`, you can call `apply` which will give you the ability to do `if`/`else` and call `addStatement` and the like

JakeWharton (approved):

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/testingGeneration.kt
alec-manabat: Yep, currently they would all be in one set of braces. Will try to set it up like above 👍

PR #1972, opened 2024-04-16
In UIViewLazyList, fix UInt to UIColor conversion math used for pullRefreshContentColor.

dnagler:
Previously, almost all ARGB values would result in a white UIColor, since all RGB values above 1.0 were being interpreted as 1.0.

---


dellisd (approved):
Sweet!

swankjesse (approved):
😅

PR #1981, opened 2024-04-18
First pass at a DOM Box implementation

dellisd:
I'm opening this PR primarily to get some feedback before trying to complete the implementation.

This _sort of_ works by giving the children of the box (a flex container) `absolute` positioning. The alignment modifiers (start, end, center) work as expected with the children overlapping. `stretch` on the other hand does not work!

One way I thought to get around this limitation is to have the Box set the width/height of its children to 100% if a stretch alignment is requested, but this could end up interacting poorly with sizing modifiers if the order of operations isn't controlled carefully.

---


JakeWharton (approved):
To get 100% of the features we may need to do JS-based layout rather than relying solely on CSS. There's probably not any choice, really, since behaviors like stretch rely on knowing the size along each axes of the other elements.

swankjesse (approved):
Obsoleted by https://github.com/cashapp/redwood/pull/2811 !

Issue #1692, opened 2023-11-20
iOS snapshot metadata different for local vs. committed


swankjesse:
Our iOS snapshot tool is emitting more PNG metadata than what we have in GitHub, and that is triggering unnecessary deltas on code changes. This occurs for me when running tests in Xcode 15.0.

The new metadata includes this extra line:

> Orientation **1 (Normal)**

### Before (currently on GitHub)
![Screenshot 2023-11-20 at 1 22 08 PM](https://github.com/cashapp/redwood/assets/133019/77327df0-2af2-455b-80ae-819c1db555dd)

### After (running tests locally)
![Screenshot 2023-11-20 at 1 21 58 PM](https://github.com/cashapp/redwood/assets/133019/481b66dd-e93a-4b1c-be4c-9a6ee846f46e)

JakeWharton:
See https://github.com/cashapp/redwood/pull/1988#issuecomment-2077838600

Maybe fixed? Maybe not? Let me know.

PR #1988, opened 2024-04-24
Newer iPhone, no code signing

JakeWharton:
---


colinrtwhite (approved):

JakeWharton:
Probably gotta capture new screenshots

JakeWharton:
As far as I can tell every screenshot is the same. The only thing that changes in them is the EXIF data.

Screenshot 2024-04-25 at 1 49 26 PM

Jesse also noticed that EXIF data changed in #1692. I'm not sure where the library gets its libpng from. Since the library seemingly only compares pixels it ignores this discrepancy (for better or worse).

PR #1995, opened 2024-04-29
Layout modifier screenshot tests

### LINKED ISSUE #1877: Add row/col/box screenshot tests where we remove one modifier element from a child
Ensure the image changes as expected

ahmed3elshaer:
Fixes #1877 

**This is a draft to collect feedback about the screenshot test results in View and ComposeUI.**

#### ComposeUI
- **Box**
⚠️ There is no difference in the test results whether or not the margin is assigned. The margin is always **"empty"** in the screenshots.

- **Column & Row**
✅ Screenshots recorded as expected. 

#### View
- **Box**, **Column & Row**
⚠️ There is no difference, The margin is always applied in the screenshots, even when it's **empty**

#### UIView
- **Box**
⚠️ The margin is always **"empty"** in the screenshots.

- **Column & Row**
⚠️ The margin is always **"Applied"** in the screenshots.

See https://github.com/cashapp/redwood/issues/1877#issuecomment-2067501442 or File changes for a visual explanation

Is my test approach correct? If so, should we create issues for the identified findings?


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -549,6 +549,37 @@ abstract class AbstractFlexContainerTest<T : Any> {
...
+
+  @Test
+  fun testRowWithChildModifierChanges() {
+    testContainerWithChildrenModifierChanges(FlexDirection.Row)
+  }
+
+  private fun testContainerWithChildrenModifierChanges(
+    flexDirection: FlexDirection,
+  ) {
+    assumeTrue(flexDirection in listOf(FlexDirection.Row, FlexDirection.Column))
```
JakeWharton:
```suggestion
```
Seem unnecessary. If you called it with an invalid value it would just skip the test. If it's a prerequisite we should do `assertTrue`, but I think we simply can omit any check.

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -549,6 +549,37 @@ abstract class AbstractFlexContainerTest<T : Any> {
...
+    val first = text(longText(), backgroundColor = Red)
+    first.modifier = MarginImpl(30.dp)
+
+    container.add(first)
+    container.add(text(mediumText(), backgroundColor = Green))
+    container.add(text(shortText(), backgroundColor = Blue))
+    container.onEndChanges()
+    verifySnapshot(container, "Margin")
+    first.modifier = Modifier
+    container.onEndChanges()
```
JakeWharton, line 580: In the other test you call `widget.children.onModifierUpdated(0, redColor)`. Seems like both tests should have the same behavior here. We probably want to call both `onModifierUpdated()` and `onEndChanges()` in both tests.

JakeWharton (approved):
Looks great. We can file individual bugs for each implementation whose behavior is broken once this merges.

JakeWharton:
Actually the problem with Compose UI's Box and margins on children is known: #2006. It was simply an omission on the initial PR.

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
ahmed3elshaer: I modified `TestFlexContainer` to expose the `children` property so that we can use `children.onModifierUpdated(0, widget)` inside [AbstractFlexContainerTest.kt](https://github.com/cashapp/redwood/pull/1995/files#diff-999b67f699c44ef77a269cd6e1dbe2b9537decb4d0060d8cca94d68c3e7c4c48).

Should we add `container.onEndChanges()` to `AbstractBoxTest`? This will require us to provide a `ChangeListener` implementation for the `Box` widget under test. In my opinion, this is a significant change. So, should we stick with using `children.onModifierUpdated` in both tests or work to support the `ChangeListener` for the `Box` widget?

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
ahmed3elshaer: omitted the check ✅

ahmed3elshaer:
Updated the description with **UIView** findings

underscoretang:
this broke something for me locally when i pulled main. i can't seem to undirty theses images, no matter what i do.

screenshot stang 2024 05 09 15 21 32

maybe some git-lfs or .gitignore setup i don't have. any ideas?

swankjesse:
I tried [this](https://stackoverflow.com/questions/46704572/git-error-encountered-7-files-that-should-have-been-pointers-but-werent) and it worked?

```
git rm --cached -r .
git reset --hard
```

JakeWharton:
The pointfree library has some non-determinism. I don't know why.

Explored a bit here: https://github.com/cashapp/redwood/pull/1988#issuecomment-2077838600

PR #1996, opened 2024-04-29
In YogaUIView's setScrollEnabled method, only call setNeedsLayout if the scrollEnabled value is actually changing.

dnagler:


colinrtwhite (approved):
Nice!

PR #1999, opened 2024-05-01
In YogaUIView's layoutNodes method, return early for nested YogaUIViews to prevent redundant frame calculations.

dnagler:
Follow-up to PR #1996

---


swankjesse (approved):

underscoretang (approved):
Whoa. Nice! Just FYI. We've been digging into bugs here where we're not getting enough layout passes for views that get more children and should resize but don't. Not sure how this will play into that

@aridokmecian @edahle

dnagler:
> Whoa. Nice! Just FYI. We've been digging into bugs here where we're not getting enough layout passes for views that get more children and should resize but don't. Not sure how this will play into that
> 
> @aridokmecian @edahle

Interesting--this was found while trying to solve the opposite problem. When a `YogaUIView` is nested within another `YogaUIView` (in the case I was investigating, it was a `Column` within a `Row`), then the child view calls `layoutSubviews` twice in a row on appearance, which is totally unnecessary--I made some local changes so that the second call no-op'ed, and the resulting layout was still as expected.

This double `layoutSubviews` call is still occurring with this change in place, but this felt like a step in the right direction regardless.

PR #2021, opened 2024-05-10
Fix screenshots not being stored in LFS.

colinrtwhite:


JakeWharton (approved):
Can we detect this in CI somehow for the future?

PR #2025, opened 2024-05-13
Add a CI check to enforce that files use git lfs when necessary.

colinrtwhite:
Follow up to https://github.com/cashapp/redwood/pull/2021#pullrequestreview-2050927580.

### DIFF
--- a/scripts/enforce_git_lfs.sh
+++ b/scripts/enforce_git_lfs.sh
@@ -3,7 +3,7 @@
 output=$(git checkout .)
 
 if echo "$output" | grep -q "files that should have been pointers, but weren't"; then
-    echo "Error: Encountered files that should be checked in using Git LFS, but weren't."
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):

scripts/enforce_git_lfs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
JakeWharton:
```suggestion
#!/usr/bin/env bash
```

scripts/enforce_git_lfs.sh
```diff
@@ -0,0 +1,10 @@
...
+
+output=$(git checkout .)
+
+if echo "$output" | grep -q "files that should have been pointers, but weren't"; then
```
JakeWharton: You can do a substring match natively in bash with https://stackoverflow.com/a/229606/132047 if you want

PR #2032, opened 2024-05-16
Reenable the debug build type for libraries.

colinrtwhite:
Looking for feedback on this. We need the `debug` build type for use with `includeBuild`.

I explored only enabling `debug` if redwood is used as part of `includeBuild` in another repo, but couldn't find an easy way (the projects maintain separate properties maps).


JakeWharton (approved):
Ugh. Makes sense but I don't have to like it!

Lines 193-198 can also be deleted then.

PR #2033, opened 2024-05-16
Break more cycles in widget bindings

swankjesse:
The ProtocolNode is a Kotlin object and the Widget is potentially a Swift object. This introduces a potential retain cycle problem.

---


redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ListeningTestRow.kt
```diff
@@ -47,6 +47,9 @@ class ListeningTestRow : TestRow<WidgetValue>, ChangeListener {
 
     override fun onModifierUpdated(index: Int, widget: Widget<WidgetValue>) {
     }
+
+    override fun detach() {
```
swankjesse, line 51: Not doing any detach work for Compose UI until we need it. Especially since it doesn’t have a separate list for its views!

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
```diff
@@ -48,7 +48,7 @@ public abstract class LazyListUpdateProcessor<V : Any, W : Any> {
    * The first placeholder ever returned. We use it to choose measured dimensions for created
    * placeholders if the pool ever runs out.
    */
-  private var firstPlaceholder: Widget<W>? = null
+  private var firstPlaceholder: W? = null
```
swankjesse, line 51: The easiest way to clear a widget reference is to not have a widget reference to begin with.

This file might require a follow-up as our individual row bindings will have a Widget object. I need to validate that in the app once I’ve got the new `detach()` function in place.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/ProtocolBridge.kt
```diff
@@ -164,17 +164,23 @@ public class ProtocolBridge<W : Any>(
    */
   public fun close() {
     closed = true
+    for (node in nodes.values) {
+      node.detach()
+    }
     nodes.clear()
     pool.clear()
```
swankjesse, line 176: We should detach these

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/Widget.kt
```diff
@@ -69,5 +69,20 @@ public interface Widget<W : Any> {
...
+     * After this is called there will be no further calls to insert, remove, or move widgets, and
+     * no further updates to widgets' modifiers.
+     *
+     * Implementations must clear all direct references to [Widget] instances. This is necessary
+     * because widget implementations may be reference-counted (as in Swift) and may also contain
+     * references to Kotlin objects to create a retain cycle. We require implementations of this
+     * class to be where this cycle is broken because it's a non-invasive place to do so.
+     *
+     * This function is not recursive.
+     */
```
JakeWharton, line 85: I'm a bit concerned how this works (or doesn't work) if you aren't using the protocol. It seems like non-protocol widget systems should do a recursive detach, whereas protocol widget systems on the guest could no-op and leave the recursion to happen in the protocol host code, as it does now.

It's fine as-is for now, but I'm going to file an issue for the automated leak detection, and I suspect when we do that on plain Redwood usage on iOS we'd see leaks if entire subtrees were removed since only the root widget would see detach and break its reference cycle.

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolHostGeneration.kt
```diff
@@ -251,23 +251,24 @@ internal class ProtocolButton<W : Any>(
...
-  private var _widget: Button<W> = widget
-  override val widget: Widget<W> get() = _widget
+  private var _widget: Button<W>? = widget
+  override val widget: Widget<W> get() = _widget ?: error("detached")
 
   private val serializer_0: KSerializer<String?> = json.serializersModule.serializer()
   private val serializer_1: KSerializer<Boolean> = json.serializersModule.serializer()
 
   public override fun apply(change: PropertyChange, eventSink: EventSink): Unit {
+    val widget = this.widget
```
JakeWharton: Not that it matters all that much...
```suggestion
    val widget = _widget ?: error("detached")
```

JakeWharton (approved):

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/Widget.kt
swankjesse: Good catch. Yeah I totally missed that.

JakeWharton: Yeah. Future problems, though.

PR #2042, opened 2024-05-19
Don't crash reading modifiers after widgets are detached

swankjesse:
We support rendering a widget after Redwood detaches from it. Make this work by never assuming the widget count equals the subject view's child count.

---

### DIFF
--- a/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
+++ b/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
@@ -364,6 +364,26 @@ abstract class AbstractBoxTest<T : Any> {
     verifySnapshot(widget.value, "Empty")
   }
 
+  /** The view shouldn't crash if its displayed after being detached. */
+  @Test
+  fun testLayoutAfterDetach() {
+    val widget = box().apply {
+      width(Constraint.Wrap)
+      height(Constraint.Wrap)
+      horizontalAlignment(CrossAxisAlignment.Start)
+      verticalAlignment(CrossAxisAlignment.Start)
+      children.insert(
+        0,
+        text().apply {
+          text(mediumText())
+          bgColor(Green)
+        },
+      )
+      children.detach()
+    }
+    verifySnapshot(widget.value)
+  }
+
   private fun coloredText(modifier: Modifier = Modifier, text: String, color: Int) = text().apply {
     text(text)
     bgColor(color)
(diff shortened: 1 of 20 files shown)


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
```diff
@@ -364,6 +364,26 @@ abstract class AbstractBoxTest<T : Any> {
...
+  @Test
+  fun testLayoutAfterDetach() {
+    val widget = box().apply {
+      width(Constraint.Wrap)
+      height(Constraint.Wrap)
+      horizontalAlignment(CrossAxisAlignment.Start)
+      verticalAlignment(CrossAxisAlignment.Start)
+      children.insert(
+        0,
+        text().apply {
```
underscoretang, line 377: I think a good test to add is one with a bunch of children, all with different modifiers to make sure they don't clobber each other or get shared when we don't want them to. (specifically .height and .width ones)

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -31,26 +31,32 @@ import platform.darwin.NSInteger
 internal class UIViewFlexContainer(
   direction: FlexDirection,
 ) : YogaFlexContainer<UIView>, ChangeListener {
+  private val modifiers = mutableListOf<Modifier>()
   private val yogaView: YogaUIView = YogaUIView(
     applyModifier = { node, index ->
-      node.applyModifier(children.widgets[index].modifier, Density.Default)
+      node.applyModifier(modifiers[index], Density.Default)
```
underscoretang, line 37: I'm a little unclear on this. Wasn't the previous code looking at the individual widgets modifiers? but are the modifiers we're saving into this mutableList() the same here.

underscoretang (approved):

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
swankjesse: Great idea

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
swankjesse: Yeah, we’re now keeping our own private copy of the modifiers because keeping a reference to the Widget creates a retain cycle.

redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt
swankjesse: I’m gonna send a PR with this test as a fast-follow-up!

swankjesse: https://github.com/cashapp/redwood/pull/2045

PR #2047, opened 2024-05-21
Don't store modifiers in an adjacent List

swankjesse:
There's a 'context' field in Node that's a better fit.

The problem with storing modifiers in an adjacent list is that Yoga occasionally will create a mutable copy of the layout nodes to do some layout calculation, and this mutable copy may drift from our true copy with respect to child counts.

---


underscoretang:
sneaky yoga, starting to feel more like pilates

underscoretang (approved):

colinrtwhite (approved):
This looks good to me :+1:

PR #2050, opened 2024-05-22
TreehouseTester

swankjesse:
This is a new testing facet that launches real code with a real zipline.

In a follow-up PR I intend to put this to work to confirm that we don't hold references to various types beyond the required lifetime.


redwood-treehouse-host/src/appsTest/kotlin/app/cash/redwood/treehouse/TreehouseTesterTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    content.bind(view)
+
+    content.awaitContent(1)
+    val textInputValue = view.views.single() as TextInputValue
+    assertThat(textInputValue.text).isEqualTo("what would you like to see?")
+    textInputValue.onChange!!.invoke("TreehouseTesterTestHappyPathStep2")
+
+    tester.sendFrame()
+    content.awaitContent(2)
```
JakeWharton, line 51: Is this waiting one change or two? I can't tell if the count is kept and so you have to increment it each time, or if it's reset and you could use 1 over and over if you wanted.

redwood-treehouse-host/build.gradle
```diff
@@ -35,10 +35,21 @@ kotlin {
...
+        implementation projects.redwoodLayoutTesting
+        implementation projects.redwoodLazylayoutTesting
+        implementation projects.testApp.schema.protocolHost
+        implementation projects.testApp.schema.testing
+      }
+      if (!rootProject.hasProperty('redwoodNoApps')) {
+        kotlin.srcDir('src/appsTest/kotlin')
+        dependencies {
+          implementation projects.testApp.presenterTreehouse
+        }
```
JakeWharton, line 52: I'm pretty sure this means this code isn't going to be tested on CI. You would need to add `:redwood-treehouse-host:check` to the 'test-app' shard.

redwood-treehouse-host/build.gradle
```diff
@@ -80,3 +95,10 @@ android {
     androidResources = true
   }
 }
+
+// TreehouseTester expects to load our test application.
+if (!rootProject.hasProperty('redwoodNoApps')) {
+  tasks.withType(Test.class).configureEach {
+    it.dependsOn(":test-app:presenter-treehouse:compileDevelopmentExecutableKotlinJsZipline")
```
JakeWharton, line 102: 😬

JakeWharton (approved):

redwood-treehouse-host/build.gradle
swankjesse: Ooooh good catch

swankjesse: On my follow-up list.

redwood-treehouse-host/src/appsTest/kotlin/app/cash/redwood/treehouse/TreehouseTesterTest.kt
swankjesse: It’s waiting for version 2 of the content.

swankjesse: (I’ll find a way to make this API more clear.)

redwood-treehouse-host/build.gradle
swankjesse: https://github.com/cashapp/redwood/pull/2065

PR #2051, opened 2024-05-23
LeaksTest

swankjesse:
I recently worked on explicitly breaking reference cycles to prevent objects from leaking when we mix garbage-collected Kotlin objects with reference-counted Swift objects.

But this work was likely to regress because we didn't have a mechanism to prevent these reference cycles from recurring.

This PR introduces a bunch of machinery to explicitly test for reference cycles. It works on the JVM because that's a capable platform for this kind of dynamic analysis, and because it's consistent with the Kotlin/Native platform that is actually where we need to defend against for leaks.

The first new class is JvmHeap, which lazily inspects an object for its outbound references. This correctly captures the Kotlin compiler-inserted indirect references. It also includes lots of special cases to avoid traversing into implemnetation details of kotlinx.serialization and coroutines.

The second new class is CycleFinder, which is a basic Dijkstra breadth-first search. It's one neat feature is that it can reconstruct the list of properties that participate in a cycle. The cycle in LeaksTest (before the widget is removed) is:

         * callHandler
           * endpoint
             * inboundServices[3]
               * value
                 * service
                   * viewOrNull
                     * mutableListChildren
                       * container[0]
                         * onChange
                           * receiver
                             * eventSink
                               * delegate

This PR adds a test for widgets leaking. In a follow-up PR I intend to add tests for other interfaces and extension points in Redwood that can be implemented in Swift. This includes event listeners, Zipline services, TreehouseContentSource, and CodeListener.

### DIFF
--- a/redwood-treehouse-host/build.gradle
+++ b/redwood-treehouse-host/build.gradle
@@ -58,6 +58,11 @@ kotlin {
         implementation libs.robolectric
       }
     }
+    jvmTest {
+      if (!rootProject.hasProperty('redwoodNoApps')) {
+        kotlin.srcDir('src/appsJvmTest/kotlin')
+      }
+    }
   }
 }
 
--- a/redwood-treehouse-host/src/appsJvmTest/kotlin/app/cash/redwood/treehouse/leaks/JvmHeap.kt
+++ b/redwood-treehouse-host/src/appsJvmTest/kotlin/app/cash/redwood/treehouse/leaks/JvmHeap.kt
@@ -49,16 +49,16 @@ internal object JvmHeap : Heap {
       javaClass.isArray -> {
         when (instance) {
           is Array<*> -> references(instance.toList())
-          else -> listOf()
+          else -> listOf() // Primitive array.
         }
       }
-      instance is Collection<*> && javaPackageName.isPlatformPackage -> {
+      instance is Collection<*> && javaPackageName.isDescendant("java", "kotlin") -> {
         instance.mapIndexed { index, value -> Edge("[$index]", value) }
       }
-      instance is Map<*, *> && javaPackageName.isPlatformPackage -> {
+      instance is Map<*, *> && javaPackageName.isDescendant("java", "kotlin") -> {
         references(instance.entries)
       }
-      instance is Map.Entry<*, *> && javaPackageName.startsWith("java.") -> listOf(
+      instance is Map.Entry<*, *> && javaPackageName.isDescendant("java", "kotlin") -> listOf(
         Edge("key", instance.key),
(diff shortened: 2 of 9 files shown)


redwood-treehouse-host/src/appsJvmTest/kotlin/app/cash/redwood/treehouse/leaks/JvmHeap.kt
```diff
@@ -0,0 +1,119 @@
...
+      instance is EventLog -> listOf()
+      instance is Int -> listOf()
+      instance is Job -> listOf()
+      instance is Json -> listOf()
+      instance is SerializersModule -> listOf()
+      instance is KSerializer<*> -> listOf()
+      instance is Class<*> -> listOf()
+      instance is String -> listOf()
+
+      // Don't traverse further on types that are unlikely to contain application-scoped data.
```
swankjesse:
```suggestion
      // Explore everything else by reflecting on its fields.
```

redwood-treehouse-host/src/appsJvmTest/kotlin/app/cash/redwood/treehouse/leaks/JvmHeapTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import assertk.assertions.isEmpty
+import kotlin.test.Test
+
+class JvmHeapTest {
+  @Test
+  fun happyPath() {
+    val album = Album(
+      name = "Lateralus",
+      artist = Artist("Tool", 1990),
+      releaseYear = 2001,
```
JakeWharton, line 29: 🎸

redwood-treehouse-host/build.gradle
```diff
@@ -58,6 +58,11 @@ kotlin {
         implementation libs.robolectric
       }
     }
+    jvmTest {
+      if (!rootProject.hasProperty('redwoodNoApps')) {
+        kotlin.srcDir('src/appsJvmTest/kotlin')
+      }
+    }
```
JakeWharton, line 65: I mentioned this on the other PR, but these almost certainly aren't going to be run on CI as a result. Need to add the `check` task for this module to the 'test-app' shard.

JakeWharton (approved):

redwood-treehouse-host/build.gradle
swankjesse: Yep, on it.

swankjesse: Done. https://github.com/cashapp/redwood/pull/2065

PR #2052, opened 2024-05-23
Test that we don't leak services

swankjesse:
This reveals that we don't expose enough state information on TreehouseApp.


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/TreehouseApp.kt
```diff
@@ -74,8 +75,8 @@ public class TreehouseApp<A : AppService> private constructor(
    * It is unwise to use this instance for anything beyond measurement and monitoring, because the
    * instance may be replaced if new code is loaded.
    */
-  public val zipline: Zipline?
-    get() = (codeHost.codeSession as? ZiplineCodeSession)?.zipline
+  public val zipline: StateFlow<Zipline?>
```
swankjesse, line 78: I want a StateFlow.map {} function

JakeWharton (approved):

PR #2067, opened 2024-05-30
Add support for an onScroll property to Row and Column.

colinrtwhite:
Adds an `onScroll` callback to `Row` and `Column`. There was interest to add this so we can show/hide UI elements based on the scroll offset.

---


redwood-layout-schema/src/main/kotlin/app/cash/redwood/layout/widgets.kt
```diff
@@ -80,6 +80,16 @@ public data class Row(
...
 
+  /**
+   * Invoked when the container scrolls. The function's `offset` is represented in units in the
+   * host's coordinate system.
+   *
+   * @see Overflow.Scroll
+   */
+  @Property(7)
+  @Default("null")
+  val onScroll: ((offset: Double) -> Unit)?,
```
colinrtwhite, line 91: Unsure if this should be `Double` (px on Android, points on iOS) or if we should standardize by emitting `offset: Dp`.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
```diff
@@ -112,17 +121,36 @@ internal class ViewFlexContainer(
...
+
     init {
       layoutParams = LayoutParams(WRAP_CONTENT, WRAP_CONTENT)
       updateViewHierarchy()
     }
 
+    fun attachOrDetachScrollListeners() {
+      val child = getChildAt(0)
+      if (child is NestedScrollView) {
+        val listener = (onScrollListener as? OnScrollChangeListenerCompat)
```
JakeWharton:
```suggestion
        val listener = (onScrollListener as OnScrollChangeListenerCompat?)
```
This is a nit, but `as? Foo` is a cast that can fail and falls back to null. `as Foo?` is a cast to a nullable type.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
```diff
@@ -112,17 +121,36 @@ internal class ViewFlexContainer(
...
+          ?: OnScrollChangeListenerCompat { _, _, scrollY, _, _ -> onScroll?.invoke(scrollY.toDouble()) }
+            .also { onScrollListener = it }
+        child.setOnScrollChangeListener(listener)
+      } else if (SDK_INT >= 23 && child is HorizontalScrollView) {
+        val listener = (onScrollListener as? OnScrollChangeListener)
```
JakeWharton:
```suggestion
        val listener = (onScrollListener as OnScrollChangeListener?)
```

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
JakeWharton: Fewer bytecodes, too! https://kotlin.godbolt.org/z/YKz7bfao4

redwood-layout-dom/src/commonMain/kotlin/app/cash/redwood/layout/dom/HTMLElementRedwoodLayoutWidgetFactory.kt
```diff
@@ -101,6 +105,27 @@ private class HTMLFlexContainer(
...
+        override fun handleEvent(event: Event) {
+          val offset = when (value.style.flexDirection) {
+            "row" -> value.scrollTop
+            "column" -> value.scrollLeft
+            else -> throw AssertionError()
+          }
+          onScroll(offset)
+        }
+      }.also { scrollEventListener = it }
+      value.addEventListener("scroll", eventListener)
```
JakeWharton, line 125: I think this will add multiple listeners if you change the lambda. Probably need to move the remove code from the `else` branch to happen unconditionally, and then only if there's a non-null `onScroll` lambda do a set.

JakeWharton: I didn't check the view-based implementation, but if it's addXxx vs. setXxx it may also need the same treatment.

redwood-layout-composeui/src/androidUnitTest/kotlin/app/cash/redwood/layout/composeui/ComposeUiFlexContainerTest.kt
```diff
@@ -68,13 +70,24 @@ class ComposeUiFlexContainerTest(
...
       },
     )
 
+    override fun onScroll(onScroll: (Double) -> Unit) {
+      delegate.onScroll(onScroll)
+    }
+
+    override suspend fun scroll(offset: Double) {
+      delegate.scrollState?.scrollTo(offset.toInt())
+      delay(20.milliseconds)
```
JakeWharton: Seems like every impl is doing this delay. Maybe make this non-suspending and move to the test call-site instead?

JakeWharton (approved):

redwood-layout-schema/src/main/kotlin/app/cash/redwood/layout/widgets.kt
JakeWharton: Yeah not sure. I suspect (hope?) most implementations will be doing a simple `> 0` or something, because any kind of synchronized scrolling isn't really going to work

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewFlexContainer.kt
colinrtwhite: Neat!

redwood-layout-dom/src/commonMain/kotlin/app/cash/redwood/layout/dom/HTMLElementRedwoodLayoutWidgetFactory.kt
colinrtwhite: I think we already handle this case? We keep a reference to the listener added in `addEventListener` and pass it back to `removeEventListener` if `onScroll` is back to null.

redwood-layout-composeui/src/androidUnitTest/kotlin/app/cash/redwood/layout/composeui/ComposeUiFlexContainerTest.kt
colinrtwhite: I added it to the method since we can skip the delay on no-op impls like Compose UI/View `LazyList`.

redwood-layout-dom/src/commonMain/kotlin/app/cash/redwood/layout/dom/HTMLElementRedwoodLayoutWidgetFactory.kt
JakeWharton: The case I'm talking about is if the lambda changes its value.
```kotlin
var state by remember { mutableStateOf(false) }
val lambda = if (state) {
  { println("hey") }
} else {
  { println("hello") }
}
LaunchedEffect(Unit) {
  while (true) {
    delay(1.seconds)
    state = !state
  }
}

Column(onScroll = lambda)
```
This code will never send a null, but will send a lambda different from the existing one every second.

colinrtwhite: Ah gotcha! Fixed

colinrtwhite:
Hmm using `runTest` requires Paparazzi >= 1.3.3, but upgrading to that introduces other issues.

JakeWharton:
> Hmm using `runTest` requires Paparazzi >= 1.3.3, but upgrading to that introduces other issues.

Is it the same failure in #1769? There's workarounds for it if so. I was just being lazy and not applying them.

JakeWharton:
Starting to unblock Paparazzi upgrades here: https://github.com/cashapp/redwood/pull/2079

colinrtwhite:
@JakeWharton Yep, though I managed to work around needing `runTest` by using `runBlocking` for Compose UI. I'll take a look at the 1.3.4 Compose UI double measurement Paparazzi error this afternoon.

PR #2070, opened 2024-05-31
Unify multiplatform targets into named groups

JakeWharton:
This gives us more control over what targets are built for which parts of the codebase based on their type.

The big change is the drop of MacOS targets which were wholly unused. They will likely return at some point, along with new targets for common modules as well as for Compose UI, and this new grouping will make that easy to do.

Before:

    $ gw build --dry-run | grep -c SKIPPED
    7469

After:

    $ gw build --dry-run | grep -c SKIPPED
    6405

Refs #361 

---


JakeWharton:
This is like half of what I actually want to build but I'm sick of waffling so this is a small step in that direction.

colinrtwhite (approved):
Nice!

PR #2071, opened 2024-05-31
Always build test-app presenter

JakeWharton:
I would rather take the slight build hit than maintain the conditional logic and extra CI job.

---


colinrtwhite (approved):

.github/workflows/build.yaml
```diff
@@ -85,19 +85,6 @@ jobs:
 
       - run: xcodebuild -project redwood-layout-uiview/RedwoodLayoutUIViewTests.xcodeproj -scheme RedwoodLayoutUIViewTests -destination 'platform=iOS Simulator,name=iPhone 15,OS=latest' test
 
-  treehouse-host-check:
```
swankjesse, line 167: Good call!

swankjesse (approved):

swankjesse:
Ugh, we can build the development executable, or the production executable, but not both in the same run.

I guess we need to decide which one we run in CI.

```
* What went wrong:
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
A problem was found with the configuration of task ':test-app:presenter-treehouse:jsBrowserProductionWebpack' (type 'KotlinWebpack').
  - Gradle detected a problem with the following location: '/Users/runner/work/redwood/redwood/build/js/packages/test-app-presenter-treehouse'.
    
    Reason: Task ':test-app:presenter-treehouse:jsBrowserProductionWebpack' uses this output of task ':test-app:presenter-treehouse:jsDevelopmentExecutableCompileSync' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':test-app:presenter-treehouse:jsDevelopmentExecutableCompileSync' as an input of ':test-app:presenter-treehouse:jsBrowserProductionWebpack'.
      2. Declare an explicit dependency on ':test-app:presenter-treehouse:jsDevelopmentExecutableCompileSync' from ':test-app:presenter-treehouse:jsBrowserProductionWebpack' using Task#dependsOn.
      3. Declare an explicit dependency on ':test-app:presenter-treehouse:jsDevelopmentExecutableCompileSync' from ':test-app:presenter-treehouse:jsBrowserProductionWebpack' using Task#mustRunAfter.
```

JakeWharton:
Maybe they use different directories in KGP 2.0? Can wait and see.

JakeWharton:
The distribution directories are different, but it still fails. I filed https://youtrack.jetbrains.com/issue/KT-68813 about it.

PR #2089, opened 2024-06-06
WIP: experiment skipping kotlinx.serialization for Changes

swankjesse:
---

### DIFF
--- a/redwood-protocol-guest/src/jsMain/kotlin/app/cash/redwood/protocol/guest/ProtocolStateJs.kt
+++ b/redwood-protocol-guest/src/jsMain/kotlin/app/cash/redwood/protocol/guest/ProtocolStateJs.kt
@@ -1,0 +1,167 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package app.cash.redwood.protocol.guest
+
+import app.cash.redwood.RedwoodCodegenApi
+import app.cash.redwood.protocol.Change
+import app.cash.redwood.protocol.ChildrenTag
+import app.cash.redwood.protocol.Id
+import app.cash.redwood.protocol.ModifierElement
+import app.cash.redwood.protocol.PropertyTag
+import app.cash.redwood.protocol.RedwoodVersion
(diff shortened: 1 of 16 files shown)


redwood-protocol-guest/src/jsMain/kotlin/app/cash/redwood/protocol/guest/ProtocolStateJs.kt
```diff
@@ -0,0 +1,167 @@
...
+    val value = value
+    changes.add(js("""["property",{"id":id,"tag":tag,"value":value}]"""))
+  }
+
+  public actual fun appendModifierChange(
+    id: Id,
+    elements: List<ModifierElement>,
+  ) {
+    val id = id
+    val elements = Json.encodeToDynamic(elements)
```
JakeWharton, line 93:
```suggestion
    val elements = json.encodeToDynamic(elements)
```

redwood-protocol-guest/src/jsMain/kotlin/app/cash/redwood/protocol/guest/ProtocolStateJs.kt
```diff
@@ -0,0 +1,167 @@
...
+/** @suppress For generated code use only. */
+@RedwoodCodegenApi
+@OptIn(ExperimentalSerializationApi::class)
+public actual class ProtocolState actual constructor(
+  private val json: Json,
+  hostVersion: RedwoodVersion,
+) {
+  private var nextValue = Id.Root.value + 1
+  private val widgets = PlatformMap<Int, ProtocolWidget>()
+  private var changes = PlatformList<Change>()
```
JakeWharton, line 41: This type parameter is inaccurate now. Anyone on JS trying to actually interact with the `Change` instances (from `getChangesOrNull()`) will get weird errors.

redwood-protocol-guest/src/nonJsMain/kotlin/app/cash/redwood/protocol/guest/ProtocolStateNonJs.kt
```diff
@@ -0,0 +1,146 @@
...
+import app.cash.redwood.protocol.PropertyTag
+import app.cash.redwood.protocol.RedwoodVersion
+import app.cash.redwood.protocol.WidgetTag
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonPrimitive
+
+/** @suppress For generated code use only. */
+@RedwoodCodegenApi
+public actual class ProtocolState actual constructor(
```
JakeWharton, line 36: Instead of expect/actual-ing this entire class, I think we can get away with just doing the protocol. Calls to things like `Create(id, tag)` can change from a constructor to a function which we can then use to vary the implementation across each platform. On JVM it can call an internal constructor and use the regular serializer, but on JS we can return your raw JS objects and maybe use `external interface` for the subtypes like `Create` which proxy to them.

swankjesse: I like that a lot

swankjesse:
Obsoleted by https://github.com/cashapp/redwood/pull/2106

PR #2091, opened 2024-06-07
Add `UiConfiguration.layoutDirection` and populate it on all platforms

drewhamilton:
Introduces the ability to read the host's `layoutDirection` in guest UI declarations. The Android and iOS implementations should be correct; the HTML implement is iffy.

---


redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
```diff
@@ -67,6 +68,12 @@ private class RedwoodHTMLElementView(
       UiConfiguration(
         darkMode = colorSchemeQuery.matches,
         viewportSize = Size(width = element.offsetWidth.dp, height = element.offsetHeight.dp),
+        layoutDirection = when (element.dir) {
```
drewhamilton, line 71: Not sure this will work consistently. It seems like many HTML elements inherit `dir` without explicitly specifying it. But maybe it's OK to say Redwood elements need to specify `dir` if they want layout direction to work?

redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/TreehouseUIViewTest.kt
```diff
@@ -128,4 +131,21 @@ class TreehouseUIViewTest {
...
+  @Test
+  fun uiConfigurationReflectsInitialLayoutDirection() {
+    val parent = UIWindow()
+
+    val layout = TreehouseUIView(throwingWidgetSystem)
+    parent.addSubview(layout.view)
+
+    val expectedLayoutDirection = when (val direction = parent.effectiveUserInterfaceLayoutDirection) {
+      UIUserInterfaceLayoutDirectionLeftToRight -> LayoutDirection.Ltr
+      UIUserInterfaceLayoutDirectionRightToLeft -> LayoutDirection.Rtl
```
JakeWharton, line 144: I didn't realize iOS had this. We have some missing tests for the layout, it seems! https://github.com/cashapp/redwood/issues/2092

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
drewhamilton: Also there's a bunch of caveats listed on the [documentation page](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir).

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
```diff
@@ -67,6 +68,12 @@ private class RedwoodHTMLElementView(
...
+          "ltr" -> LayoutDirection.Ltr
+          "rtl" -> LayoutDirection.Rtl
+          "auto" -> LayoutDirection.Auto
+          else -> LayoutDirection.Auto
```
drewhamilton: Reading the docs closer, I think this should default to `Ltr` instead of `Auto`

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
JakeWharton, line 71: I think this is reading the attribute as defined on this specific element, and I'm not sure it's value gets inherited from any parents.

The only thing I can find about this is using a computedStyle on the element and then reading the CSS `direction` property. But even then I don't know how to observe and react to changes in that.

We can leave this as-is for now.

JakeWharton (approved):

redwood-widget/src/jsMain/kotlin/app/cash/redwood/widget/RedwoodHTMLElementView.kt
JakeWharton: Yeah we can say you need to explicitly define it for now. No one is really using HTML seriously right now, and certainly not in a context where RTL would show up. But it's nice to not forget about it.

JakeWharton:
I was going to write a test for the HTML/JS, but... let's just merge.

PR #2098, opened 2024-06-12
Extract an interface for ProtocolState

swankjesse:
Also begin to convert this interface into a special purpose service. I'm working towards making it support an alternate mode for guest code that skips the Changes objects.

---


swankjesse:
Next steps:
 * [x] Rename `ProtocolBridge.Factory` to `WidgetSystem.Factory` and have it produce a `WidgetSystem`.
 * [x] Combine `ProtocolBridge` and `ProtocolState`, and name the resulting interface `ProtocolBridge`. The default implementation can be created by calling a `ProtocolBridge()` function.
 * [x] Create a new implementation of `ProtocolBridge` that’s optimized for Kotlin/JS and that uses [Zipline Dynamic Functions](https://github.com/cashapp/zipline/pull/1347).

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/DefaultProtocolState.kt
```diff
@@ -0,0 +1,144 @@
...
+  private var nextValue = Id.Root.value + 1
+  private val widgets = mutableMapOf<Int, ProtocolWidget>()
+  private var changes = mutableListOf<Change>()
+  private lateinit var changesSink: ChangesSink
+
+  public override val synthesizeSubtreeRemoval: Boolean = hostVersion < RedwoodVersion("0.10.0-SNAPSHOT")
+
+  override fun initChangesSink(changesSink: ChangesSink) {
+    this.changesSink = changesSink
+  }
```
JakeWharton, line 50: Eeek not my favorite. I mean I know we have to have a break in the cycle _somewhere_ for it to all work, but I prefer to pretend it doesn't exist.

JakeWharton (approved):
Needing to make this whole class an interface feels a bit like too much. Almost as if it needs some of its responsibility split between management of the nodes (ID generation, host version, widget instance map, etc.) as one class, and aggregation of the changes as the interface. I guess I'll see where it goes in the coming PRs.

swankjesse:
Yeah, I think I might best address some design / separation follow-ups after the last PR.

PR #2101, opened 2024-06-12
Combine ProtocolState and ProtocolBridge

swankjesse:
They're 1:1 with each other and there's little utility in keeping them independent.


redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ChangeListenerTest.kt
```diff
@@ -59,11 +59,8 @@ class ProtocolChangeListenerTest : AbstractChangeListenerTest() {
...
   ): TestRedwoodComposition<T> {
-    val state = DefaultProtocolState(
+    val compositionBridge = DefaultProtocolBridge(
       hostVersion = hostRedwoodVersion,
-    )
-    val compositionBridge = app.cash.redwood.protocol.guest.ProtocolBridge(
-      state = state,
       widgetSystemFactory = TestSchemaProtocolWidgetSystemFactory,
     )
     val widgetBridge = ProtocolBridge(
```
JakeWharton, line 66: Probably should be `hostBridge` and `guestBridge`. The existing `widgetBridge` likely just predates the pervasive use of host/guest terms.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolRedwoodComposition.kt
```diff
@@ -30,6 +31,7 @@ import kotlinx.coroutines.flow.StateFlow
  * @param scope A [CoroutineScope] whose [coroutineContext][kotlin.coroutines.CoroutineContext]
  * must have a [MonotonicFrameClock] key which is being ticked.
  */
+@OptIn(RedwoodCodegenApi::class)
 @Suppress("FunctionName")
 public fun ProtocolRedwoodComposition(
```
JakeWharton, line 36: This function now has a type in its public API that is reserved for codegen use only. Ideally you should be able to setup a protocol-based system relying on public API only. Maybe something fixed in future change.

JakeWharton (approved):

redwood-compose/src/commonTest/kotlin/app/cash/redwood/compose/ChangeListenerTest.kt
swankjesse: YES! Will do.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolRedwoodComposition.kt
swankjesse: I think I can do that. ‘Set up a protocol-based system with the public API only’ is a nice standard to reach for.

swankjesse:
I’m gonna address the feedback in a follow-up PR!

PR #2116, opened 2024-06-14
Require latest of snapshot testing library

JakeWharton:
This works around a bug on virtualized hosts which don't have Metal always failing to compare images.

---


JakeWharton:
We were getting the latest

PR #2118, opened 2024-06-14
Rename ProtocolBridge to ProtocolHost and ProtocolGuest

swankjesse:
I'm not in love with these names, but I do like them better than ProtocolBridge and ProtocolBridge.

I considered and rejected some others:
 - HostCaller - I liked this name for the guest's use of the host. But it only conveys have the responsibilities; it also needs to receive calls from the host.
 - HostLink/HostConnection/HostClient - I don't like these names because it's unclear whether they talk to the host, or they run on the host.
 - BridgeToHost - I don't hate this one as much as I should?
 - Host - This one wins the compactness award, and it also is nice because from the guest's perspective this _is_ the host. But the actual responsibilities aren't just talking to the host; it also needs to do some client-side things to do those communications well.

---


JakeWharton:
I still think a common noun should be shared between them (like "bridge"). They both serve the same purpose, but on opposite sides of the boundary. The noun should be the purpose and the "host" and "guest" are like qualifiers.

JakeWharton:
Maybe HostProtocolAdapter and GuestProtocolAdapter?

swankjesse:
Adapter works for me!

JakeWharton (approved):
Looks good otherwise

PR #2132, opened 2024-06-21
Extract an abstract class for TreehouseApp

swankjesse:
This should make it easier to write unit tests for code that uses a TreehouseApp but doesn't need to actually download code.

---


JakeWharton:
Is this working towards #322? Or something else?

JakeWharton (approved):

swankjesse:
Yep!

Issue #2104, opened 2024-06-12
Close TreehouseDispatchers in TreehouseApp.Factory.close()


swankjesse:
Otherwise these run forever!

swankjesse:
And the SingleThreadCoroutineDispatcher code is might be wrong because it might not complete already-enqueued tasks.

swankjesse:
We now close the right dispatchers at the right time.

PR #2134, opened 2024-06-24
Update IosTreehouseDispatchersTest

hunterestrada:
This confirms the current behavior works as expected and helps reduce the risk of a future regression: https://github.com/cashapp/redwood/issues/2104.

### DIFF
--- a/redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/IosTreehouseDispatchersTest.kt
+++ b/redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/IosTreehouseDispatchersTest.kt
@@ -15,10 +15,45 @@
  */
 package app.cash.redwood.treehouse
 
+import app.cash.turbine.withTurbineTimeout
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.time.DurationUnit.SECONDS
+import kotlin.time.toDuration
+import kotlinx.coroutines.Runnable
+import kotlinx.coroutines.test.runTest
+import platform.Foundation.NSThread
+
 class IosTreehouseDispatchersTest : AbstractTreehouseDispatchersTest() {
-  override val treehouseDispatchers: TreehouseDispatchers = IosTreehouseDispatchers()
+  private val ziplineDispatcher = SingleThreadCoroutineDispatcher()
+  override val treehouseDispatchers: TreehouseDispatchers = IosTreehouseDispatchers(ziplineDispatcher = ziplineDispatcher)
 
   /** We haven't set done the work to dispatch to the UI thread on iOS tests. */
   override val ignoreTestsThatExecuteOnUiThread: Boolean
     get() = true
+
+  @Test
+  fun closeFinishesZiplineThreadWithoutExecutingSubsequentRunnable() = runTest {
+    val loggedRunnableIds = mutableListOf<Char>()
+    val runnable = { id: Char -> Runnable { loggedRunnableIds.add(id) } }
+
+    awaitZiplineThread(isExecuting = true)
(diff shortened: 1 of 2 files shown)


redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/IosTreehouseDispatchersTest.kt
```diff
@@ -15,10 +15,29 @@
...
+
+  @Test
+  fun closeOnlyStopsNewZiplineRunnable() = runTest {
+    val channel = Channel<Unit>()
+    val runnable = Runnable { launch { channel.send(Unit) } }
+    treehouseDispatchers.zipline.dispatch(coroutineContext, runnable)
+    treehouseDispatchers.close()
+    treehouseDispatchers.zipline.dispatch(coroutineContext, runnable)
+
+    channel.consumeAsFlow().test { awaitItem() }
```
JakeWharton: Unfortunately this isn't sufficient to test the absence of the second invocation of the `Runnable`. Turbine can only help you wait and test for the presence of events.

In general there's no way to test for absence without reaching into implementation details. I think the best we can do here is to make `_zipline` and `SingleThreadCoroutineDispatcher` be `internal` and then reach into that to get at the `NSThread` and wait (with a timeout) until it exits.

With that we don't even need the channel or to schedule work, since we can just call close and ensure the thread goes away. Of course we're writing a bit of a change-detector test exposing the internals, but it's still fairly high level in that we only ensure we don't leak the thread and aren't too privy to the internal mechanisms.

hunterestrada: > Unfortunately this isn't sufficient to test the absence of the second invocation of the Runnable. Turbine can only help you wait and test for the presence of events.

Ah, I thought this was working without an exception for not all items consumed being thrown.

> I think the best we can do here is to make _zipline and SingleThreadCoroutineDispatcher be internal and then reach into that to get at the NSThread and wait (with a timeout) until it exits.

Sounds good.

> Of course we're writing a bit of a change-detector test exposing the internals, but it's still fairly high level in that we only ensure we don't leak the thread and aren't too privy to the internal mechanisms.

Yeah, I was trying to avoid diving into the internals too much with the first approach, but this makes sense.

redwood-treehouse-host/src/iosTest/kotlin/app/cash/redwood/treehouse/IosTreehouseDispatchersTest.kt
```diff
@@ -15,10 +15,45 @@
...
   /** We haven't set done the work to dispatch to the UI thread on iOS tests. */
   override val ignoreTestsThatExecuteOnUiThread: Boolean
     get() = true
+
+  @Test
+  fun closeFinishesZiplineThreadWithoutExecutingSubsequentRunnable() = runTest {
+    val loggedRunnableIds = mutableListOf<Char>()
+    val runnable = { id: Char -> Runnable { loggedRunnableIds.add(id) } }
+
+    awaitZiplineThread(isExecuting = true)
```
hunterestrada, line 41: Intended to fix
```
app.cash.redwood.treehouse.IosTreehouseDispatchersTest.closeFinishesZiplineThreadWithoutExecutingSubsequentRunnable[iosSimulatorArm64] FAILED
    kotlin.AssertionError: Expected value to be true.

JakeWharton:
There was an issue with how the timeout was being used which I fixed and then I ended up restructuring the internals of the main class a bit since I was already there. I think there's still a bug in the main class, but i'll address separately. Set to merge!

JakeWharton (approved):

swankjesse (commented):
Love it

PR #2135, opened 2024-06-24
Close dispatchers from app factory

hunterestrada:
Issue: https://github.com/cashapp/redwood/issues/2104
`LeaksTest`

### DIFF
--- a/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/RealTreehouseApp.kt
+++ b/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/RealTreehouseApp.kt
@@ -19,7 +19,6 @@ import app.cash.zipline.EventListener as ZiplineEventListener
 import app.cash.zipline.Zipline
 import app.cash.zipline.loader.LoadResult
 import app.cash.zipline.loader.ManifestVerifier
-import app.cash.zipline.loader.ZiplineCache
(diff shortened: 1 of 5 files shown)


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/RealTreehouseApp.kt
```diff
@@ -192,7 +191,11 @@ internal class RealTreehouseApp<A : AppService> private constructor(
     ): TreehouseApp<A> = RealTreehouseApp(this, appScope, spec, eventListenerFactory)
 
     override fun close() {
-      cache.close()
+      if (cache.isInitialized()) {
+        cache.value.close()
+      }
```
hunterestrada, line 196: Otherwise, we'd hit `error("unexpected call")` @ https://github.com/cashapp/redwood/blob/fc483a2de5530bf5992c62d50170bf150886c08c/redwood-treehouse-host/src/appsJvmTest/kotlin/app/cash/redwood/treehouse/TreehouseTester.kt#L72-L74.

A fake `TreehousePlatform` isn't so convenient to set up yet because there's not already a fake `ZiplineCache`.

JakeWharton (approved):
Just pushed some small nits you needn't be bothered with

swankjesse (commented):
👍🏻

PR #2143, opened 2024-06-26
Strictly validate parameters in SparseList

swankjesse:


dnagler (approved):
Nice! Potentially creates more "footguns", but better to have clearer exceptions thrown.

PR #2151, opened 2024-06-27
Fix Yoga styles not updating

### LINKED ISSUE #2026: Column/Row margins do not change when updated
```kotlin
Column(margin = Margin(start = if (thisOrThat) 16.dp else 0.dp)) { }
```

In this example, if `thisOrThat` is some state that mutates, the margins of this column don't get updated. This is especially noticeable in lazy columns with view reuse enabled.

dellisd:
Consider the value of a `CompactValue` in equality comparison, not just its unit

Fixes #2026 

---


colinrtwhite (approved):
Nice find!

PR #2152, opened 2024-06-27
In UIViewLazyList, make the UITableView's background color clear

dnagler:
This is a partial revert of PR #2146; we still want the individual cells to have non-clear backgrounds for performance reasons, but the underlying table should be clear

---


dellisd (approved):
Is there a way this could be set up so that a background color modifier would be able to override this?

JakeWharton:
@dellisd On the LazyColumn it should do just that

PR #2160, opened 2024-07-03
KotlinPoet helper: `buildFileSpec`

JakeWharton:
I have like 10 commits that build up KotlinPoet helpers to reduce the noise around code gen from a few months ago. They're going stale, so I'm going to rebase one at a time over the next couple weeks and PR separately. The juicy helpers won't show up until later.

---


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/composeGeneration.kt
```diff
@@ -232,14 +230,13 @@ fun Modifier.something(...): Modifier {
 internal fun generateUnscopedModifiers(schema: Schema): FileSpec? {
   if (schema.unscopedModifiers.isEmpty()) return null
 
-  return FileSpec.builder(schema.composePackage(), "unscoped")
-    .addAnnotation(suppressDeprecations)
-    .apply {
```
JakeWharton, line 237: Losing these `.apply` calls are really the only win here so far

colinrtwhite (approved):

PR #2183, opened 2024-07-16
Use a different thread for each TreehouseApp

swankjesse:
This should prevent different apps from competing with each other for CPU resources, especially during the first few seconds of the host app's launch, when there's a bunch of CPU-bound work to do.

---

### DIFF
--- a/redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/AndroidTreehouseDispatchers.kt
+++ b/redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/AndroidTreehouseDispatchers.kt
@@ -17,20 +17,22 @@ package app.cash.redwood.treehouse
 
 import android.os.Looper
 import java.util.concurrent.Executors
+import kotlinx.coroutines.CloseableCoroutineDispatcher
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.asCoroutineDispatcher
 
 /**
  * Implements [TreehouseDispatchers] suitable for production Android use. This creates a background
  * thread for all Zipline work.
  */
-internal class AndroidTreehouseDispatchers : TreehouseDispatchers {
+internal class AndroidTreehouseDispatchers(applicationName: String) : TreehouseDispatchers {
   private var ziplineThread: Thread? = null
 
   /** The single thread that runs all JavaScript. We only have one QuickJS instance at a time. */
   private val executorService = Executors.newSingleThreadExecutor { runnable ->
-    Thread(null, runnable, "Treehouse", ZIPLINE_THREAD_STACK_SIZE.toLong())
+    Thread(null, runnable, "Treehouse $applicationName", ZIPLINE_THREAD_STACK_SIZE.toLong())
       .also { ziplineThread = it }
   }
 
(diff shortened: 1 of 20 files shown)


redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/RealTreehouseApp.kt
```diff
@@ -122,6 +123,8 @@ internal class RealTreehouseApp<A : AppService> private constructor(
     if (!spec.loadCodeFromNetworkOnly) {
       loader = loader.withCache(
         cache = factory.cache.value,
+        // TODO(jwilson): use this once we update Zipline.
+        // cacheDispatcher = factory.ziplineLoaderDispatcher,
```
swankjesse, line 127: * [ ] Do not merge until this fix lands!

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/AndroidTreehouseDispatchers.kt
```diff
@@ -49,3 +51,11 @@ internal class AndroidTreehouseDispatchers : TreehouseDispatchers {
     executorService.shutdown()
   }
 }
+
+@OptIn(ExperimentalCoroutinesApi::class) // CloseableCoroutineDispatcher is experimental.
+internal fun ziplineLoaderDispatcher(): CloseableCoroutineDispatcher {
+  val executorService = Executors.newSingleThreadExecutor { runnable ->
```
JakeWharton, line 57: Does this need a dedicated thread? Shouldn't `Disaptchers.IO` be sufficient? Same for iOS.

JakeWharton (approved):

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/AndroidTreehouseDispatchers.kt
swankjesse: Unfortunately Dispatchers.IO isn’t thread-confined, and I need that for SQLite.

swankjesse: I might be able to borrow a thread from that pool though. Somehow.

JakeWharton: You can limit the parallelism to 1 if concurrency/parallelism is the problem, but that doesn't guarantee the same thread is used.

swankjesse: Yeah, that could work. Lemme try.

swankjesse: (In follow-up, as I do.)

dnagler:
Belated LGTM, thanks for taking this on!

redwood-treehouse-host/src/androidMain/kotlin/app/cash/redwood/treehouse/AndroidTreehouseDispatchers.kt
swankjesse: https://github.com/cashapp/redwood/pull/2188/files

hunterestrada:
Yeah, this is great!

Issue #2037, opened 2024-05-16
Add rudimentary leak detection to all platforms


JakeWharton:
We know when things should go away. We know when a widget is removed from the hierarchy. We know when a treehouse app is replaced. In those cases, we can do the world's most basic leak detector of holding those references weakly after removal, and periodically check + force GC to see if they get collected.

Refs #2029. Refs #2033.

JakeWharton:
This is tracked by the expanded scope of #575

PR #2191, opened 2024-07-17
Add leak detection API and module

JakeWharton:
This can be used to watch a reference when it is known to no longer be used, and to check whether or not it was collected by the Kotlin garbage collector or not.

Refs #2037 

---


colinrtwhite (approved):
Should this live in the Zipline repo instead of Redwood? Seems like it doesn't rely on any Redwood-specific classes.

swankjesse (approved):

JakeWharton:
It doesn't rely on Zipline either. I'm only testing with Zipline to confirm its behavior inside QuickJS (which is that it both doesn't have a mechanism to trigger GC and that it doesn't support `WeakRef`).

It can live in its own repository someday if we want.

PR #2192, opened 2024-07-18
Add special modifier testing to FIR

JakeWharton:
Refs #19

---


redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/schemaParser.kt
```diff
@@ -482,9 +482,9 @@ private fun parseModifier(
   )
 }
 
-/** Returns true if [memberType] is a known reserved tag name. */
-private fun isReservedModifier(tag: Int, memberType: KClass<*>): Boolean {
```
JakeWharton, line 486: We already use "reserved" to denote tags which are reserved by the user, so changed the name to "special" just to be different.

colinrtwhite (approved):

PR #2196, opened 2024-07-19
Add schema extension property to use FIR-based parsing

JakeWharton:
Mostly so I can test internally before removing the old impl (and once the other PRs land).

Refs #19

---


redwood-gradle-plugin/src/main/kotlin/app/cash/redwood/gradle/RedwoodGeneratorExtension.kt
```diff
@@ -18,10 +18,13 @@ package app.cash.redwood.gradle
 import org.gradle.api.artifacts.dsl.DependencyHandler
 import org.gradle.api.provider.Property
 
-public abstract class RedwoodGeneratorExtension : RedwoodSchemaExtension() {
+public abstract class RedwoodGeneratorExtension {
```
JakeWharton, line 21: Doesn't make sense for this to get existing `apiTracking` or the new `useFir` properties.

colinrtwhite (approved):

PR #2200, opened 2024-07-23
WIP: UiConfiguration.windowInsets

swankjesse:
This introduces a new mutable property, windowInsets, that is propagated to the composition where it can be consumed.

We put the burden on the user of RedwoodView to set this value to something non-zero if they have insets to be consumed. I've added an accelerator on Android for this,

  RedwoodLayout.consumeWindowInsets(mask: Int)

I'll look into a similar API for iOS and HTML to do likewise.

---


swankjesse:
Stuff to do still:

 * [ ] Screenshot tests
 * [ ] iOS helper function to set insets from the enclosing Window
 * [ ] HTML helper function to set insets from the enclosing Window

This doesn't do anything with the existing `safeAreaInsets` value. I am tempted to delete that in follow-up. I don’t want to re-purpose that value as it has a non-zero value on existing host code but that value isn’t necessarily what we want.

PR #2203, opened 2024-07-24
Fix screenshots not being LFS.

colinrtwhite:


JakeWharton (approved):

JakeWharton:
Didn't we have a CI check for this?

colinrtwhite:
Yep I thought so as well. I think that might only guard against adding new errors like this. These screenshots were introduced before the CI check was added.

Issue #2145, opened 2024-06-26
Treehouse does JSON decoding on the main thread


swankjesse:
Host side bindings look like this:

```kotlin
  override fun apply(change: PropertyChange, eventSink: EventSink) {
    val widget = _widget ?: error("detached")
    when (change.tag.value) {
      1 -> widget.avatars(json.decodeFromJsonElement(serializer_0, change.value))
      4 -> widget.size(json.decodeFromJsonElement(serializer_1, change.value))
      2 -> widget.badge(json.decodeFromJsonElement(serializer_2, change.value))
      3 -> {
        val onClick: (() -> Unit)? = if (change.value.jsonPrimitive.boolean) {
          OnClick(json, change.id, eventSink)::invoke
        } else {
          null
        }
        widget.onClick(onClick)
      }
      5 -> widget.isMerchant(json.decodeFromJsonElement(serializer_3, change.value))
      else -> mismatchHandler.onUnknownProperty(WidgetTag(15), change.tag)
    }
  }
```

We should decode the `change.value` of the `PropertyChange` from JSON to a host-side model on the Zipline thread instead.

---

* [x] Event to JSON
    * #2218
* [x] JSON to property

JakeWharton:
How do you put something on the zipline thread?

We can trivially expose the list of property IDs like we do children IDs, and then offer a function to look up the `KSerializer` for a given property ID. We can look up the protocol node given the widget ID, but to get Zipline to use that before dispatch to the UI thread we need injection into the `KSerializer` for `Change`.

JakeWharton:
Oh I'm dumb. Treehouse is what moves to the UI thread. Okay this should be doable.

JakeWharton:
Need to wait a week or two for the guest-side changes to roll out internally before we can land this.

JakeWharton:
This is unblocked!

PR #2218, opened 2024-07-31
Defer serializing event args to JSON

JakeWharton:
Until we're on the Zipline thread instead of the UI thread.

Half of #2145.

---


colinrtwhite (approved):

swankjesse (commented):
COOOL

PR #2219, opened 2024-07-31
DirectLoadingStrategy is just for testing

swankjesse:
---

### DIFF
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ScrollOptimizedLoadingStrategy.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ScrollOptimizedLoadingStrategy.kt
@@ -42,7 +42,6 @@ public class ScrollOptimizedLoadingStrategy(
   private val scrollInProgressPreloadItemCount: Int = SCROLL_IN_PROGRESS_PRELOAD_ITEM_COUNT,
   private val primaryPreloadItemCount: Int = PRIMARY_PRELOAD_ITEM_COUNT,
   private val secondaryPreloadItemCount: Int = SECONDARY_PRELOAD_ITEM_COUNT,
-  private val preloadItems: Boolean = true,
 ) : LoadingStrategy {
   /**
    * Update this to trigger a programmatic scroll. This may be updated multiple times, including
@@ -92,11 +91,6 @@ public class ScrollOptimizedLoadingStrategy(
 
     // Expand the range depending on scroll direction.
     when {
-      // Ignore preloads.
-      !preloadItems -> {
-        // No-op
-      }
-
       isScrollingDown -> {
         begin -= scrollInProgressPreloadItemCount
         end += primaryPreloadItemCount
--- a/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ScrollOptimizedLoadingStrategy.kt
+++ b/redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ScrollOptimizedLoadingStrategy.kt
@@ -53,9 +53,9 @@ public class ScrollOptimizedLoadingStrategy(
     private set
 
   /** Bounds of what the user is looking at. Everything else is placeholders! */
-  public override var firstIndex: Int by mutableIntStateOf(0)
+  public override var firstVisibleIndex: Int by mutableIntStateOf(0)
     private set
-  public var lastIndex: Int by mutableIntStateOf(0)
+  public override var lastVisibleIndex: Int by mutableIntStateOf(0)
     private set
 
   private var firstIndexFromPrevious1: Int by mutableIntStateOf(DEFAULT_SCROLL_INDEX)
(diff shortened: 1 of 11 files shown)


redwood-lazylayout-compose/src/commonTest/kotlin/app/cash/redwood/lazylayout/compose/LazyListTest.kt
```diff
@@ -149,21 +147,12 @@ class LazyListTest {
...
       lazyList.onViewportChanged(6, 10)
       awaitSnapshot()
       assertThat(index5ComposeCount).isEqualTo(1)
 
-      // The item at index 5 remains composed because (6, 10) is contiguous with (4, 8).
-      lazyList.onViewportChanged(4, 8)
-      awaitSnapshot()
-      assertThat(index5ComposeCount).isEqualTo(1)
-
-      // Scrolling to a non-contiguous range causes a recomposition.
```
swankjesse, line 162: This was awkward because it needed to pick bounds carefully.

redwood-lazylayout-testing/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/DirectLoadingStrategy.kt
```diff
@@ -0,0 +1,44 @@
...
+package app.cash.redwood.lazylayout.compose
+
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+
+/**
+ * A loading strategy that's appropriate for tests because it's simple and predictable.
+ *
+ * This is not suitable for production use it must show a placeholder before an item is loaded.
```
JakeWharton: Is this two sentences, or missing a comma or something?

redwood-lazylayout-testing/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/DirectLoadingStrategy.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class DirectLoadingStrategy : LoadingStrategy {
+  private var loadRange: IntRange by mutableStateOf(0..0)
```
JakeWharton:
```suggestion
  private var loadRange: IntRange = 0..0
```
The only read is `firstIndex` which is only used when saving. Don't think you need a full-on state object tracking this value.

redwood-lazylayout-compose/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/ScrollOptimizedLoadingStrategy.kt
```diff
@@ -92,11 +91,6 @@ public class ScrollOptimizedLoadingStrategy(
 
     // Expand the range depending on scroll direction.
     when {
-      // Ignore preloads.
-      !preloadItems -> {
```
dnagler, line 96: Seems fine, this was only ever used in one unit test anyway

JakeWharton (approved):

redwood-lazylayout-testing/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/DirectLoadingStrategy.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+
+/**
+ * A loading strategy that's appropriate for tests because it's simple and predictable.
+ *
+ * This is not suitable for production use it must show a placeholder before an item is loaded.
+ */
+public class DirectLoadingStrategy : LoadingStrategy {
```
dnagler: naming nit: I think the name should make it abundantly clear that this should never be used in production. Something like `Fake`/`Test`/`Mock` instead of `Direct` would help with this.

dnagler (approved):

redwood-lazylayout-testing/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/DirectLoadingStrategy.kt
swankjesse: Fixed!

```
 * This is not suitable for production use, because it must show a placeholder before an item is
 * loaded.
```

redwood-lazylayout-testing/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/DirectLoadingStrategy.kt
swankjesse: Don’t I need this to trigger a recomposition after the user scrolls?

redwood-lazylayout-testing/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/DirectLoadingStrategy.kt
swankjesse: Good idea. Changed to TestLoadingStrategy.

dnagler: Thanks!

redwood-lazylayout-testing/src/commonMain/kotlin/app/cash/redwood/lazylayout/compose/DirectLoadingStrategy.kt
JakeWharton: I don't think the composition reads this. Maybe I missed it.

swankjesse: It calls loadRange(), which reads variables here. I assume using State variables is the easiest way to express the data dependency.

PR #2239, opened 2024-08-07
Revert "Defer serializing event args to JSON (#2218)"

dnagler:
This reverts commit 4010e1247fef28225fb516463c98f2e3bc9f7074.

I believe this commit is the root cause of a very serious regression that results in unresponsive UI (firing `unknownEventNode` warnings in response to most tap events).

---

- [x] Confirm that this fixes the regression that we're seeing.


swankjesse (approved):

dnagler:
Confirmed--with this revert in place, the issue no longer reproduces.

PR #2242, opened 2024-08-08
Change UIViewLazyList so cells can be removed and re-added without reuse

swankjesse:
We observed in some test suites that table cells would observe a sequence of calls like this:

LazyListContainerCell.createView()
LazyListContainerCell.willMoveToSuperview(non-null) LazyListContainerCell.willMoveToSuperview(null)
LazyListContainerCell.willMoveToSuperview(non-null)

Note that we didn't receive a call to prepareForReuse() between the code being removed from its parent and added again. The code wasn't prepared for this sequence of calls.

---

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -301,14 +301,13 @@ internal class LazyListContainerCell(
     }
 
     // Confirm the cell is bound when it's about to be displayed.
-    if (superview == null && newSuperview != null) {
-      require(binding!!.isBound) { "about to display a cell that isn't bound!" }
+    if (superview == null && newSuperview != null && !binding!!.isBound) {
+      binding!!.bind(this)
     }
 
     // Unbind the cell when its view is detached from the table.
     if (superview != null && newSuperview == null) {
       binding?.unbind()
-      binding = null
     }
   }
 
--- a/redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
+++ b/redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
@@ -314,22 +314,24 @@ public abstract class LazyListUpdateProcessor<V : Any, W : Any> {
   }
 
   private fun placeholderToLoaded(
-    placeholder: Binding<V, W>?,
+    binding: Binding<V, W>?,
     loadedContent: Widget<W>,
   ): Binding<V, W> {
     // No binding for this index. Create one.
-    if (placeholder == null) {
+    if (binding == null) {
       return Binding(this).apply {
         setContentAndModifier(loadedContent)
       }
     }
 
     // We have a binding. Give it loaded content.
-    require(placeholder.isPlaceholder)
-    recyclePlaceholder(placeholder.content)
-    placeholder.isPlaceholder = false
-    placeholder.setContentAndModifier(loadedContent)
-    return placeholder
+    require(binding.isPlaceholder)
+    if (binding.isBound) {
+      recyclePlaceholder(binding.content!!)
+    }
+    binding.isPlaceholder = false
+    binding.setContentAndModifier(loadedContent)
+    return binding
   }
 
   private fun loadedToPlaceholder(loaded: Binding<V, W>): Binding<V, W>? {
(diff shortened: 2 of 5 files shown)


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -301,14 +301,13 @@ internal class LazyListContainerCell(
...
-    if (superview == null && newSuperview != null) {
-      require(binding!!.isBound) { "about to display a cell that isn't bound!" }
+    if (superview == null && newSuperview != null && !binding!!.isBound) {
+      binding!!.bind(this)
     }
 
     // Unbind the cell when its view is detached from the table.
     if (superview != null && newSuperview == null) {
       binding?.unbind()
-      binding = null
```
swankjesse, line 311: So we can bind again later

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
```diff
@@ -314,22 +314,24 @@ public abstract class LazyListUpdateProcessor<V : Any, W : Any> {
...
 
     // We have a binding. Give it loaded content.
-    require(placeholder.isPlaceholder)
-    recyclePlaceholder(placeholder.content)
-    placeholder.isPlaceholder = false
-    placeholder.setContentAndModifier(loadedContent)
-    return placeholder
+    require(binding.isPlaceholder)
+    if (binding.isBound) {
+      recyclePlaceholder(binding.content!!)
```
hunterestrada, line 330: If a bound binding must have content, wdyt about restructuring the code to guarantee that requirement before this point? It looks like we could avoid some not-null assertions and have an access pattern here and [below](https://github.com/cashapp/redwood/pull/2242/files#diff-4eb0cd03030153463a7b458145149e2532ed6edc61908679fa5662ef4a27b0a2R495) like...
```
when (val state = binding.state) {
  is Bound -> { recyclePlaceholder(state.content) }
  is Unbound -> {}
}
```

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
```diff
@@ -458,9 +465,13 @@ public abstract class LazyListUpdateProcessor<V : Any, W : Any> {
...
       get() = view != null
 
-    internal fun bind(view: V) {
+    public fun bind(view: V) {
       require(this.view == null) { "already bound" }
 
+      if (isPlaceholder && this.content == null) {
+        this.content = processor.takePlaceholder()
+      }
+
```
hunterestrada, line 474: Without the structural change to a binding's state, should we add `require(this.content != null)` for the non-placeholder case?

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: This feels a little dangerous to me--what are the odds that this causes a memory management issue, because it's being retained?

hunterestrada (approved):
LGTM

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: Good instinct. I double checked this and `unbind()` breaks the cycle.

But I don’t like that even with Redwood’s LeakTest, we don’t have dynamic coverage of leaks like this. I’ve opened [#2245](https://github.com/cashapp/redwood/issues/2245) to test it directly.

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
swankjesse: I think this is a great idea. Let’s move this logic out of the code and into the type system! I’d like to do it in a follow-up though.

redwood-lazylayout-widget/src/commonMain/kotlin/app/cash/redwood/lazylayout/widget/LazyListUpdateProcessor.kt
swankjesse: Great idea. Lemme do that in the above-mentioned follow-up.

PR #2250, opened 2024-08-16
Breaking content: UIView retain cycle in UIViewLazyList's LazyListContainerCell

dnagler:
The content view is a Swift object, so it needs to be manually nil'd out by its Kotlin retainer

---

- [x] Do not merge until this is fully verified within Cash App

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -119,7 +119,7 @@ internal open class UIViewLazyList :
     }
 
     override fun setContent(view: LazyListContainerCell, content: UIView?, modifier: Modifier) {
-      view.content = content
+      view.setContent(content)
     }
   }
 
@@ -269,16 +269,6 @@ internal class LazyListContainerCell(
   reuseIdentifier: String?,
 ) : UITableViewCell(style, reuseIdentifier) {
   internal var binding: Binding<LazyListContainerCell, UIView>? = null
-  internal var content: UIView? = null
-    set(value) {
-      field = value
-
-      removeAllSubviews()
-      if (value != null) {
-        contentView.addSubview(value)
-      }
-      setNeedsLayout()
-    }
 
   override fun initWithStyle(
     style: UITableViewCellStyle,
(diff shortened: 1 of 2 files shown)


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -315,6 +315,7 @@ internal class LazyListContainerCell(
     super.prepareForReuse()
     binding?.unbind()
     binding = null
+    content = null
```
swankjesse: Unfortunately `prepareForReuse()` isn’t called reliably by UITableView. I would prefer for it to be called when it puts a cell in its pool, but unfortunately it don’t think it calls it until it removes a cell from its pool.

dnagler: Pushed an alternative approach

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -320,13 +310,26 @@ internal class LazyListContainerCell(
   override fun layoutSubviews() {
     super.layoutSubviews()
 
-    val content = this.content ?: return
+    if (contentView.subviews.isEmpty()) return
+
+    val content = contentView.subviews.first() as UIView ?: return
```
swankjesse, line 315: Is `?: return` necessary?

swankjesse (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: Surprisingly, yes--without it, if there is no content view to be found, this throws a `List is empty.` error that bricks the whole Treehouse app.

JakeWharton: You would need to use `firstOrNull()` to receive a nullable value on which the `?:` would then be able to apply. With `first()` it will throw on empty and never execute that return.

dnagler: Fixed in #2252

PR #2252, opened 2024-08-19
Using firstOrNull instead of first in UIViewLazyList

dnagler:
Follow-up to #2250

### DIFF
--- a/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
+++ b/redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
@@ -310,17 +310,13 @@ internal class LazyListContainerCell(
   override fun layoutSubviews() {
     super.layoutSubviews()
 
-    if (contentView.subviews.isEmpty()) return
-
-    val content = contentView.subviews.first() as UIView ?: return
+    val content = contentView.subviews.firstOrNull() as UIView? ?: return
     content.setFrame(bounds)
     contentView.setFrame(bounds)
   }
 
   override fun sizeThatFits(size: CValue<CGSize>): CValue<CGSize> {
-    if (contentView.subviews.isEmpty()) return super.sizeThatFits(size)
-
-    val content = contentView.subviews.first() as UIView ?: return super.sizeThatFits(size)
+    val content = contentView.subviews.firstOrNull() as UIView? ?: return super.sizeThatFits(size)
     return content.sizeThatFits(size)
   }
 


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -312,15 +312,15 @@ internal class LazyListContainerCell(
 
     if (contentView.subviews.isEmpty()) return
 
-    val content = contentView.subviews.first() as UIView ?: return
+    val content = contentView.subviews.firstOrNull() as UIView ?: return
```
JakeWharton:
```suggestion
    val content = contentView.subviews.firstOrNull() as UIView? ?: return
```
So without this the cast will throw on null. We need to cast to a nullable version of the type we expect so that null can pass through.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -312,15 +312,15 @@ internal class LazyListContainerCell(
...
     content.setFrame(bounds)
     contentView.setFrame(bounds)
   }
 
   override fun sizeThatFits(size: CValue<CGSize>): CValue<CGSize> {
     if (contentView.subviews.isEmpty()) return super.sizeThatFits(size)
 
-    val content = contentView.subviews.first() as UIView ?: return super.sizeThatFits(size)
+    val content = contentView.subviews.firstOrNull() as UIView ?: return super.sizeThatFits(size)
```
JakeWharton:
```suggestion
    val content = contentView.subviews.firstOrNull() as UIView? ?: return super.sizeThatFits(size)
```

JakeWharton (approved):

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: Ooooh . . . I think we can skip the `isEmpty()` check above if we use `firstOrNull()` here.

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
swankjesse: Same, the isEmpty() check is unnecessary if we use `firstOrNull()`

redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
dnagler: Much cleaner, thanks!

swankjesse (approved):

swankjesse:
# :shipit:

PR #2253, opened 2024-08-21
Optimize Modifier serialization in guest code

swankjesse:
I learned that encoding a type as a JsonElement and then encoding that as JSON is significantly less efficient than going directly from the type to JSON. This is due to some inefficient code in the internals of kotlinx.serialization.

Skipping this intermediate step is useful on its own, but it also turns out to be significantly more efficient in practice.

---


redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolWidgetSystemFactory.kt
```diff
@@ -15,12 +15,23 @@
...
 
 public interface ProtocolWidgetSystemFactory {
   /** Create a new [WidgetSystem] connected to a host via [guestAdapter]. */
   public fun create(
     guestAdapter: GuestProtocolAdapter,
     mismatchHandler: ProtocolMismatchHandler = ProtocolMismatchHandler.Throwing,
   ): WidgetSystem<Unit>
+
+  @RedwoodCodegenApi
+  public fun modifierTag(element: Modifier.Element): ModifierTag
```
swankjesse: Tempted to do one function that returns a `Pair<ModifierTag, KSerializer<T>?>` instead of two functions. The only hazard of doing one function is I’d like to avoid allocating a `Pair` if possible.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolWidgetSystemFactory.kt
```diff
@@ -15,12 +15,23 @@
...
+
+  /** Returns null if the modifier is stateless and should serialize as null. */
+  @RedwoodCodegenApi
+  public fun <T : Modifier.Element> modifierSerializer(element: T): KSerializer<T>?
```
swankjesse: On my first attempt this returned a non-null `KSerializer<T>`, but I ran into a [this bug](https://github.com/Kotlin/kotlinx.serialization/pull/2712) in Kotlinx.serialization. I’d like to contribute a fix for that bug and then later I can change this to return non-null always.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolWidgetSystemFactory.kt
```diff
@@ -15,12 +15,23 @@
  */
 package app.cash.redwood.protocol.guest
 
+import app.cash.redwood.Modifier
+import app.cash.redwood.RedwoodCodegenApi
+import app.cash.redwood.protocol.ModifierTag
 import app.cash.redwood.widget.WidgetSystem
+import kotlinx.serialization.KSerializer
 
 public interface ProtocolWidgetSystemFactory {
```
swankjesse, line 24: With these two new APIs I don’t think the `Factory` name fit as well as it did previously.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolWidgetSystemFactory.kt
JakeWharton: We generate the `KSerializer` for every modifier. We can do two things:

 1. Switch to subtyping `SerializationStrategy` in the generated type rather than `KSerializer`. We generate custom serializers because we only want the serialization code, so we should only be implementing the serialization interface (no clue why I didn't do this initially).
 2. Define a `ProtocolSerializationStrategy` interface which adds `val tag: ModifierTag` to `SerializationStrategy` and switch our codegen to use that. This eliminates the need to do a double, type-based lookup.

redwood-treehouse-guest/src/jsMain/kotlin/app/cash/redwood/treehouse/ProtocolBridgeJs.kt
```diff
@@ -138,12 +138,20 @@ internal class FastGuestProtocolAdapter(
...
+    val elements = js("[]")
+
+    value.forEach { element ->
+      val tag = widgetSystemFactory.modifierTag(element)
+      val serializer = widgetSystemFactory.modifierSerializer(element)
+      val value = when {
+        serializer == null -> null
+        else -> json.encodeToDynamic(serializer, element)
+      }
+      elements.push(js("""[tag,value]"""))
```
JakeWharton: The protocol when used normally does not encode `null` values here. There's no real harm in it, though, except for the 5 extra bytes on the wire.

JakeWharton: We might need to move our protocol serialization tests upward to ensure these custom forms mirror it in behavior.

JakeWharton (approved):

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolWidgetSystemFactory.kt
swankjesse: I like both of these ideas. I’ll add a commit to this PR to do that!

redwood-treehouse-guest/src/jsMain/kotlin/app/cash/redwood/treehouse/ProtocolBridgeJs.kt
swankjesse: Ooooh good catch, I missed that. I think it’s a nice improvement to test at the JSON-encoded layer for this, though I don’t think Zipline’s `asDynamicFunction()` makes this particularly easy. It’s trying to hide the JSON layer and the test wants to exercise it!

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolWidgetSystemFactory.kt
swankjesse: Done. Unfortunately I couldn’t do it exactly as suggested 'cause we don’t have a serializer for stateless modifiers. So instead it’s a `Pair<ModifierTag, SerializationStrategy>`. Seems to work okay and read okay.

redwood-treehouse-guest/src/jsMain/kotlin/app/cash/redwood/treehouse/ProtocolBridgeJs.kt
swankjesse: Fixed to not encode null, but I haven’t yet fixed the test case to validate the JSON matches as specified. I’ll need to contemplate how to do that in a way that feels right at this layer.

PR #2266, opened 2024-08-27
Watch Zipline instances for leak after stop

### LINKED ISSUE #575: Add automated leak detection for common scenarios
There are common scenarios that occur in Redwood and Treehouse which we should use to ensure branches of the graph are correctly GC'd.

Examples:
- Remove a composable from `RedwoodView` composition and the associated view should be GC'd
- Cancel Redwood composition and the entire composition and all views should go
- Remove a composable from within Treehouse and the associated view should be GC'd
- New code is loaded using Treehouse and all old views should go
- Cancel Treehouse and the entire JS VM and `TreehouseView` subtype should go

We need to be able to do this on every platform we support.

On Android and JVM we can use LeakCanary. On JS and native we can just do what leak canary is doing: create a weak reference and spin for 5 seconds to see if the reference disappears, optionally trying to force a GC as much as possible (if possible).

JakeWharton:
Closes #575

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 New:
 - Source-based schema parser is now the default. The `useFir` Gradle property has been removed.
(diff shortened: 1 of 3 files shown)


dnagler (approved):
Nice!

Issue #2225, opened 2024-08-01
Parse default expressions instead of `@Default` annotation


JakeWharton:
Now that schema is parsed with FIR.

Probably want to have a version where one of the two can be used. Then we can migrate internal schemas, and finally come back here to fully remove the annotation.

 * [x] Parse default expression, allow only one of annotation OR default expression
    * #2279
 * [ ] Once the above ships in a version, remove default annotation completely
    * #2368

Follow-up from #19

PR #2279, opened 2024-09-03
Parse language default expressions

JakeWharton:
Deprecate old annotation-based mechanism.

First half of #2225.

---


redwood-tooling-schema/src/main/kotlin/app/cash/redwood/tooling/schema/schemaParserFir.kt
```diff
@@ -770,23 +771,31 @@ private data class ChildrenAnnotation(
...
+        ?: throw AssertionError(annotation.source?.text)
+    }
+    ?.value as String?
 
-  val expression = annotation.argumentMapping
-    .mapping[Name.identifier("expression")] as? FirLiteralExpression
-    ?: throw AssertionError(annotation.source?.text)
+  val expression = parameter.defaultValueSource?.let { defaultValue ->
+    defaultValue.treeStructure
+      .toString(defaultValue.lighterASTNode)
```
swankjesse, line 789: This is super awesome

swankjesse (approved):

swankjesse:
Will need to confirm we’re consistent when we upgrade this.

JakeWharton:
> Will need to confirm we’re consistent when we upgrade this.

Indeed, that's why I fail if both are present. When Eric and I first talked about this migration we said that we'd probably allow them to match. But I think just enforcing one or the other is best. An entire schema can be migrated in like 10 minutes, and you can diff the parsed JSON to ensure things remain semantically the same (that's how I did it for our layout schema).

colinrtwhite (approved):

PR #2292, opened 2024-09-10
Fix some git lfs files that should have been pointers

swankjesse:


swankjesse:
I manually confirmed that Android Studio says ‘Contents are Identical’ for all of these.

JakeWharton (approved):

PR #2294, opened 2024-09-10
Size the frame properly for UIViewBoxTest

swankjesse:
This was always rendering at size=0x0 and so none of the expected snapshots were correct.


redwood-layout-uiview/src/commonTest/kotlin/app/cash/redwood/layout/uiview/UIViewBoxTest.kt
```diff
@@ -41,18 +41,19 @@ class UIViewBoxTest(
     return UIViewText()
   }
 
-  override fun verifySnapshot(value: UIView, name: String?) {
+  override fun verifySnapshot(widget: UIView, name: String?) {
     val screenSize = CGRectMake(0.0, 0.0, 390.0, 844.0) // iPhone 14.
+    widget.setFrame(screenSize)
```
swankjesse, line 46: This is what we were missing.

JakeWharton (approved):

PR #2289, opened 2024-09-09
New ResizableWidget is a formal way to request layouts

swankjesse:
This is only on our UIView APIs because that's the only view system that needs an out-of-band mechanism to signal such changes.

This also introduces new optional parameters to UIViewRedwoodLayoutWidgetFactory and ViewRedwoodLayoutWidgetFactory to opt-in to experimental incremental layouts. This mechanism is still a work-in-progress so most users shouldn't opt in yet.

Working towards: https://github.com/cashapp/redwood/issues/2291

---

- [x] `CHANGELOG.md`'s "Unreleased" section has been updated, if applicable.

PR #2298, opened 2024-09-11
Fix testRecursiveLayoutIsIncremental in UIViewLazyList

swankjesse:
The test was incorrectly nesting a UITableView inside a UITableView instead of nesting a Row inside a UITableView.

See: https://github.com/cashapp/redwood/pull/2289

### DIFF
--- a/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
+++ b/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
@@ -45,8 +45,10 @@ abstract class AbstractFlexContainerTest<T : Any> {
     backgroundColor: Int = argb(51, 0, 0, 255),
   ): TestFlexContainer<T>
 
+  /** Returns a non-lazy flex container row, even if the test is for a LazyList. */
   abstract fun row(): Row<T>
 
+  /** Returns a non-lazy flex container column, even if the test is for a LazyList. */
   abstract fun column(): Column<T>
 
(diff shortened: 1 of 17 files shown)


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -691,33 +693,33 @@ abstract class AbstractFlexContainerTest<T : Any> {
     container.height(Constraint.Fill)
     container.crossAxisAlignment(CrossAxisAlignment.Start)
 
-    val rowA = flexContainer(FlexDirection.Row)
+    val rowA = row()
```
swankjesse, line 696: this is the real fix

redwood-lazylayout-uiview/src/commonTest/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyListAsFlexContainerTest.kt
```diff
@@ -48,16 +49,18 @@ class UIViewLazyListAsFlexContainerTest(
 
   override fun column() = UIViewRedwoodLayoutWidgetFactory().Column()
 
-  override fun text() = object : Text<UIView> {
+  override fun text(): Text<UIView> = object : Text<UIView>, ResizableWidget<UIView> {
```
swankjesse, line 52: ... and without `ResizableWidget`, it doesn’t react to the layout change

JakeWharton (approved):

PR #2310, opened 2024-09-20
Fix test app

AntonButov:
TestApp is broken. I think it's because of [this](https://github.com/Kotlin/kotlinx.serialization/issues/2713) issue. I fixed it by changing UInt to Int for the color in TestApp. Please review.


JakeWharton:
I'll look Monday. Thanks

JakeWharton:
I started to look at this, but I became concerned that merging this as-is will paper over a real problem. We have at least one workaround in place for the `UInt` problem (in the `GuestProtocolAdapter`), but obviously we're missing more. I don't want someone using `UInt` to get failures like this. We might need to outright reject it at the schema-level if it truly can't be worked around in the runtime.

AntonButov:
Certainly, this will be better for the users. The PR can be closed.

PR #2311, opened 2024-09-20
Use a more efficient representation for the children lookup

JakeWharton:
Use primitive-specialized collections from AndroidX collection.

With 24 keys and 37 values…

Before (`Map<WidgetTag, List<ChildrenTag>>`)

      64 bytes // LinkedHashMap
     144 bytes //  Node[] 16 + 4*(24/.75)
     960 bytes //  LinkedHashMap$Entry 40*24
     384 bytes // WidgetTag 16*24
     576 bytes // ArrayList 24*24
     532 bytes //   Object[] 16*24 + 4*37
     592 bytes // ChildrenTag 16*37
    ----------
    3302 bytes total

After (`IntObjectMap<IntArray>`)

     32 bytes // IntObjectMap
     40 bytes //   long[] 16 + 8*(24/8)
    112 bytes //   int[] 16 + 4*24
    112 bytes //   Object[] 16 + 4*24
    532 bytes // int[] 16*24 + 4*37
    ---------
    828 bytes total

…and obviously greater locality and all that helps as well.

---

### DIFF
--- a/redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolHostGeneration.kt
+++ b/redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolHostGeneration.kt
@@ -15,7 +15,6 @@
  */
 package app.cash.redwood.tooling.codegen
 
-import app.cash.redwood.tooling.codegen.Protocol.ChildrenTag
 import app.cash.redwood.tooling.codegen.Protocol.Id
 import app.cash.redwood.tooling.codegen.Protocol.WidgetTag
 import app.cash.redwood.tooling.schema.ProtocolSchema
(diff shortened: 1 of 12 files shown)


redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolHostGeneration.kt
```diff
@@ -56,15 +54,15 @@ public class ExampleProtocolFactory<W : Any>(
...
-        WidgetTag(3) to listOf(ChildrenTag(1)),
-        WidgetTag(1_000_001) to listOf(ChildrenTag(1), ChildrenTag(2)),
-        WidgetTag(1_000_002) to listOf(ChildrenTag(1)),
-      )
+  private val childrenTags: IntObjectMap<IntArray> = MutableIntObjectMap(4).apply {
+        put(1, intArrayOf(1))
+        put(3, intArrayOf(1))
+        put(1_000_001, intArrayOf(1, 2))
+        put(1_000_002, intArrayOf(1))
+      }
```
JakeWharton, line 62: Sent a CL for a builder API, but it'll be a while before we can use it: https://android-review.googlesource.com/c/platform/frameworks/support/+/3275076

swankjesse (approved):

PR #2314, opened 2024-09-23
Do not store the `WidgetTag` + base impl of `visitIds`

JakeWharton:
Storing the `WidgetTag` is a waste of space. We rarely use it, and even then it's only an `int` so we can just codegen returning it directly.

Do not generate implementations of `visitIds` for nodes with no children. Have the base class invoke the lambda with the current node ID by default. Only nodes for widgets with children will override and do codegen.

---


JakeWharton:
Working on #2145 and pulling some prerequisites out

swankjesse (approved):

PR #2315, opened 2024-09-23
Use more efficient collections on host protocol

JakeWharton:
These do not use virtual calls, have better memory locality, and avoid boxing.

---

### DIFF
--- a/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
+++ b/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
@@ -15,6 +15,9 @@
  */
 package app.cash.redwood.protocol.host
 
+import androidx.collection.MutableIntObjectMap
+import androidx.collection.mutableIntObjectMapOf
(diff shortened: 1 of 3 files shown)


redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
```diff
@@ -57,10 +60,12 @@ public class HostProtocolAdapter<W : Any>(
...
   }
 
-  private val nodes = mutableMapOf<Id, ProtocolNode<W>>(
-    Id.Root to RootProtocolNode(container),
-  )
-  private val changedWidgets = mutableSetOf<ChangeListener>()
+  private val nodes =
+    mutableIntObjectMapOf<ProtocolNode<W>>(Id.Root.value, RootProtocolNode(container))
+
+  private val removeNodeById: (Id) -> Unit = { nodes.remove(it.value) }
```
JakeWharton, line 66: This lambda causes boxing (same as before) and will need switched to a dedicated `fun interface` in a follow-up.

swankjesse (approved):

PR #2322, opened 2024-09-24
Introduce a Snapshotter interface for tests

swankjesse:
I'm attempting to test that changes to the view hierarchy don't cause unnecessary extra layouts, so I need the parent view to remain stable across multiple snapshots. This new Snapshotter interface makes that straightforward.

This is currently only implemented for 'layout' tests. I'll follow up with 'lazylayout' tests next.

### DIFF
--- a/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
+++ b/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
@@ -38,7 +38,7 @@ abstract class AbstractFlexContainerTest<T : Any> {
    * currently opt-in, but will soon be the only supported mode.
    */
   open val incremental: Boolean
-    get() = true
+    get() = false
 
   abstract fun flexContainer(
     direction: FlexDirection,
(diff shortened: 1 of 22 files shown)


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -38,7 +38,7 @@ abstract class AbstractFlexContainerTest<T : Any> {
    * currently opt-in, but will soon be the only supported mode.
    */
   open val incremental: Boolean
-    get() = false
+    get() = true
```
swankjesse: * [x] Needs to be false.

redwood-layout-uiview/src/commonTest/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainerTest.kt
```diff
@@ -62,10 +61,13 @@ class UIViewFlexContainerTest(
   class UIViewTestFlexContainer internal constructor(
     private val delegate: UIViewFlexContainer,
   ) : TestFlexContainer<UIView>,
+    ResizableWidget<UIView>,
```
swankjesse, line 64: This was necessary!

JakeWharton (approved):

PR #2326, opened 2024-09-25
Share test widgets between different tests

swankjesse:
This extracts a new TestWidgetFactory type for layout and lazylayout tests.


swankjesse:
I’m unsure about some of the Paparazzi diffs. Will investigate and re-open.

PR #2327, opened 2024-09-25
Share test widgets between different tests

swankjesse:
A few snapshots changed because we now explicitly configuring a text locale, `View.TEXT_DIRECTION_LOCALE`.

What's really surprising about this is the snapshots with the string _'12 Angry Men'_ in them become _'Angry Men 12'_. Wild.


swankjesse:
Screenshot 2024-09-25 at 1 51 48 PM

JakeWharton (approved):

PR #2328, opened 2024-09-25
Guard against iOS overscroll reporting negative offset values.

colinrtwhite:
iOS can report negative values for scroll offset when overscrolling at the beginning (top for column/left for row) of the container. Truncate them to `0.0` for now.

---


JakeWharton (approved):

dnagler (approved):
Thanks! Love an easy fix :)

Issue #1040, opened 2023-05-10
Allocation reduction pass


JakeWharton:
Just a reminder to do our normal allocation-reduction pass over every module for free performance wins, even if minute.

- Switch loops to be indexed when we can trust the source of the list to be random access.
- Replace multi-step collection manipulation with traditional loops.
- Reduce/eliminate points of value class or JVM primitive boxing, if possible

etc.

JakeWharton:
Maybe we can write Detekt rules for some of these?

PR #2330, opened 2024-09-25
Eliminate boxing of protocol types for serialization

JakeWharton:
We still have a few cases of use within a Zipline interface that causes it, but those are rarely invoked (protocol mismatch handlers).

#1040 

---

### DIFF
--- a/redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/GuestProtocolAdapter.kt
+++ b/redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/GuestProtocolAdapter.kt
@@ -36,9 +36,9 @@ import kotlinx.serialization.json.Json
  *
  * This interface is for generated code use only.
  */
-public interface GuestProtocolAdapter : EventSink {
+public abstract class GuestProtocolAdapter : EventSink {
   @RedwoodCodegenApi
-  public val json: Json
+  public abstract val json: Json
 
   /**
    * Host versions prior to 0.10.0 contained a bug where they did not recursively remove widgets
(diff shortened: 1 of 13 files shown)


redwood-protocol/src/commonMain/kotlin/app/cash/redwood/protocol/protocol.kt
```diff
@@ -33,15 +33,35 @@ import kotlinx.serialization.json.buildJsonArray
 import kotlinx.serialization.json.jsonArray
 import kotlinx.serialization.json.jsonPrimitive
 
+/*
+Note: The types in this file are written in a careful way such that their serializable properties
+are scalars which can be serialized directly rather than using value classes are boxed within the
+generated kotlinx.serialization code.
```
JakeWharton, line 39: I don't love this, but kotlinx.serialization boxes _every_ value class instance on both serialization and deserialization paths, and this avoids that without changing the public API (aside from `invoke` vs constructor).

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/GuestProtocolAdapter.kt
```diff
@@ -127,14 +127,44 @@ public interface GuestProtocolAdapter : EventSink {
...
-    removedIds: List<Id> = listOf(),
+    removedIds: List<Id>,
   )
 
   @RedwoodCodegenApi
-  public fun removeWidget(id: Id)
+  public abstract fun removeWidget(id: Id)
+
+  @RedwoodCodegenApi
+  public val childrenVisitor: ProtocolWidget.ChildrenVisitor = if (synthesizeSubtreeRemoval) {
```
JakeWharton, line 145: Allocate only once, rather than once per children wrapper.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/GuestProtocolAdapter.kt
```diff
@@ -127,14 +127,44 @@ public interface GuestProtocolAdapter : EventSink {
...
 
   @RedwoodCodegenApi
-  public fun appendRemove(
+  public abstract fun appendRemove(
     id: Id,
     tag: ChildrenTag,
     index: Int,
     count: Int,
-    removedIds: List<Id> = listOf(),
+    removedIds: List<Id>,
```
JakeWharton, line 138: This caused boxing of the `Id` and `ChildrenTag` in JS if the caller omitted this last argument and relied on the default.

colinrtwhite (approved):
Nice

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/GuestProtocolAdapter.kt
colinrtwhite: Is there a way we can validate boxing doesn't occur in a test case? Or is it only possible when inspecting the generated code?

JakeWharton: I'll think about it. You definitely can see the boxing in a full compiler plugin, and maybe in FIR only. Probably also as a lint check.

Filed #2336

Issue #2128, opened 2024-06-20
Root component’s dimensions inconsistent across platforms


swankjesse:
Here’s a composition with a label that draws wider than the screen.

```
    @Composable
    override fun Show(flow: CashFlow.Environment, scope: CoroutineScope) {
      Column(
        width = Constraint.Wrap,
      ) {
        Text("hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello ")
      }
    }
```

And here’s what that looks like on an iOS and Android device:
![Screenshot 2024-06-20 at 10 40 08 AM](https://github.com/cashapp/redwood/assets/133019/6a19faaa-2da1-44d7-a5d1-d3413912b584)

These should be the same.

JakeWharton:
Related: #1859

swankjesse:
Looking into this now. On iOS our root view doesn’t care if its child component’s dimensions exceed its own. Here I’ve got a label that’s 965px wide in a parent that’s 393px wide.

<img width="720" alt="Screenshot 2024-09-26 at 2 27 03 PM" src="https://github.com/user-attachments/assets/739be8b3-30e2-466c-85cd-3045faf4ce92">

swankjesse:
# Root layout is pass-through

The root layout should be a pass-through layout:
 * On measure, it should forward its parent’s constraints down to its children.
 * On layout, its children should be forced to exactly match the parent.

I need to confirm we’re doing this on both platforms.

swankjesse:
# Column(width = WRAP) is different between platforms

```
    Box(
      width = Constraint.Fill,
      height = Constraint.Fill,
    ) {
      Column(
        width = Constraint.Wrap,
      ) {
        Text(
          modifier = Modifier.margin(Margin(start = 350.dp)),
          text = "hello_world"
        )
      }
    }
```

On Android we don’t permit child components to exceed the columns’ own bounds.
On iOS we do.

### Android
<img width="423" alt="Screenshot 2024-09-26 at 3 29 35 PM" src="https://github.com/user-attachments/assets/890871fc-8c4b-447d-9331-9e2dfe229e13">

### iOS
![Simulator Screenshot - iPhone 15 simulator - 2024-09-26 at 15 30 44](https://github.com/user-attachments/assets/14e35580-d775-49f2-aa46-53dc3e9108c2)

We should pick one and be consistent.

PR #2337, opened 2024-09-27
Fix Constraint.Wrap layouts when a view is larger than its parent

swankjesse:
These were incorrectly overflowing the parent container because we were applying the target size at the wrong layer.

This fixes an ugly inconsistency between iOS and Android, and makes the iOS layouts less surprising.

https://github.com/cashapp/redwood/issues/2128

---

### DIFF
--- a/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testChildIsConstrainedToParentWidth.1.png
+++ b/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testChildIsConstrainedToParentWidth.1.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:0a79c0d38522567fa6530cde363d3c0902f02b8a3dce2c7417183e745065a404
+size 82103
--- a/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testColumnThenRow.1.png
+++ b/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testColumnThenRow.1.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ef515d673afc46ae662ca0dc227efd17a8903434ede8fa7567f22bec6930db8c
-size 98527
+oid sha256:a6a7f711ad8a208ce6b6a0c7dac78d2dce3b55866fd062beaa4ec5a66f99d66e
+size 98586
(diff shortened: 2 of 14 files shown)


redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testColumnThenRow.1.png
swankjesse, line 1: I think this might be a difference in how we round floating point values, like `20.3` to either `20` or `21`. I might fix this small unexpected diff in a follow-up PR.

redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testChildIsConstrainedToParentWidth.1.png
swankjesse, line 1: Here’s what this one looks like without the fix. Horrible!

testChildIsConstrainedToParentWidth 1

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -44,12 +44,11 @@ internal class YogaUIView(
   }
 
   override fun intrinsicContentSize(): CValue<CGSize> {
-    return calculateLayoutWithSize(CGSizeMake(Size.UNDEFINED.toDouble(), Size.UNDEFINED.toDouble()))
+    return calculateLayoutWithSize(Size.UNDEFINED, Size.UNDEFINED)
   }
 
   override fun sizeThatFits(size: CValue<CGSize>): CValue<CGSize> {
-    val constrainedSize = size.useContents { sizeForConstraints(this) }
```
swankjesse, line 51: Rewriting the caller’s requested size was our previous mechanism to implement Constraint.Wrap. It didn’t work in special cases.

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -96,6 +95,24 @@ internal class YogaUIView(
...
+      Constraint.Fill -> width
+      else -> Size.UNDEFINED
+    }
+    rootNode.requestedMaxWidth = width
+
+    rootNode.requestedHeight = when (heightConstraint) {
+      Constraint.Fill -> height
+      else -> Size.UNDEFINED
+    }
+    rootNode.requestedMaxHeight = height
```
swankjesse, line 114: Previously we were never setting these on iOS! But we were setting them on *Android. Now we’re consistent.

JakeWharton (approved):

PR #2345, opened 2024-09-30
Use a single toString implementation for protocol nodes

JakeWharton:
This saves even more. For emoji sample:

 - Before: 11299596
 - After: 10953834

---


colinrtwhite (approved):
Nice!

PR #2346, opened 2024-10-01
Add a test to demonstrate a layout bug in Yoga

swankjesse:
We've accidentally set the maximum width of a cell to its maximum height.


redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testRowWithFixedWidthHasChildWithFixedHeight[LTR].png
Comment (swankjesse) [inline redwood-layout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiFlexContainerTest_testRowWithFixedWidthHasChildWithFixedHeight[LTR].png:1]: The Square 'A A A' box is broken. We’ve configured it to have a height of 50.dp, but it’s also getting a width of 50.dp.

JakeWharton (approved):
Is the UIView image what we expect then? Or is it also wrong for different reasons?

colinrtwhite (approved):

swankjesse:
The lazy layout images don’t support rows well.

PR #2352, opened 2024-10-01
Don't forget that snapshot tests must start with 'test'

swankjesse:


swankjesse:
The first commit adds a symbol processor to require `@Test`-annotated functions start with `test`.

The second commit fixes the lone infraction.

JakeWharton (approved):
We will need to migrate to KSP 2 or be blocked from upgrading to Kotlin 2.1.0 when it is released. Perhaps the strategy is to just get off my ass and build the MPP parameterization plugin so we can get rid of all of this

PR #2358, opened 2024-10-02
Apply modifiers as they change in UIViewFlexContainer

swankjesse:
This reduces the amount of preparation required to measure a layout.


swankjesse:
See also https://github.com/cashapp/redwood/pull/2357

JakeWharton (approved):

PR #2362, opened 2024-10-03
Fix a bug with Modifier.flex() nested in a row and column

### LINKED ISSUE #2018: Android: A widget with Modifier.flex() will do very bad things if nested inside of a Row + Column
### Summary

On Android, putting an element with `Modifier.flex()` inside a Row which is inside a Column will do... strange things. It seems to shrink the Row's width to 0, and also somehow expands the height by some large amount.

Works as expected on iOS.

### Repro

```
Column {
  Row {
    Text("Some text", modifier = Modifier.flex(1.0))
  }
}
```

### Screenshots 
image

### Attempted tweaks

* Works fine if using `Modifier.shrink` or `Modifier.grow` instead of flex

swankjesse:
This is a bug in Android only caused by not explicitly handling MeasureSpec.UNSPECIFIED.

Closes: https://github.com/cashapp/redwood/issues/2018

### DIFF
--- a/redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
+++ b/redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
@@ -73,11 +73,13 @@ internal class YogaLayout(context: Context) : ViewGroup(context) {
     calculateLayout(
       requestedWidth = when {
         widthMode == MeasureSpec.EXACTLY -> widthSize.toFloat()
+        widthMode == MeasureSpec.UNSPECIFIED -> Size.UNDEFINED
         widthConstraint == Constraint.Fill -> widthSize.toFloat()
         else -> Size.UNDEFINED
       },
       requestedHeight = when {
         heightMode == MeasureSpec.EXACTLY -> heightSize.toFloat()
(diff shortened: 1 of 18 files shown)


redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/YogaLayout.kt
```diff
@@ -73,11 +73,13 @@ internal class YogaLayout(context: Context) : ViewGroup(context) {
     calculateLayout(
       requestedWidth = when {
         widthMode == MeasureSpec.EXACTLY -> widthSize.toFloat()
+        widthMode == MeasureSpec.UNSPECIFIED -> Size.UNDEFINED
         widthConstraint == Constraint.Fill -> widthSize.toFloat()
         else -> Size.UNDEFINED
       },
       requestedHeight = when {
         heightMode == MeasureSpec.EXACTLY -> heightSize.toFloat()
+        heightMode == MeasureSpec.UNSPECIFIED -> Size.UNDEFINED
```
swankjesse, line 82: Without the fix this nested spacer is asked to size itself using (mode=AT_MOST, size=0) and it returns a height of 0.

JakeWharton (approved):

PR #2368, opened 2024-10-04
Remove old default expression annotation

### LINKED ISSUE #2225: Parse default expressions instead of `@Default` annotation

JakeWharton:
Regular default expressions have worked since the last release.

Closes #2225

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ New:
 
 Changed:
 - Drop support for non-incremental layouts in `Row` and `Column`.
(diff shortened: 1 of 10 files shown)


swankjesse (approved):
⚰️ ☠️ 🪦

colinrtwhite (approved):

PR #2374, opened 2024-10-08
Replace CodeListener with RedwoodView.Root

### LINKED ISSUE #2174: Use a Redwood schema for loading and error UIs
Right now loading and error states are pretty hacky! [CodeListener](https://github.com/cashapp/redwood/blob/c6dcde6d7aebce016092690c6095bc9d507304c1/redwood-treehouse-host/src/commonMain/kotlin/app/cash/redwood/treehouse/CodeListener.kt#L21) has callbacks for each of our 3 states (loading, loaded, detached), and users must write special code for each platform to build views when they receive these callbacks.

I propose we use Redwood’s mechanisms for widgets to implement loading and error states. We would start with a schema like this:

```
@Schema([DynamicContentRoot::class])
public interface DynamicContent

@Widget(1)
public data class DynamicContentRoot(
  /** 
   * How many different versions of code have been loaded.
   * 
   * This may be used to trigger an animation each time that new code
   * is updated. 
   */
  @Property(1)
  val loadCount: Int,

  /**
   * True if this UI is actively being choreographed by application
   * logic. False if it's not yet loaded, crashes, or otherwise stopped.
   */
  @Property(2)
  val attached: Boolean,

  /** 
   * The current dynamic content. If the application logic is stopped or
   * crashed this will retain a snapshot of the most-recent content, but
   * the content will ignore user actions.
   */
  @Children(1)
  val content: () -> Unit,

  /**
   * The uncaught exception that caused this UI to be detached.
   */
  @Property(3)
  @Default("null")
  val uncaughtException: Throwable? = null,

  /**
   * Call this to restart the UI if is has hung or crashed.
   */
  @Property(4)
  val restart: () -> Unit,
)
```

For each host platform, the app would implement a binding for DynamicContentRoot.

But the guest wouldn’t know about this schema. Instead, the Treehouse content bindings would synthesize updates in response to app lifecycle events.

squarejesse:
This moves loading/error/ready tracking to live beside the content, rather than as something that side-effects the content.

This is not compatible with the old API and there is no shim.

Closes: https://github.com/cashapp/redwood/issues/2174

---


redwood-treehouse-host-composeui/src/commonMain/kotlin/app/cash/redwood/treehouse/composeui/TreehouseContent.kt
```diff
@@ -106,8 +113,72 @@ public fun <A : AppService> TreehouseContent(
...
+
+  override fun restart(restart: (() -> Unit)?) {
+    this.restart = restart
   }
+
+  override val value: @Composable () -> Unit = {
+    Render()
+  }
+
+  @Composable
```
squarejesse: * [x] Move this out of here and into the EmojiSearch sample

redwood-treehouse-host/src/commonTest/kotlin/app/cash/redwood/treehouse/FakeRoot.kt
```diff
@@ -0,0 +1,60 @@
...
+
+  override val value: WidgetValue
+    get() = error("unexpected call")
+  override var modifier: Modifier = Modifier
+
+  override fun contentState(loadCount: Int, attached: Boolean, uncaughtException: Throwable?) {
+    // Canonicalize "java.lang.Exception(boom!)" to "kotlin.Exception(boom!)".
+    val exceptionString = uncaughtException?.toString()?.replace("java.lang.", "kotlin.")
+
+    // TODO(jwilson): this is a backwards-compatibility shim. Emit a simpler event.
```
squarejesse, line 50: For follow-up

redwood-widget/src/androidMain/kotlin/app/cash/redwood/widget/RedwoodLayout.kt
```diff
@@ -37,17 +37,16 @@ import kotlinx.coroutines.flow.StateFlow
 @SuppressLint("ViewConstructor")
 public open class RedwoodLayout(
   context: Context,
+  public final override val root: RedwoodView.Root<View> = ViewRoot(context),
```
squarejesse: On Android, this accepts the _interface type_, `RedwoodView.Root` ...

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIView.kt
```diff
@@ -38,17 +38,16 @@ import platform.UIKit.UIUserInterfaceStyle
 import platform.UIKit.UIView
 
 public open class RedwoodUIView(
-  public val view: UIView,
+  final override val root: UIViewRoot,
```
squarejesse, line 41: But on iOS, this accepts our concrete base implementation. This is because I haven’t yet figured out how to add a bounds change listener to a view I didn’t subclass

samples/counter/ios-uikit/CounterApp/CounterViewController.swift
```diff
@@ -6,19 +6,18 @@ class CounterViewController : UIViewController {
     private var delegate: CounterViewControllerDelegate!
 
     override func viewDidLoad() {
-        let container = UIStackView()
```
squarejesse, line 9: This is a behavior change. For a follow-up I need to use a `Row` in the Compose-side code for the counter.

samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchActivity.kt
```diff
@@ -92,26 +93,48 @@ class EmojiSearchActivity : ComponentActivity() {
...
-          LinearLayout.LayoutParams(MATCH_PARENT, MATCH_PARENT),
-        )
+      if (uncaughtException != null) {
+        scope.launch {
+          delay(2_000.milliseconds)
+          restart?.invoke()
+        }
+      }
+
+      for (child in viewGroupChildren) {
```
squarejesse: This still isn’t that great . . . I prefer the way this works in Compose

samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/LoadingView.kt
```diff
@@ -0,0 +1,38 @@
...
+import android.annotation.SuppressLint
+import android.content.Context
+import android.view.Gravity.CENTER
+import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
+import android.widget.FrameLayout
+import android.widget.ProgressBar
+import androidx.appcompat.view.ContextThemeWrapper
+
+@SuppressLint("ViewConstructor")
+internal class LoadingView(
```
squarejesse, line 27: New: now there’s a spinner when launching emoji search

squarejesse:
@JakeWharton this is ready for review.

I did an in-place swap from the old type to the new type rather than a graceful migration, which makes the PR larger than I generally like.

JakeWharton:
Yeah it's... Big. I started today but had to leave early. Will finish tomorrow morning.

samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchActivity.kt
```diff
@@ -92,26 +95,48 @@ class EmojiSearchActivity : ComponentActivity() {
...
+        if (child is ExceptionView || child is LoadingView) {
+          removeView(child)
```
JakeWharton: It is not safe to remove a view like this while iterating. However, the iterator is actually a `MutableIterator` that you can call `remove()` on: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core-ktx/src/main/java/androidx/core/view/ViewGroup.kt;l=89-99;drc=1e02caf7787e6fc64ec1913bf6e5cbbab1724a13. Unfortunately that means you can't use a nice for loop anymore.

samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchActivity.kt
```diff
@@ -92,26 +95,48 @@ class EmojiSearchActivity : ComponentActivity() {
...
+        }
+      }
+      if (loadCount == 0) {
+        addView(LoadingView(context))
+      }
+      if (uncaughtException != null) {
+        addView(ExceptionView(context, uncaughtException))
       }
```
JakeWharton, line 135: Seems a bit suspect that the loop will leave both `LoadingView` and `ExceptionView` instances, but then also possible add new ones as well.

samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchActivity.kt
```diff
@@ -92,26 +95,48 @@ class EmojiSearchActivity : ComponentActivity() {
...
     ) {
-      if (exception != null) {
-        treehouseLayout.reset()
-        treehouseLayout.addView(
-          ExceptionView(treehouseLayout, exception),
-          LinearLayout.LayoutParams(MATCH_PARENT, MATCH_PARENT),
-        )
+      if (uncaughtException != null) {
+        scope.launch {
+          delay(2_000.milliseconds)
```
JakeWharton: Not that it matters, but could be
```suggestion
          delay(2.seconds)
```

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/RedwoodView.kt
```diff
@@ -22,15 +22,53 @@ import kotlinx.coroutines.flow.StateFlow
...
   public val savedStateRegistry: SavedStateRegistry?
 
-  /**
-   * This should at minimum clear all [children].
-   *
-   * Invoke when switching the backing composition to prepare the view for an initial load.
-   */
-  public fun reset()
+  @ObjCName("Root", exact = true)
+  public interface Root<W : Any> : Widget<W> {
```
JakeWharton, line 31: This seems like it exists at the wrong layer of abstraction since Redwood has no concept of loading or errors–just content. Eliminating `CodeListener` is a big enough win for now. We can figure out how to move it later.

JakeWharton: I don't see why this extends `Widget`. Do we use that capability anywhere? I haven't seen any usage that relies on it, but perhaps that's just because the diff is pretty big.

JakeWharton (approved):
I reviewed as much as I could. Don't want to hold it up any more.

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/RedwoodView.kt
squarejesse: We’re using the `value` field to get a component to place in the parent container. But that doesn’t need to be polymorphic.

samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchActivity.kt
squarejesse: Agreed. In practice it’s unlikely to transition from loading to loading again, or exception to exception again.

samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchActivity.kt
squarejesse: Oh hell yes. I was exploiting the fact that this will only contain one child element if its either of those types. But I prefer your fix with an iterator.

PR #2388, opened 2024-10-16
Adopt Burst in Redwood

squarejesse:


JakeWharton (approved):
We have a bunch of hand-rolled ones, too. But I can get them in a follow-up.

squarejesse:
Blocked on this: https://github.com/cashapp/burst/pull/28

PR #2389, opened 2024-10-17
Explicitly configure layouts in snapshot tests

squarejesse:
We weren't fully configuring Row and Column layouts, which caused the (undocumented) yoga defaults to be used in tests.


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -119,7 +134,14 @@ abstract class AbstractFlexContainerTest<T : Any> {
     val container = flexContainer(flexDirection)
     container.width(width)
     container.height(height)
-    container.add(widgetFactory.text(movies.first()))
+    container.add(
+      widgetFactory.text(
+        text = movies.first(),
+        modifier = Modifier
+          .then(HorizontalAlignmentImpl(CrossAxisAlignment.Stretch))
+          .then(VerticalAlignmentImpl(CrossAxisAlignment.Stretch)),
```
squarejesse, line 142: This restores the old behavior to preserve the expectation of the test

JakeWharton (approved):

PR #2391, opened 2024-10-21
Flatten RedwoodView and RedwoodView.Root

squarejesse:
This is a step in simplifying the complex interaction between these two classes. It also removes an unnecessary layer from the view tree.

We also had a potential bug on hot reloads, where the previous content was detached but not necessarily removed. With this change we clear the child views by default on a hot reload.

---


samples/emoji-search/android-views/src/main/kotlin/com/example/redwood/emojisearch/android/views/EmojiSearchActivity.kt
```diff
@@ -94,39 +94,35 @@ class EmojiSearchActivity : ComponentActivity() {
...
         }
       }
 
-      val i = iterator()
-      while (i.hasNext()) {
-        val child = i.next()
-        if (child is ExceptionView || child is LoadingView) {
-          i.remove()
-        }
-      }
```
squarejesse, line 129: Let the base class just clear everything

JakeWharton (approved):

redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/RedwoodView.kt
```diff
@@ -22,53 +22,51 @@ import kotlinx.coroutines.flow.StateFlow
...
+   *     content stops before it emits its initial view tree.
+   * @param attached true if the content is interactive. This is false if it has stopped or
+   *     crashed.
+   * @param uncaughtException the exception that caused the content to detach.
+   */
+  public fun contentState(
+    loadCount: Int,
+    attached: Boolean,
+    uncaughtException: Throwable? = null,
+  )
```
squarejesse, line 66: For a follow-up, I’ll refactor these arguments to match what the content can actually be:

```kotlin
sealed class ContentState {
  /** 
   * This view has no content. This is the initial state of the view.
   * 
   * This may occur if the content is gracefully stopped, in which case
   * the view may continue to display the previous content, though it
   * will not be interactive.
   */
  object Detached : ContentState

  /** 
   * Content isn't ready yet. It’s busy downloading, launching, or computing 
   * the initial view tree.
   */
  object Loading : ContentState()

  /** 
   * Content is ready and responsive to interactions. 
   */
  data class Attached(val loadCount: Int): ContentState()

  /** 
   * This view has crashed due to a bug.
   */
  data class Crashed(uncaughtException: Throwable?): ContentState()
}
```

PR #2395, opened 2024-10-22
Apply Defaults explicitly

swankjesse:


redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testChildren_Wrap_Center_Center.png
squarejesse, line 1: We now frame the iOS and Android box widgets the same way, so we get consistent results between platforms. 

https://github.com/cashapp/redwood/blob/trunk/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewBoxTestHost/testChildren_Wrap_Center_Center.1.png

JakeWharton (approved):

PR #2397, opened 2024-10-22
Implement ViewBox with manual measurement

squarejesse:
Previously ViewBox delegated to FrameLayout. This worked okay but it was potentially difficult to track behaviors between View and UIView.

This implementation is based on the UIViewBox class implementation.

This fixes a bug where layouts incorrectly didn't handle changes when modifiers were removed.

This changes how Box's margins are implemented. Previously the margins were external, which required the hosting layout to honor them as layout constraints. That was not supported when the enclosing layout was a Row or Column.

---


redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testChildrenModifierChanges_empty.png
squarejesse, line 1: Now matches iOS!
https://github.com/cashapp/redwood/blob/trunk/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewBoxTestHost/testChildrenModifierChanges.Empty.png

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testMargins.png
squarejesse, line 1: Previously: the bounds of the Box excluded the margins, and so the background wasn't drawn there
Now: the bounds of the Box include the margins, so the background is drawn there

This should be easier to work with!

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testMarginsAndAlignment.png
squarejesse, line 1: FrameLayout and centering worked differently for asymmetric margins. This is now consistent with iOS, and a pretty reasonable behavior.

redwood-layout-view/src/main/kotlin/app/cash/redwood/layout/view/ViewBox.kt
```diff
@@ -15,210 +15,307 @@
...
-      when (childModifier) {
-        is HorizontalAlignment -> {
-          horizontalAlignment = childModifier.alignment
-        }
+/**
+ * Measures and lays out one child view at a time.
+ *
+ * This class is mutable and reused to avoid object allocation.
+ */
+private class Measurer {
```
squarejesse, line 152: The iOS implementation doesn’t have a class like this, and I think it’s worked out quite nicely. It makes it easy to share layout logic between measure and layout steps. It also avoids allocations!

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testChildrenModifierChanges_empty.png
JakeWharton, line 1: This one seems broken? The red text box has 30dp of margin: https://github.com/cashapp/redwood/blob/f377665687b0de4415cef38f9ab38c70099badd8/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt#L452-L453

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testMargins.png
JakeWharton, line 1: I think this is also a breakage. If someone creates a widget
```kotlin
@Widget
data class Toolbar(
  @Children(1) val left: () -> Unit,
  @Children(2) val right: () -> Unit,
)
```
and then implements those slots with `FrameLayout`s, someone putting a `Box(modifier = Modifier.margin(30.dp))` would not see their margins reflected.

https://github.com/cashapp/redwood/blob/f377665687b0de4415cef38f9ab38c70099badd8/redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractBoxTest.kt#L347-L348

JakeWharton (approved):

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testChildrenModifierChanges_empty.png
JakeWharton: Oh I get it. This is the "after". The "empty" is misleading because we use that for the initial snapshot in other tests...

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testMargins.png
JakeWharton: Is this just the padding vs. margin naming thing we talked about yesterday then?

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testChildrenModifierChanges_empty.png
squarejesse: Yep. Need to remove a modifier!

redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testMargins.png
squarejesse: In this case, yes exactly.

Box has two kinds of margins: 
 - the margin property on the Box itself, that works like padding
 - the margin modifier on each child, which works like a margin
 
I think it would be good to rename the margin property to padding.

PR #2400, opened 2024-10-23
Make UIViewBox use the same patterns as ViewBox

swankjesse:
We now include the margin when measuring on 'Stretch' alignment. This was causing a layout bug previously.

We now use sizeThatFits exclusively, rather than using sizeThatFits when measuring and systemLayoutSizeFittingSize during layout.


redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewBoxTestHost/testMarginsAndStretch.1.png
swankjesse, line 1: Bug fix!

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
```diff
@@ -127,242 +128,187 @@ internal class UIViewBox :
...
+    // Measure the view if don't have an exact width or height.
+    val mustMeasureWidth = requestedWidth.isNaN() &&
+      horizontalAlignment != CrossAxisAlignment.Stretch
+    val mustMeasureHeight = requestedHeight.isNaN() &&
+      verticalAlignment != CrossAxisAlignment.Stretch
 
-        is Height -> with(Density.Default) {
-          height = maxOf(height, childModifier.height.toPx())
-        }
+    if (!mustMeasureWidth && !mustMeasureHeight) {
```
swankjesse, line 265: Android doesn’t have this optimization. It might benefit from it.

swankjesse: (It’s more difficult because on Android measuring may have essential side-effects)

JakeWharton (approved):

Issue #2231, opened 2024-08-02
Make generated `Widget` subinterfaces into abstract classes


JakeWharton:
We want to generate behavior into them. This will also eliminate the annoying `override var modifier: Modifier = Modifier` boilerplate that every implementation has to do.

JakeWharton:
This will allow generating a nice `toString()` automatically which will be useful for leak detection

JakeWharton:
This doesn't actually solve our problems. Swift lacks abstract classes.

We may have to generate two types instead. A `Widget` and a `Binding`. You implement the `Binding` which is a pure delegate (maybe as a stateless thing), and then `Widget` wraps an instance + state (like modifier, tree walking) + stateless binding.

JakeWharton:
That is now tracked by #2619

PR #2402, opened 2024-10-23
Migrate `Row` and `Column` implementations away from subtypes

JakeWharton:
When widgets switch to abstract classes, this will no longer be allowed.

Refs #2231 

---


JakeWharton:
Argh I need to pull in more local changes for this one. It might get BIG.

JakeWharton:
Going to do LazyList split PR first since it doesn't have so many internal users

PR #2405, opened 2024-10-23
Honor margins internally in UIViewBox

swankjesse:
Previously the margins were external, which required the enclosing layout to honor them. That was not supported when the enclosing layout was a Box, Row, or Column.


JakeWharton (approved):
This does not seem right to me. The box honors margins on its children by offsetting them (exterior space) but applies its own margins against its children (interior space).

What happens if you nest boxes? Do the margins of the inner one get applied twice? Once to size itself and once in the parent to offset it as a child?

swankjesse:
What Box is calling margins is actually padding. They stack.

swankjesse:
Critically, the `margin` thing that Box is honoring here is *not* a modifier, it’s a regular property.

JakeWharton:
Ah, right. So should we rename it?

PR #2406, opened 2024-10-23
Fix the centering math in Box on Android

squarejesse:
The code was incorrectly not subtracting margins off before centering.


redwood-layout-view/src/test/snapshots/images/app.cash.redwood.layout.view_ViewBoxTest_RTL_testChildExplicitHeight.png
squarejesse, line 1: The left one wasn’t centered. Ouch.

JakeWharton (approved):

PR #2410, opened 2024-10-24
Migrate Compose UI `Row`+`Column` away from subtype

JakeWharton:
When widgets switch to abstract classes, this will no longer be allowed.

Refs #2231 

---


JakeWharton:
I swear pulling these into their own PRs is taking longer than actually migrating them in the first place

squarejesse (approved):

colinrtwhite (approved):

PR #2411, opened 2024-10-24
Remove child helpers from `TestFlexContainer` interface

JakeWharton:
We can implement these generally now that the backing widget list provides the existing count for new insertion positions.

---


redwood-layout-shared-test/src/commonMain/kotlin/app/cash/redwood/layout/AbstractFlexContainerTest.kt
```diff
@@ -63,6 +63,18 @@ abstract class AbstractFlexContainerTest<T : Any> {
     onScroll(null)
   }
 
+  protected fun <T : Any> TestFlexContainer<T>.add(widget: Widget<T>) {
+    addAt(children.widgets.size, widget)
+  }
+
+  protected fun <T : Any> TestFlexContainer<T>.addAt(index: Int, widget: Widget<T>) {
```
squarejesse, line 70: image

squarejesse (approved):

PR #2413, opened 2024-10-26
Add Kotlin delegate

AntonButov:
Why don’t you use Kotlin delegates? Maybe you want code-generate the delegates, and it will have problems...


JakeWharton:
These types are migrating to abstract classes (#2231) where delegation won't work.

Issue #2392, opened 2024-10-21
`RedwoodView` should not have Treehouse lifecycle


JakeWharton:
This class is meant to provide a frame and integrations to the platform UI toolkit. The Treehouse loading/error stuff should exist entirely at the Treehouse layer and render _within_ the frame–not as part of it.

Related to #1475.

squarejesse:
I have a plan:

### RedwoodView

I’m going to make loading & error screens _children_ of the `RedwoodView`, rather than _states_ of the `RedwoodView`:

```
public interface DynamicContentWidgetFactory<W : Any> {
  public fun Loading(): Widget
  public fun Crashed(throwable: Throwable): Widget
}
```

I’ll need to create a default implementation that uses an empty widget.

### Observing Content State

With the above I still want to notify the host of code reloads and detached content. I’ll do that separately, with a new `Content.State` sealed class and a new property on `Content`:

```
public interface Content {
  public val state: StateFlow<State>

  sealed class State {
    public data object Detached : ContentState()
    public data object Loading : ContentState()
    public data class Attached(val loadCount: Int) : ContentState()
    public data class Crashed(val uncaughtException: Throwable): ContentState()
  }
}
```

### Redundant?

One thing I struggled with was exposing states like `Loading` two ways: as a child element and as a `StateFlow`. But I think it’s the best option for giving the application the power to put loading state in the UI or not. We need that power in Cash App!

It should also let us avoid detaching the root view’s `children`, which is currently super awkward.

PR #2418, opened 2024-10-30
Split Treehouse lifecycle out of RedwoodView

swankjesse:
https://github.com/cashapp/redwood/issues/2392

---


swankjesse:
My BIG INSIGHT in this PR is that everything works better if I expose content state two ways:
 - First as child components. So `TreehouseViews` can show loading spinners and sad-tab icons inline if that’s their thing.
 - ALSO as state. So `TreehouseViews` can also display that stuff outside of the content if that’s their thing.

All of my previous attempts were clumsy because they treated these two cases as symmetric.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
```diff
@@ -425,7 +425,7 @@ private class RootProtocolNode<W : Any>(
     }
 
   override fun detach() {
-    children.detach()
+    // Do nothing because 'children' is owned by the host's TreehouseView.
```
JakeWharton: The Redwood protocol doesn't _strictly_ have to be used by Treehouse, although that is certainly the overwhelmingly common usage. Maybe put a note on the class docs up above, or on `close()`.

JakeWharton (approved):

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
swankjesse: Yeah, good point. I should have said `RedwoodView` here instead.

JakeWharton: Oh even better. I thought it was something specific to treehouse.

PR #2426, opened 2024-11-04
Fix measurement of Row and Column on iOS inside of UIStackView

squarejesse:
UIStackView uses UIView.intrinsicContentSize on its subviews, but our Row and Column implementations don't have a natural implementation of this API. In particular, if any subview wraps its contents (so its height depends on its available width), there's no single intrinsicContentSize().

We work around this by following some online advice: invalidate the intrinsic size when the bounds change, and then use those bounds when computing the new intrinsic size.

This is more fragile and stateful than I'd like, but it seems like a common practice for working around UIStackView's API.

---

### DIFF
--- a/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testIntrinsicContentSizeWhenSubviewsRequireScrolling.1.png
+++ b/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testIntrinsicContentSizeWhenSubviewsRequireScrolling.1.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:02084488d7748ef0220969e57142513fd94837b231f8f50b0a5b743cabf872fa
+size 271099
--- a/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testIntrinsicContentSizeWhenSubviewsWrap.1.png
+++ b/redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testIntrinsicContentSizeWhenSubviewsWrap.1.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:8c32f78bc7d550562be65654b83509d108266714ddf87169022369f56e149e3f
+size 98569
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
@@ -38,8 +38,8 @@ internal class YogaUIView : UIScrollView(cValue { CGRectZero }), UIScrollViewDel
 
   var onScroll: ((Px) -> Unit)? = null
 
-  private var widthForIntrinsicSize = UIViewNoIntrinsicMetric
-  private var heightForIntrinsicSize = UIViewNoIntrinsicMetric
+  private var fillWidth = UIViewNoIntrinsicMetric
+  private var fillHeight = UIViewNoIntrinsicMetric
 
   init {
     // TODO: Support scroll indicators.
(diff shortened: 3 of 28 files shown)


redwood-lazylayout-composeui/src/test/snapshots/images/app.cash.redwood.layout.composeui_ComposeUiLazyListTest_RTL_testIntrinsicContentSizeWhenSubviewsRequireScrolling.png
squarejesse, line 1: More ComposeUI layouts that are broken. This is a problem, but it’s not new and it’s not urgent.

redwood-snapshot-testing/src/iosMain/kotlin/app/cash/redwood/snapshot/testing/UIViewTestWidgetFactory.kt
```diff
@@ -120,8 +120,8 @@ class UIViewSimpleColumn : SimpleColumn<UIView> {
 
   override val value = UIStackView(CGRectZero.readValue()).apply {
     this.axis = UILayoutConstraintAxisVertical
-    this.alignment = UIStackViewAlignmentLeading
-    this.distribution = UIStackViewDistributionFill
+    this.alignment = UIStackViewAlignmentFill
+    this.distribution = UIStackViewDistributionEqualSpacing
```
squarejesse: I had better luck with these ones

redwood-snapshot-testing/src/iosMain/kotlin/app/cash/redwood/snapshot/testing/UIViewTestWidgetFactory.kt
```diff
@@ -144,8 +144,8 @@ class UIViewScrollWrapper : ScrollWrapper<UIView> {
       scrollView.addSubview(value)
       value.translatesAutoresizingMaskIntoConstraints = false
       value.leadingAnchor.constraintEqualToAnchor(scrollView.leadingAnchor).active = true
+      value.widthAnchor.constraintEqualToAnchor(scrollView.widthAnchor).active = true
```
squarejesse: It’s way too cute, but UIScrollView scrolls horizontally if you bind `leading` and `trailing`, but doesn’t scroll horizontally if you bind `leading` and `width`.

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -45,7 +49,48 @@ internal class YogaUIView : UIScrollView(cValue { CGRectZero }), UIScrollViewDel
...
+      }
+    }
+
+    super.setBounds(bounds)
+  }
+
+  override fun invalidateIntrinsicContentSize() {
+    super.invalidateIntrinsicContentSize()
+    this.widthForIntrinsicSize = UIViewNoIntrinsicMetric
+    this.heightForIntrinsicSize = UIViewNoIntrinsicMetric
```
squarejesse: I don’t have a test case to prove that this override is necessary. I might try to make one? That might be tricky

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -45,7 +49,48 @@ internal class YogaUIView : UIScrollView(cValue { CGRectZero }), UIScrollViewDel
...
+   *
+   *  1. Making [intrinsicContentSize] depend on the mostly-recently applied bounds
+   *  2. Invalidating it each time the bounds change
+   *
+   * This will result in an additional layout pass when the parent view uses [intrinsicContentSize].
+   */
+  override fun setBounds(bounds: CValue<CGRect>) {
+    val newWidth = bounds.useContents { size.width }
+    val newHeight = bounds.useContents { size.height }
+    if (isColumn()) {
```
squarejesse: (Maybe instead I wanna check if the width mode is Fill here)

redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testIntrinsicContentSizeWhenSubviewsRequireScrolling.1.png
squarejesse, line 1: Without the fix, this thing’s intrinsicContentSize is is 10,000 pixels wide and 20 pixels tall

testIntrinsicContentSizeWhenSubviewsRequireScrolling 1

redwood-layout-uiview/RedwoodLayoutUIViewTests/__Snapshots__/UIViewFlexContainerTestHost/testIntrinsicContentSizeWhenSubviewsWrap.1.png
squarejesse, line 1: Similarly, without the fix each subview is super wide and not very tall

testIntrinsicContentSizeWhenSubviewsWrap 1

edahle (approved):
🔥 🔥

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
```diff
@@ -45,7 +49,54 @@ internal class YogaUIView : UIScrollView(cValue { CGRectZero }), UIScrollViewDel
...
-  override fun intrinsicContentSize(): CValue<CGSize> = calculateLayout()
+  /**
+   * The intrinsic size is broken by design if any subview's height depends on its width (or
+   * vice-versa). For example, if a subview is UILabel that wraps, we need to know how wide the
+   * label is before we can compute that label's height.
+   *
+   * We work around this by:
+   *
+   *  1. Making [intrinsicContentSize] depend on the mostly-recently applied bounds
+   *  2. Invalidating it each time the bounds change
```
colinrtwhite, line 60: This work around sounds good to me, though @edahle do you know how a `UITextView` and other views that depend on knowing the width to compute the height provide a reasonable `intrinsicContentSize`? Maybe private APIs of some kind?

colinrtwhite (approved):
Thanks for fixing this! 🙏🏻

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/YogaUIView.kt
squarejesse: UILabel has a feature called [preferredMaxLayoutWidth](https://developer.apple.com/documentation/uikit/uilabel/1620534-preferredmaxlayoutwidth) that you can use for this.

PR #2429, opened 2024-11-05
Fix a bug where Box's margins weren't included in the measurement

squarejesse:
We included it in layout but not measurement, which led to bad renders.

Also fix UIViewBox to override intrinsicContentSize() so that it can be used as a child of a UIStackView.

### DIFF
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
@@ -36,12 +36,12 @@ import kotlinx.cinterop.convert
 import kotlinx.cinterop.readValue
 import kotlinx.cinterop.useContents
 import platform.CoreGraphics.CGFloat
-import platform.CoreGraphics.CGRect
 import platform.CoreGraphics.CGRectMake
 import platform.CoreGraphics.CGRectZero
 import platform.CoreGraphics.CGSize
 import platform.CoreGraphics.CGSizeMake
 import platform.UIKit.UIView
+import platform.UIKit.UIViewNoIntrinsicMetric
 import platform.darwin.NSInteger
 
 internal class UIViewBox :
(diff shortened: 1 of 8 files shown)


redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
```diff
@@ -118,6 +118,9 @@ internal class UIViewBox :
       }
     }
 
+    override fun intrinsicContentSize() =
+      sizeThatFits(CGSizeMake(UIViewNoIntrinsicMetric, UIViewNoIntrinsicMetric))
```
squarejesse, line 122: The easiest way to confirm that measurement is happening properly is to put the `Box` itself in another container that does measurement. And the easiest way to do that is to make sure the `Box` overrides this function.

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
```diff
@@ -212,9 +221,13 @@ private class Measurer {
       boxMarginTop = boxMargin.top.toPx()
       boxMarginBottom = boxMargin.bottom.toPx()
     }
-    boxFrame.useContents {
-      frameWidth = (size.width - boxMarginStart - boxMarginEnd).coerceAtLeast(0.0)
-      frameHeight = (size.height - boxMarginTop - boxMarginBottom).coerceAtLeast(0.0)
+    this.frameWidth = when {
+      boxWidth == UIViewNoIntrinsicMetric -> UIViewNoIntrinsicMetric
+      else -> (boxWidth - boxMarginStart - boxMarginEnd).coerceAtLeast(0.0)
+    }
```
squarejesse, line 227: Avoid doing math when one of the values is `UIViewNoIntrinsicMetric`

dellisd (approved):

PR #2436, opened 2024-11-08
Accept a ZiplineHttpClient creating a TreehouseAppFactory

swankjesse:
The current OkHttpClient is what we ultimately want to provide, but that forces us to create the OkHttpClient eagerly and we'd rather do that lazily.

---


JakeWharton (approved):
iOS too?

swankjesse:
I don’t think iOS is as opinionated on how its networking is initialized.

PR #2438, opened 2024-11-08
Switch back to `MutableMap` to avoid scatter map bug

JakeWharton:
Sometimes the scatter-based maps can return values from other keys when inserting a new key. See https://issuetracker.google.com/issues/378077858.

---

### DIFF
--- a/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
+++ b/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
@@ -36,6 +36,7 @@ import app.cash.redwood.protocol.RedwoodVersion
 import app.cash.redwood.protocol.WidgetTag
 import app.cash.redwood.widget.ChangeListener
 import app.cash.redwood.widget.Widget
+import kotlin.collections.mutableMapOf
 import kotlin.native.ObjCName
 
 /**
@@ -60,8 +61,9 @@ public class HostProtocolAdapter<W : Any>(
     is GeneratedHostProtocol -> factory
   }
 
+  // TODO Use MutableIntObjectMap once https://issuetracker.google.com/issues/378077858 is fixed.
   private val nodes =
-    mutableIntObjectMapOf<ProtocolNode<W>>(Id.Root.value, RootProtocolNode(container))
+    mutableMapOf<Int, ProtocolNode<W>>(Id.Root.value to RootProtocolNode(container))
 
   private val removeNodeById = IdVisitor { nodes.remove(it.value) }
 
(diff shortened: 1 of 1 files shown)


redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
```diff
@@ -235,9 +236,10 @@ public class HostProtocolAdapter<W : Any>(
       // Must have a Create node that precedes it.
       if (lastCreatedId != change.id) continue
 
-      idToNode[change.id.value] = ReuseNode(
-        widgetId = change.id,
-        childrenTag = ChildrenTag.Root,
+      idToNode[lastCreatedId.value] = ReuseNode(
+        widgetId = lastCreatedId,
```
JakeWharton, line 241: Switch to local which already has the same value as `change.id`

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
```diff
@@ -359,19 +361,17 @@ public class HostProtocolAdapter<W : Any>(
      * descendants into the nodes map.
      */
     fun assignPooledNodeRecursive(
-      nodes: MutableIntObjectMap<ProtocolNode<W>>,
+      nodes: MutableMap<Int, ProtocolNode<W>>,
       changesAndNulls: Array<Change?>,
       pooled: ProtocolNode<W>,
     ) {
       // Reuse the node.
+      pooled.id = widgetId
```
JakeWharton, line 370: No need to lookup the `Create`, we already have a member with the new ID.

JakeWharton:
Took a quick look around at other scatter-based map and set usage. They don't have this long-lived put/remove pattern.

redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
```diff
@@ -235,9 +236,10 @@ public class HostProtocolAdapter<W : Any>(
       // Must have a Create node that precedes it.
       if (lastCreatedId != change.id) continue
 
-      idToNode[change.id.value] = ReuseNode(
-        widgetId = change.id,
-        childrenTag = ChildrenTag.Root,
+      idToNode[lastCreatedId.value] = ReuseNode(
+        widgetId = lastCreatedId,
+        // This is the root of the reuse tree, so it will never have a children tag from a parent.
```
squarejesse, line 242: Thanks

squarejesse (approved):

PR #2456, opened 2024-11-15
Emit all Treehouse guest changes at the end of a frame

JakeWharton:
Prior to this changes were emitted when the applier's `onEndChanges` callback was invoked. Compose can do multiple rounds of changes within a single setContent call or sendFrame call resulting in multiple begin/end change callbacks on the applier.

Refs #1902 

---


redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
```diff
@@ -66,6 +66,13 @@ internal class NodeApplier<W : Any>(
     }
   }
 
+  override fun onBeginChanges() {
```
swankjesse, line 69: Does this defer changes by 16ms?

swankjesse (approved):

redwood-compose/src/commonMain/kotlin/app/cash/redwood/compose/WidgetApplier.kt
JakeWharton: Since the callback is now called earlier, it means you have to rely on the frame ending to know when to send changes. Since most of the time there's only a single set of changes performed to the applier, it's mostly a no-op.

PR #2458, opened 2024-11-18
Host-side API for view-owned widgets

squarejesse:
This introduces a new mutable property, windowInsets, that is propagated to the composition where it can be consumed.

We put the burden on the user of RedwoodView to set this value to something non-zero if they have insets to be consumed. I've added an accelerator on Android for this,

```kotlin
RedwoodLayout.consumeWindowInsets(mask: Int)
```

### DIFF
--- a/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/UiConfiguration.kt
+++ b/redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/UiConfiguration.kt
@@ -25,13 +25,23 @@ public class UiConfiguration(
    * True if the device is configured to use a dark color palette.
    */
   public val darkMode: Boolean = false,
+
   /**
-   * A set of distances from the edges of the display where system UI and other elements will be
-   * drawn.
-   *
-   * Use this margin to inset your content to avoid drawing under system UI elements.
+   * The insets of the host window, independent of where the Redwood composition is positioned
+   * within it. The Redwood composition is not responsible for consuming these insets.
    */
   public val safeAreaInsets: Margin = Margin.Zero,
+
+  /**
+   * The insets of the viewport that the composition is responsible for consuming.
+   *
+   * This may be zero if the host view isn't attached to a view hierarchy and therefore doesn't
+   * know its insets.
+   *
+   * See https://developer.android.com/develop/ui/views/layout/edge-to-edge
+   */
+  public val viewInsets: Margin = Margin.Zero,
+
   /**
    * The size of the viewport into which the composition is rendering. This could be as lage as the
    * entire screen or as small as an individual view within a larger native screen.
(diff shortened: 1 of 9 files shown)


redwood-widget/src/androidMain/kotlin/app/cash/redwood/widget/RedwoodLayout.kt
```diff
@@ -112,7 +118,7 @@ public open class RedwoodLayout(
 
   private fun computeUiConfiguration(
     config: Configuration = context.resources.configuration,
-    insets: Insets = rootWindowInsetsCompat.safeDrawing,
```
squarejesse, line 115: # ☠️

This is the core problem of the old API. If we end up calling this function for a configuration change, it forgets what the *view-local* insets were and reverts to the *window* insets.

redwood-runtime/src/commonMain/kotlin/app/cash/redwood/ui/UiConfiguration.kt
```diff
@@ -25,13 +25,23 @@ public class UiConfiguration(
...
+
+  /**
+   * The insets of the viewport that the composition is responsible for consuming.
+   *
+   * This may be zero if the host view isn't attached to a view hierarchy and therefore doesn't
+   * know its insets.
+   *
+   * See https://developer.android.com/develop/ui/views/layout/edge-to-edge
+   */
+  public val viewInsets: Margin = Margin.Zero,
```
squarejesse, line 43: Unfortunately `safeAreaInsets` are potentially non-zero on already-shipping clients, but they don’t necessarily apply to the view we want them to.

Instead I’m creating another property that’s my attempt at fixing the window-local vs. view-local problem.

redwood-treehouse-host/src/androidUnitTest/kotlin/app/cash/redwood/treehouse/TreehouseLayoutTest.kt
```diff
@@ -112,7 +115,9 @@ class TreehouseLayoutTest {
           bottom = insets.bottom.toDp(),
         )
       }
-      assertThat(awaitItem().safeAreaInsets).isEqualTo(expectedInsets)
+      val value2 = awaitItem()
+      assertThat(value2.safeAreaInsets).isEqualTo(Margin.Zero)
+      assertThat(value2.viewInsets).isEqualTo(expectedInsets)
```
JakeWharton, line 120: Change log item for this behavior change? Probably want to give a heads-up to internal users as well, if they still exist, that is.

JakeWharton (approved):

redwood-treehouse-host/src/androidUnitTest/kotlin/app/cash/redwood/treehouse/TreehouseLayoutTest.kt
squarejesse: Yep, lemme do.

Issue #919, opened 2023-04-18
Support for consumable host safe area insets


colinrtwhite:
Follow up from: https://github.com/cashapp/redwood/pull/913

Currently, we're exposing host safe area insets through `LocalHostConfiguration.current.safeAreaInsets`. Long term we'd like to use a different strategy, copying [Jake's comment](https://github.com/cashapp/redwood/pull/913#pullrequestreview-1388106302):

Long term, I think we should change the approach to be more like Compose UI. This would mean two things:

- The insets are available as their own composition local within pure Redwood (not Treehouse).
- The insets are consumable rather than static so that they reflect the composables position in the native UI toolkit view hierarchy.

Can we use the same API as Compose UI in Redwood? My first thought is we can't really use layout modifiers given the parent widget needs to understand them for them to have an effect.

JakeWharton:
Yeah I'm not strictly tied to the exact Compose UI API–I actually don't even know what it looks like. More of the consumable mechanism that it uses and how it works within Compose, presumably through a composition local.

chrisbanes:
Insets is tricky for anything multi-platform as Android has a million inset types (actually ~10) and other platforms have 1 (or zero). So we don't want to pollute the API just for the sake of Android being extremely granular.

A composition local could work here, but it would have slightly different guarantees to the way Android works. Consuming some insets would then stop them being 'dispatched' to children, but there's siblings to think about too. Off the top of my head, a composition local would mean that siblings would all get to consume some insets at the same level in the 'stack', which could lead to insets being applied multiple times. Probably fine but something to check.

squarejesse:
This is done.

PR #2462, opened 2024-11-19
Host-side insets for iOS UIViews

swankjesse:


swankjesse:
https://github.com/cashapp/redwood/issues/919

redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIView.kt
```diff
@@ -54,11 +55,14 @@ public open class RedwoodUIView : RedwoodView<UIView> {
   override val children: Widget.Children<UIView>
     get() = _children
 
+  private val density = Density.Default
```
JakeWharton:
```suggestion
  private val density: Density get() = Density.Default
```
No need to burn a backing field on a value that never changes

JakeWharton (approved):

swankjesse:
The newly-added test fails like this in CI:

```
Child process terminated with signal 5: Trace/BPT trap
```

I haven’t been able to extract more information than that. I can’t catch an exception or anything. And it passes locally. 

I assume something I’m doing with the UIWindow is failing 'cause either the code to do it isn’t there (a linking problem), or because the environment it’s running in isn’t there (headless?).

squarejesse:
https://publicobject.com/2024/11/20/testing-safe-area-insets-on-ios/

PR #2464, opened 2024-11-19
Use the new insets API in EmojiSearch

swankjesse:


redwood-lazylayout-uiview/src/commonMain/kotlin/app/cash/redwood/lazylayout/uiview/UIViewLazyList.kt
```diff
@@ -275,6 +276,7 @@ internal class UIViewLazyList :
       rowHeight = UITableViewAutomaticDimension
       separatorStyle = UITableViewCellSeparatorStyleNone
       backgroundColor = UIColor.clearColor
+      contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentNever
```
swankjesse, line 279: Without this, `UIScrollView` adds the appropriate amount of space after the last element automatically, and we get double the space that we want.

swankjesse: (Though I didn’t see this behavior in `Column(overflow = Scroll)`, which is surprising)

swankjesse:
# Android

We draw all the way to the bottom edge of the screen:
Screenshot_20241119_155128

And put space under the last element so it isn’t obscured by the bottom line thing:
Screenshot_20241119_155219


# iOS

Draw all the way to the bottom edge:
Simulator Screenshot - iPhone 15 Pro - 2024-11-19 at 16 28 28

Space under the last element:
Simulator Screenshot - iPhone 15 Pro - 2024-11-19 at 16 28 02

JakeWharton (approved):

PR #2491, opened 2024-12-04
New API, RedwoodLayout.additionalInsets

swankjesse:
Consumers can use this to impose additional insets for system bars and floating action buttons.

---


swankjesse:
I would prefer to do this with a built-in mechanism, but I couldn’t find one.

My consuming code doesn’t urgently need this for iOS, but the asymmetry bothers me. Should I add a similar property for iOS?

Issue #2248, opened 2024-08-14
Disposal callbacks aren't called when disposing of a TreehouseUi


dellisd:
The following does not work when a TreehouseUi screen is closed/disposed of on both Android and iOS:
```kotlin
DisposableEffect(Unit) {
  println("Hello")
  
  onDispose {
    // Never called!
    println("World!")
  }
}
```

A workaround is to use a `LaunchedEffect`:
```kotlin
LaunchedEffect(Unit) {
  try {
    println("Hello")
    awaitCancellation()
  } finally {
    // This prints fine
    println("World")
  }
}
```

JakeWharton:
If you are tearing down the whole Zipline instance, what's probably happening is that this is running after the binding through which logs are sent is removed.

dellisd:
When you say Zipline instance, do you mean for a single screen?

My understanding was that a Zipline instance would survive for the lifetime of the treehouse _app_, but I'm seeing this issue when tearing down a single screen, e.g. navigating back from a treehouse screen.

JakeWharton:
I don't know how it's wired internally, but I would expect each screen to be a Zipline instance, yes. I can look next week.

swankjesse:
The Zipline instance spans multiple screens, so I think this is likely a lifecycle step I’ve missed in the Treehouse part of Redwood.

swankjesse:
Fixed.

PR #2494, opened 2024-12-09
Dispose of the composition when we're done with it.

swankjesse:
Initially I'm adding just the test we already pass, where the effect is executed because it is removed from the composition.

https://github.com/cashapp/redwood/issues/2248


redwood-treehouse-host/src/appsJvmTest/kotlin/app/cash/redwood/treehouse/leaks/JvmHeap.kt
```diff
@@ -79,6 +80,7 @@ internal object JvmHeap : Heap {
 
       // Don't traverse further on types that are unlikely to contain application-scoped data.
       // We want to avoid loading the entire application heap into memory!
+      instance is AtomicLongFieldUpdater<*> -> listOf()
```
swankjesse: Tests demanded this

JakeWharton (approved):

PR #2503, opened 2024-12-11
Remove widgets individually in the protocol

JakeWharton:
This is required to efficiently implement `movableContentOf` in a future change.

See #1902 

---

### DIFF
--- a/redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
+++ b/redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
@@ -16,8 +16,10 @@
 package app.cash.redwood.protocol.guest
 
 import androidx.compose.runtime.BroadcastFrameClock
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
 import app.cash.redwood.compose.WidgetVersion
 import app.cash.redwood.protocol.Change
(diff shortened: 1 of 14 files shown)


redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
```diff
@@ -288,6 +291,46 @@ class ProtocolTest {
...
+      .containsExactly(
+        ChildrenChange.Remove(Id.Root, ChildrenTag.Root, 2, false),
+        ChildrenChange.Remove(Id.Root, ChildrenTag.Root, 1, false),
+        ChildrenChange.Remove(Id.Root, ChildrenTag.Root, 0, false),
+      )
+  }
+
+  @Suppress("DEPRECATION") // Testing for old behavior.
+  @Test
+  fun removeRangeForOldVersions() = runTest {
```
squarejesse, line 305: nice

redwood-treehouse-guest/src/jsMain/kotlin/app/cash/redwood/treehouse/ProtocolBridgeJs.kt
```diff
@@ -208,7 +214,13 @@ internal class FastGuestProtocolAdapter(
     val tag = tag
     val index = index
     val count = count
-    changes.push(js("""["remove",{"id":id,"tag":tag,"index":index,"count":count}]"""))
+    if (hostSupportsRemoveDetachAndNoCount) {
```
squarejesse, line 217: I assume in follow-up we'll change the caller to use a different function?

squarejesse (approved):

redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
JakeWharton: I'm actually thinking this behavior switching is pointless now. I broke this change out of the large one, but it's always valid to send individual removes to any host. So we always can do it and only worry about whether or not it's valid to send `detach=true` or not.

JakeWharton: Just going to leave it for now, land the rest which are kinda stacked on it, and then maybe delete.

JakeWharton: Yeah I got rid of this in #2509.

PR #2509, opened 2024-12-12
Defer removing guest nodes until change list is realized

JakeWharton:
In the future, they may be re-added before the change list is captured and thus shouldn't be removed.

For #1902 

---


redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/ProtocolWidgetChildren.kt
```diff
@@ -37,13 +37,10 @@ public class ProtocolWidgetChildren(
   }
 
   override fun remove(index: Int, count: Int) {
-    for (i in index until index + count) {
-      val widget = _widgets[i]
-      guestAdapter.removeWidget(widget.id)
-      widget.depthFirstWalk(guestAdapter.childrenRemover)
+    for (i in index + count - 1 downTo index) {
+      val widget = widgets[i]
+      guestAdapter.appendRemove(id, tag, index, widget.id)
```
JakeWharton, line 42: Notable changes here:
- We do not actually remove the widget anymore, only record the remove change
- We unconditionally emit single-widget removes

redwood-protocol/src/commonTest/kotlin/app/cash/redwood/protocol/ProtocolTest.kt
```diff
@@ -92,8 +91,8 @@ class ProtocolTest {
       """["create",{"id":1,"tag":2}],""" +
       """["add",{"id":1,"tag":2,"childId":3,"index":4}],""" +
       """["move",{"id":1,"tag":2,"fromIndex":3,"toIndex":4,"count":5}],""" +
-      """["remove",{"id":4,"tag":3,"index":2,"count":10}],""" +
-      """["remove",{"id":5,"tag":3,"index":2,"detach":true}],""" +
+      """["remove",{"id":4,"tag":3,"index":2,"count":1}],""" +
+      """["remove",{"id":5,"tag":3,"index":2,"count":1,"detach":true}],""" +
```
JakeWharton, line 95: - Not being clever trying to omit `count`. We can start omitting it when we drop support for <0.17.0 in a year.
- Removing the default for count property to ensure that we always send the count, even if it's 1 for compatibility with <0.17.0 hosts. Should have never changed this in the previous PR.

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/DefaultGuestProtocolAdapter.kt
```diff
@@ -156,22 +157,24 @@ public class DefaultGuestProtocolAdapter(
...
 
   /** Returns the changes accumulated since the last call to this function. */
   public fun takeChanges(): List<Change> {
     val result = changes.toList()
     this.changes.clear()
+
+    for (id in removed) {
+      val widget = widgets.remove(id)
+        ?: throw IllegalStateException("Removed widget not present in map: $id")
+      widget.depthFirstWalk(childrenRemover)
```
squarejesse, line 174: Whoa

redwood-protocol/src/commonTest/kotlin/app/cash/redwood/protocol/ProtocolTest.kt
squarejesse: Simple

squarejesse (approved):

squarejesse:
I gotta do something about that flaky test

JakeWharton:
It's not so much flaky as the emulator is dead slow on CI. We can probably bump the timeout, but 10s is already an eternity. I guess what's another 10 seconds in timeout?

JakeWharton:
Maybe we could get money to use a hosted service? This shit is load bearing, sure we can scrape a few dollars to make our CI better...

PR #2510, opened 2024-12-13
Update re-attached widget removes to use detach

### LINKED ISSUE #1902: Support `movableContentOf` across protocol

JakeWharton:
This ensures that the host-side maintains their instances to correctly support re-attach.

Closes #1902 

[movable.webm](https://github.com/user-attachments/assets/744ccb40-0f9c-40e9-acaf-afe8b6684056)

---


colinrtwhite (approved):
Nice!

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/DefaultGuestProtocolAdapter.kt
```diff
@@ -136,9 +137,18 @@ public class DefaultGuestProtocolAdapter(
...
-      "Attempted to add widget with ID ${child.id} but one already exists"
+    val childId = child.id.value
+    if (child.removeIndex != INVALID_INDEX) {
+      check(hostSupportsRemoveDetach) { "Host v$hostVersion does not support widget re-attach" }
+      check(childId in widgets) { "Attempted to re-attach unknown widget with ID $childId" }
+      removed.remove(childId)
+
+      // Update the remove change to indicate it's only a detach.
+      val remove = changes[child.removeIndex] as ChildrenChange.Remove
+      changes[child.removeIndex] = ChildrenChange.Remove(remove.id, remove.tag, remove.index, true)
```
squarejesse: name the parameter `true` ?

redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
```diff
@@ -288,6 +295,171 @@ class ProtocolTest {
...
+          }
+        }
+      }
+      when (state) {
+        0 -> Row { oneTwoThree() }
+        1 -> Column { oneTwoThree() }
+      }
+    }
+
+    assertThat(composition.awaitSnapshot().filter { it !is ValueChange }).containsExactly(
```
squarejesse, line 318: Beautiful!

redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
```diff
@@ -288,6 +295,171 @@ class ProtocolTest {
...
+      ChildrenChange.Add(Id(2), ChildrenTag(1), Id(4), 1),
+      // Text("three") to Box
+      Create(Id(5), WidgetTag(3)),
+      ChildrenChange.Add(Id(2), ChildrenTag(1), Id(5), 2),
+      // Box to Row
+      ChildrenChange.Add(Id(1), ChildrenTag(1), Id(2), 0),
+    )
+
+    state = 1
+    assertThat(composition.awaitSnapshot().filter { it !is ValueChange }).containsExactly(
```
squarejesse, line 338: The `ValueChange` instances all target the new `Column` here, right?

redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
```diff
@@ -288,6 +295,171 @@ class ProtocolTest {
...
+      // Box from Row
+      ChildrenChange.Remove(Id(1), ChildrenTag(1), 0, detach = true),
```
squarejesse, line 340: YASS

redwood-protocol-guest/src/commonTest/kotlin/app/cash/redwood/protocol/guest/ProtocolTest.kt
```diff
@@ -288,6 +295,171 @@ class ProtocolTest {
...
+      ChildrenChange.Remove(Id.Root, ChildrenTag.Root, 0, detach = false),
+      // Column to Root
+      Create(Id(6), WidgetTag(1000002)),
+      ChildrenChange.Add(Id.Root, ChildrenTag.Root, Id(6), 0),
+      // Box to Column
+      ChildrenChange.Add(Id(6), ChildrenTag(1), Id(2), 0),
+    )
+  }
+
+  @Test fun multipleMovableContentButOnlyOneReused() = runTest {
```
squarejesse, line 351: So cool

squarejesse (approved):

squarejesse:
You made it look easy

redwood-tooling-codegen/src/main/kotlin/app/cash/redwood/tooling/codegen/protocolGuestGeneration.kt
```diff
@@ -194,6 +194,7 @@ internal class ProtocolButton(
 ) : ProtocolWidget, Button<Unit> {
   public override val id: Id = guestAdapter.nextId()
   public override val tag: WidgetTag get() = WidgetTag(3)
+  public override var removeIndex: Int = InvalidRemoveIndex
```
JakeWharton:
```suggestion
  public override var removeIndex: Int = INVALID_INDEX
```

redwood-protocol-guest/src/commonMain/kotlin/app/cash/redwood/protocol/guest/DefaultGuestProtocolAdapter.kt
JakeWharton:
```suggestion
      changes[child.removeIndex] = ChildrenChange.Remove(remove.id, remove.tag, remove.index, detach = true)
```

JakeWharton:
> You made it look easy

Extracting like 8 infrastructure PRs out from before it certainly helped!

PR #2548, opened 2025-01-28
Add `setContent` variant which returns initial snapshot

JakeWharton:
---

### DIFF
--- a/redwood-testing/src/commonTest/kotlin/app/cash/redwood/testing/TestRedwoodCompositionTest.kt
+++ b/redwood-testing/src/commonTest/kotlin/app/cash/redwood/testing/TestRedwoodCompositionTest.kt
@@ -15,8 +15,10 @@
  */
 package app.cash.redwood.testing
 
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.movableContentOf
+import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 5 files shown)


redwood-testing/src/commonTest/kotlin/app/cash/redwood/testing/TestRedwoodCompositionTest.kt
```diff
@@ -28,12 +30,51 @@ import app.cash.redwood.widget.MutableListChildren
...
+      var number by mutableIntStateOf(0)
+      val initial = setContentAndSnapshot {
+        Text("The number is: $number")
+        LaunchedEffect(Unit) {
+          number = 1
+        }
+      }
+
+      // Defer to allow effect to run.
+      delay(10.milliseconds)
```
squarejesse, line 54: Would `yield()` be sufficient?

squarejesse: I guess your goal is to guarantee any amount of yielding would still not get that value

squarejesse (approved):

redwood-testing/src/commonTest/kotlin/app/cash/redwood/testing/TestRedwoodCompositionTest.kt
JakeWharton: Yes. This is a stand in for arbitrary code that absolutely causes a dispatch. Yield is probably enough, but might not be. Could also be a `runCurrent()` which is basically what `delay` becomes in `runTest`.

PR #2555, opened 2025-02-07
Don't use the nested scroll view in Column

squarejesse:
This breaks overscroll.

I looked into the PR that added this, which was an attempt to work-around a problem where a RecyclerView was nested in a scrollable column. We shouldn't ever do that, so this work-around should be unnecessary.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Changed:
 - Nothing yet!
 
 Fixed:
-- Nothing yet!
+- Don't use NestedScrollView for Android's Overflow.Scroll. It doesn't implement overscroll as well.
 
 
 ## [0.17.0] - 2025-01-30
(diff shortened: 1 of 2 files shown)


squarejesse:
Related: https://github.com/cashapp/redwood/pull/746

JakeWharton (approved):
If an NSV isn't overscrolling it would be because a parent reports itself as being able to consume the scroll events. Is something else scrollable up in the hierarchy? I will still approve, but since NSV is used everywhere normally anything like this should be some other problem that we created.

squarejesse:
Lemme figure out what’s happening and follow-up here.

JakeWharton:
We're unlikely to actually _need_ NSV, unless we attempt something like a collapsing header component. So it's not a huge deal to switch for now. Just was a surprising culprit.

PR #2582, opened 2025-03-07
Support marking widget composble functions as internal

### LINKED ISSUE #2580: Allow marking a schema widget as generating an `internal` function
This requires you to refine that function to a more friendly public API in the module which contains the Compose codegen.

Use cases:
* Defining a more complex protocol between the guest and host, but hiding it from users
* Migration of existing widgets to different APIs while maintaining compatiblity
  * #2124

JakeWharton:
Closes #2580 

---


redwood-lazylayout-schema/src/main/kotlin/app/cash/redwood/lazylayout/widgets.kt
```diff
@@ -31,7 +31,7 @@ import app.cash.redwood.ui.Margin
  *
  * @see LazyList
  */
-@Widget(1)
+@Widget(1, internalComposable = true)
 public data class LazyList(
```
JakeWharton, line 35: Turns out we already needed this!

swankjesse (approved):

PR #2589, opened 2025-03-19
Switch test app to using basic `Text`

### LINKED ISSUE #2275: Add "basic" widget set for samples, prototyping, first steps
If you want to start using Redwood today you need to build a schema and widget bindings for literally everything. It is paramount that you do this, but it's not important if you're just getting started and trying things out. Instead, we can provide a "basic" set of widgets, similar to those we create in our own samples, test-app, and tests. This will also greatly simplify our samples and test-app!

To be included:
- [x] Text widget
- [x] Button widget
- [x] TextInput widget
- [x] Image widget
- [ ] #2602

The ability to support the test-app and counter is a good litmus test for the design of these schema types. We want them to feel useful enough to build simple things. We don't want them to be used in production by anything. Not sure if emoji-search should migrate or not... we might want that to be our pretty sample. Maybe we add a calculator sample or something to show something that still can be built with these basic widgets but is also slightly more complicated.

JakeWharton:
Closes #2275

---


JakeWharton:
Always forget about the damn Swift...

colinrtwhite (approved):

swankjesse (approved):

PR #2639, opened 2025-05-08
Split the host protocol from the widget system

### LINKED ISSUE #2360: Migrate protocol host codegen to be entirely stateless
With respect to the native UI tree and even node map. We basically are generating a reflection-like set of descriptors which the host adapter is using to manipulate the native UI tree and node map. But the two should not be coupled directly as they are now.

The protocol mismatch handler should also be implemented by the host adapter as it queries the descriptor, not directly in the code gen.

Writing this down so I don't forget...

JakeWharton:
This will allow us to access the protocol on the zipline thread in a future change.

Closes #2360 

---


JakeWharton:
Hmm this is missing some `Json` plumbing. A bit worrying so few tests fail...

JakeWharton:
Okay should be pretty close now.

squarejesse (approved):

squarejesse:
Just cutting down the line count is a good reason to do this.

PR #2663, opened 2025-05-28
Enforce event lambda types always return Unit

### LINKED ISSUE #2656: Reject non-`Unit` return types in schema event property lambdas
I swear we used to do this, but apparently it broke/doesn't.

JakeWharton:
Closes #2656

---


drewhamilton (approved):

squarejesse (approved):
I love how the tests work

PR #2697, opened 2025-07-17
Set up file access for snapshot testing

swankjesse:
I'm hoping to get something working where we can use our regular Kotlin test harness.


JakeWharton (approved):
So is the plan to use element capture API and then do the image comparison all in the browser? I wonder if you'll have better luck inverting the relationship and having the browser ask the host to do the capture and compare.

swankjesse:
Yep. Let me see what obstacles I run into with this approach. If I hit a brick wall I might try something else.

PR #2731, opened 2025-08-14
Fix RedwoodUIView to implement sizeThatFits correctly

swankjesse:
We had delegated to UIStackView which works great if the RedwoodUIView is full screen, but which fails awkwardly if the RedwoodUIView is in a layout that calls through sizeThatFits.

Also fix RedwoodUIView to propagate size changes regardless of which layout system we're using:
 - autolayout, invalidated by setNeedsLayout().
 - SwiftUI, invalidated by invalidateIntrinsicContentSize().

---

### DIFF
--- a/redwood-snapshot-testing/src/iosMain/kotlin/app/cash/redwood/snapshot/testing/UIViewTestWidgetFactory.kt
+++ b/redwood-snapshot-testing/src/iosMain/kotlin/app/cash/redwood/snapshot/testing/UIViewTestWidgetFactory.kt
@@ -21,8 +21,10 @@ import app.cash.redwood.ui.Density
 import app.cash.redwood.ui.Dp
 import app.cash.redwood.widget.ResizableWidget
 import app.cash.redwood.widget.ResizableWidget.SizeListener
+import kotlin.math.max
 import kotlinx.cinterop.CValue
 import kotlinx.cinterop.readValue
+import kotlinx.cinterop.useContents
 import platform.CoreGraphics.CGRectZero
 import platform.CoreGraphics.CGSize
 import platform.CoreGraphics.CGSizeMake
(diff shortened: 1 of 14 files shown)


redwood-widget-uiview-test/RedwoodWidgetUIViewTests/__Snapshots__/UIViewRedwoodViewFillWrapTestHost/testSingleChildElement.1.png
swankjesse, line 1: Without the code changes in `RedwoodUIView`, this gets a height of 0

image

redwood-snapshot-testing/src/iosMain/kotlin/app/cash/redwood/snapshot/testing/UIViewTestWidgetFactory.kt
```diff
@@ -116,7 +118,20 @@ class UIViewColor :
 class UIViewSimpleColumn : SimpleColumn<UIView> {
   override var modifier: Modifier = Modifier
 
-  override val value = UIStackView(CGRectZero.readValue()).apply {
+  override val value = object : UIStackView(CGRectZero.readValue()) {
+    override fun intrinsicContentSize(): CValue<CGSize> {
```
swankjesse, line 122: We had *two* problematic `UIStackViews` in our codebase, and this is the 2nd one. This was the simplest way to make this test make sense when the caller is using `intrinsicContentSize()`.

redwood-widget-uiview-test/src/commonTest/kotlin/app/cash/redwood/widget/uiview/UIViewRedwoodViewFillWrapTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import app.cash.redwood.widget.Widget
+import platform.UIKit.UIView
+
+/**
+ * This is like [UIViewRedwoodViewTest] but it sizes the component differently.
+ *
+ * We had a bug where `RedwoodUIView` incorrectly returned height=0 from [UIView.sizeThatFits]
+ * because its internal [platform.UIKit.UIStackView] didn't implement that operation.
+ */
+class UIViewRedwoodViewFillWrapTest(
```
swankjesse, line 34: There isn’t a `WrapWrapTest` or a `WrapFillTest` yet, but that’s just cause I don’t have urgency for those.

JakeWharton (approved):

Issue #2619, opened 2025-04-15
Split `Widget` responsibility into `Binding` and `NewWidget`


JakeWharton:
`Binding` is the current `Widget`. `NewWidget` is a class that holds state for things like unscoped modifier's changing the view type, diffing modifier changes, and tree walking. Then, eventually, rename `NewWidget` to `Widget`.

JakeWharton:
Definitely struggling to make this work as a simple change. It may be worth combining `NewWidget` into `WidgetNode` as the tree type, where that type becomes the tree node.

However, it's still not clear how the new `Binding` type (which is hyper-local and deals only with adapting data to the UI node) should talk to the `NewWidget` for things like size notification.

Issue #2232, opened 2024-08-02
Add a mechanism to walk the `Widget` tree


JakeWharton:
We want parents to be able to act on the entire subtree. We have this at the protocol-layer, but we want it at the `Widget` layer.

Ideally it's something on `Widget`, like an `allChildren: List<Widget.Children<W>>` or maybe even `allChildren: Sequence<Widget<W>>`.

However, because of Objective-C export, we cannot add generated functions to a `Widget` sub-interface as they appear unimplemented in the header. So we're blocked by ~#2231~ #2619.

PR #2745, opened 2025-08-20
New API, Widget.allChildren

squarejesse:
This interface function exposes the live Widget.Children collections for generic tree-traversal code.

I'd like to use that capability to implement widget focus by searching for a widget with the appropriate modifier.

---

### DIFF
--- a/redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/Widget.kt
+++ b/redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/Widget.kt
@@ -39,6 +39,13 @@ public interface Widget<W : Any> {
   public var modifier: Modifier
 
   /**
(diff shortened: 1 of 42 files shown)


redwood-widget/src/commonMain/kotlin/app/cash/redwood/widget/Widget.kt
squarejesse, line 1: This is the actual API change! Everything else is just making it work.

JakeWharton (approved):

squarejesse:
Incredible ouch, I can’t enjoy the benefits of interface functions’ default implementations from Swift.

I have an idea to combat this.

JakeWharton:
I wouldn't spend too much time being clever. This will move to the NewWidget as part of https://github.com/cashapp/redwood/issues/2619 and bindings will never see it. This whole feature is tracked by https://github.com/cashapp/redwood/issues/2232.

PR #2746, opened 2025-08-20
Scaffolding for the new redwood-ui-core module

squarejesse:
This is initially a placeholder for the forthcoming focus director API. It may also be useful for other, similar modifiers.


redwood-ui-core-api/src/commonMain/kotlin/app/cash/redwood/ui/core/api/FocusRequester.kt
squarejesse, line 1: not the final form of this type!

test-app/schema/src/main/kotlin/com/example/redwood/testapp/schema.kt
```diff
@@ -45,8 +46,9 @@ import kotlin.time.Duration
   ],
   dependencies = [
     Dependency(1, RedwoodUiBasic::class),
-    Dependency(2, RedwoodLayout::class),
-    Dependency(3, RedwoodLazyLayout::class),
+    Dependency(2, RedwoodUiCore::class),
+    Dependency(3, RedwoodLayout::class),
+    Dependency(4, RedwoodLazyLayout::class),
```
squarejesse: In general I don’t think we can renumber these.

JakeWharton (approved):
Good with this as a start.

I think for the future:
- UI basic should depend on UI core
- We might be able to eliminate the "redwood-runtime" and merge it with UI core into its "api" module. I'm not sure there's much difference in them.

test-app/schema/src/main/kotlin/com/example/redwood/testapp/schema.kt
JakeWharton: Indeed. I'll confirm they're tracked in the schema API file.

JakeWharton: Whoops! We don't track it for normal schemas! https://github.com/cashapp/redwood/issues/2754

PR #2750, opened 2025-08-21
Fix RedwoodUIView measurement using intrinsicContentSize

squarejesse:
When I recently changed this to not use UIStackView as its parent component, I introduced a regression where measurements made with intrinsicContentSize() didn't honor the frame's dimensions.

---

### DIFF
--- a/redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIView.kt
+++ b/redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIView.kt
@@ -45,8 +45,14 @@ import platform.UIKit.UIUserInterfaceLayoutDirection.UIUserInterfaceLayoutDirect
 import platform.UIKit.UIUserInterfaceLayoutDirection.UIUserInterfaceLayoutDirectionRightToLeft
 import platform.UIKit.UIUserInterfaceStyle
 import platform.UIKit.UIView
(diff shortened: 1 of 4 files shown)


redwood-widget/src/iosMain/kotlin/app/cash/redwood/widget/RedwoodUIView.kt
```diff
@@ -169,8 +189,60 @@ public open class RedwoodUIView : RedwoodView<UIView> {
...
+     * label is before we can compute that label's height.
+     *
+     * We work around this by:
+     *
+     *  1. Making [intrinsicContentSize] depend on the mostly-recently applied frame
+     *  2. Invalidating it each time the frame changes
+     *
+     * This will result in an additional layout pass when the superview uses [intrinsicContentSize].
+     */
+    private fun updateFrameForIntrinsicSize(frame: CValue<CGRect>) {
```
squarejesse, line 214: This technique is lifted from `YogaUIView`. I suspect I can stop using it there as long as that component always has this one as an ancestor.

squarejesse: Or rather, we can get rid of it once we never use `intrinsicContentSize()` on our views.

JakeWharton (approved):

PR #2756, opened 2025-08-22
Web snapshot tests

swankjesse:


rqcai (approved):

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.w3c.dom.Element
+
+public class DomSnapshotter @PublishedApi internal constructor(
+  private val path: String,
+) {
+  private val snapshotStore = SnapshotStore()
+
+  public suspend fun snapshot(
+    element: Element,
+    name: String? = "snapshot",
```
JakeWharton:
```suggestion
    name: String = "snapshot",
```
`null` is never used. It's just turned back into "snapshot" down below.

JakeWharton (approved):

PR #2757, opened 2025-08-22
Stop configuring the max width and height in YogaUIView

### LINKED ISSUE #2753: Bad layouts when mixing flex(1.0) inside a column that uses height=Constraint.Wrap
I don’t like what we render with this layout:

```kotlin
        Box(
          height = Constraint.Fill,
          width = Constraint.Fill,
        ) {
          Box(
            modifier = Modifier.size(300.dp, 600.dp),
            margin = insets,
            height = Constraint.Fill,
            width = Constraint.Fill,
          ) {
            Column(
              width = Constraint.Fill,
            ) {
              Box(
                modifier = Modifier.flex(1.0),
              ) {
                Text(text = "hello")
              }
            }
          }
        }
```

It gives the Text element height = 0.

Dropping the flex modifier on the enclosing Box fixes it.

squarejesse:
These attributes only cause us pain.

In particular they aren't necessary to pass any of our layout tests, and by removing them I can fix a layout bug.

Note that on Android we do not use these attributes, so this brings the two platforms closer together.

Closes: https://github.com/cashapp/redwood/issues/2753

---


dnagler (approved):

JakeWharton:
You might need to use an older simulator to generate the snapshot because of the GitHub runner debacle. I can upgrade us to macOS 15, Xcode 16.4, and iOS 18.4 next week.

JakeWharton:
Or just `@Ignore` with a TODO and I'll un-ignore next week when I regenerate them all...

redwood-lazylayout-uiview/.DS_Store
JakeWharton, line 1:
```
$ git config --global core.excludesfile ~/.gitignore
$ echo ".DS_Store" > ~/.gitignore
```

JakeWharton:
Until I do the switch, I was able to generate new screenshots for my PR by switching to Xcode 15.4 and using iOS 17.5 as the simulator.

squarejesse:
I generated the screenshots okay on my Mac.

JakeWharton:
I think something with text hinting changed in iOS 18 so when we bump lots of the goldens change. I had to roll back to Xcode 15.4 and iOS 17.5 for mine.

PR #2762, opened 2025-08-22
Eliminate the test-only modifier on Compose UI `Spacer`

JakeWharton:
---


JakeWharton:
Probably have to regenerate iOS since I change the default spacer color

JakeWharton:
Unfortunately I think this is still wrong.

PR #2773, opened 2025-08-23
Verify snapshot

rqcai:


redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -40,9 +50,75 @@ public class DomSnapshotter @PublishedApi internal constructor(
       },
     ).await()
 
-    snapshotStore.put("$path/${name ?: "snapshot"}.png", image)
+    val fileName = "$path/${name ?: "snapshot"}.png"
```
JakeWharton: This is unrelated to your PR, but was changed to be non-null in Jesse's PR and I guess we missed updating this reference.
```suggestion
    val fileName = "$path/$name.png"
```

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -40,9 +50,75 @@ public class DomSnapshotter @PublishedApi internal constructor(
...
+
+    snapshotStore.getBlob(fileName)?.let { existing ->
+      check(existing.contentEquals(image)) {
+        "Current snapshot does not match the existing file $fileName"
+      }
+    } ?: snapshotStore.put(fileName, image)
   }
 
+  private suspend fun Blob.contentEquals(other: Blob): Boolean {
```
JakeWharton, line 62: We may have to implement more advanced strategies in the future if the variation across machines turns out to be a problem. Something like https://zschuessler.github.io/DeltaE/ maybe.

JakeWharton (approved):

squarejesse (commented):
This is awesome

PR #2778, opened 2025-08-26
Revert a bunch of widget subclass splits from 'layout'

JakeWharton:
This was when we thought we could use `abstract class`es for `Widget`s (#2231), which we cannot. Re-combining these saves a tiny bit of memory.

---


JakeWharton:
There's actually one more to revert in the 'layout' schema, but it's very messy so I'll probably do that some other time.

squarejesse (approved):

PR #2784, opened 2025-08-29
Start using Burst's TestInterceptor in snapshot tests

squarejesse:


gradle/libs.versions.toml
```diff
@@ -11,7 +11,7 @@ paparazzi = "1.3.2"
 zipline = "1.22.0"
 coil = "3.3.0"
 okio = "3.16.0"
-burst = "2.6.0"
+burst = "2.7.1"
```
squarejesse, line 14: Still happening: https://github.com/cashapp/burst/actions/runs/17330400444

JakeWharton (approved):
As a follow up I think we'll want to start using the test name from Burst and supplying it to the platform snapshotter rather than relying on Swift's macro and Paparazzi's stack reflection. It'll be nice to have all the tests use the same exact name so we can trivially compare across toolkits.

PR #2792, opened 2025-09-04
Make Snapshotter.snapshot() suspending

swankjesse:
This makes all of the snapshot-using tests suspending.

This requires an update to Burst, which previously didn't suspend nicely for Kotlin/JS.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -11,7 +11,7 @@ paparazzi = "1.3.2"
 zipline = "1.22.0"
 coil = "3.3.0"
 okio = "3.16.0"
-burst = "2.7.1"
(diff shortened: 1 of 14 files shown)


gradle/libs.versions.toml
```diff
@@ -11,7 +11,7 @@ paparazzi = "1.3.2"
 zipline = "1.22.0"
 coil = "3.3.0"
 okio = "3.16.0"
-burst = "2.7.1"
+burst = "2.8.1"
```
swankjesse, line 14: I shipped 2.8.0 foolishly, and I need this fix before this PR runs correctly:
https://github.com/cashapp/burst/pull/187

JakeWharton (approved):

PR #2798, opened 2025-09-08
Avoid copy-pasting the snapshot-testing.js file

swankjesse:
We need it in each project we do snapshot testing in, but I'm unaware of a way to load it from anywhere but the current project's directory.


rqcai (approved):

build-support/src/main/kotlin/app/cash/redwood/buildsupport/WriteSnapshotTestingJsTask.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.gradle.api.tasks.TaskAction
+
+/**
+ * Write snapshot-testing.js to the current project's karma.config.d directory. (Unfortunately
+ * there's no API to tell Karma to load extensions from a build directory.)
+ */
+internal abstract class WriteSnapshotTestingJsTask : DefaultTask() {
+  @get:OutputFile
+  val snapshotTestingTargetFile: File by lazy {
+    project.projectDir.resolve("karma.config.d/generated-redwood-snapshot-testing.js")
```
JakeWharton: This breaks configuration caching.

Should be defined as
```kotlin
@get:OutputDirectory
abstract val projectDirectory: DirectoryProperty
```
and set at the registration site with
```
projectDirectory.set(project.layout.projectDirectory)
```

JakeWharton (approved):

build-support/src/main/kotlin/app/cash/redwood/buildsupport/WriteSnapshotTestingJsTask.kt
swankjesse: Oooh I had no idea. Thanks!

JakeWharton: This is the link I was trying to find: https://docs.gradle.org/current/userguide/configuration_cache_requirements.html#config_cache:requirements:use_project_during_execution. I knew it was an issue but couldn't find documentation for it.

PR #2799, opened 2025-09-09
Consistently frame the target element

swankjesse:
Support a target element that isn't already a child of the document's root element.

When snapshotting, wrap the element in a frame, and then remove the element from that frame afterwards. This avoids mutating the passed-in element as a
side-effect of snapshotting.


redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -37,27 +37,50 @@ public class DomSnapshotter @PublishedApi internal constructor(
...
     )
+    wrapper.appendChild(element)
+    document.documentElement!!.appendChild(wrapper)
+
+    val image = try {
+      val boundingClientRect = wrapper.getBoundingClientRect()
+      HtmlToImage.toBlob(
+        element = element,
+        options = Options().apply {
+          this.backgroundColor = backgroundColor
```
rqcai: I suppose you want to use the background color and pixel ratio from frame

rqcai (approved):

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/Frame.kt
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.redwood.dom.testing
+
+/**
+ * Configure the canvas that we will snapshot our element on.
+ */
+public class Frame(
```
JakeWharton, line 21: Hmm I don't think this should live here. Or at least should immediately move after merging.

I think this should live in the shared test infrastructure so that each abstract test class can define their default frame (or even per-snapshot frames). This results in each platform producing images that are the same size and (should) look the same.

redwood-dom-testing/src/commonTest/kotlin/app/cash/redwood/dom/testing/DomSnapshotterSampleTest.kt
```diff
@@ -29,32 +29,69 @@ import kotlinx.dom.clear
...
+            |top: 10px;
+            |right: 20px;
+            """.trimMargin(),
+          )
+        },
+      )
+    }
+
+    snapshotter.snapshot(yellowRect, "exactSizeWithFrame_None", Frame.None)
+    snapshotter.snapshot(yellowRect, "exactSizeWithFrame_Iphone14", Frame.Iphone14)
```
JakeWharton: No `burstValues`?

JakeWharton (approved):

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
swankjesse: Ooooh yep. Thank you!

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/Frame.kt
swankjesse: Fantastic idea. Will do in follow-up.

redwood-dom-testing/src/commonTest/kotlin/app/cash/redwood/dom/testing/DomSnapshotterSampleTest.kt
swankjesse: Great idea. Done.

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
swankjesse: Turns out I don’t need the background color at all. The way the HTML-to-image library works, it just clobbers the component’s own background color and I don’t want to offer two competing ways to do the same thing.

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/Frame.kt
JakeWharton: https://github.com/cashapp/redwood/issues/2806

PR #2803, opened 2025-09-10
Generate diff images for web snapshots

rqcai:


.github/workflows/build.yaml
```diff
@@ -154,6 +154,15 @@ jobs:
 
       - run: xcodebuild -project redwood-widget-uiview-test/RedwoodWidgetUIViewTests.xcodeproj -scheme RedwoodWidgetUIViewTests -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.4' test
 
+      - run: ./gradlew jsBrowserTest
```
JakeWharton: This runs the tests for all modules (which have browser tests enabled). You probably want to target this more locally to specific modules.

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -69,6 +72,7 @@ public class DomSnapshotter @PublishedApi internal constructor(
           this.canvasWidth = this.width
           this.canvasHeight = this.height
           this.pixelRatio = frame.pixelRatio
+          this.backgroundColor = "transparent"
```
rqcai, line 75: redwood-layout-dom ci was failing because the default background color is white instead of transparent. Here I'm explicitly setting the background color to transparent so the snapshot background color would be consistent.

testBothSpacer diff from https://github.com/cashapp/redwood/actions/runs/17656504943/artifacts/3990317185

JakeWharton (approved):
So cool!

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -83,57 +87,145 @@ public class DomSnapshotter @PublishedApi internal constructor(
...
-      return true
+      if (differentPixels == 0L) {
+        return DiffResult(isDifferent = false)
+      }
+
+      // Draw delta image in the middle
+      ctx.putImageData(deltaData, maxWidth.toDouble(), 0.0)
+
+      // Convert canvas to blob
+      val deltaBlob = suspendCancellableCoroutine<Blob> { continuation ->
```
squarejesse, line 186: I don’t see why this call is necessary?

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -83,57 +87,145 @@ public class DomSnapshotter @PublishedApi internal constructor(
...
+          },
+          "image/png",
+        )
+      }
+
+      // Calculate percentage difference
+      val total = maxHeight.toLong() * maxWidth.toLong() * 3L * 256L
+      var percentDifference = (delta * 100 / total.toDouble()).toFloat()
+
+      // Fallback to pixel difference if color delta is 0 but pixels are different
```
squarejesse: This case is for when it’s only different on the alpha channel?

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -83,57 +87,145 @@ public class DomSnapshotter @PublishedApi internal constructor(
...
+      if (differentPixels > 0 && percentDifference == 0f) {
+        percentDifference = (differentPixels * 100 / (maxWidth * maxHeight).toDouble()).toFloat()
```
squarejesse: Should percentDifference be the max of these two measurements?

val percentDifferenceRgb = ...
val percentDifferenceAlpha = ...
val percentDifference = maxOf(percentDifferenceRgb, percentDifferenceAlpha)

?

As is if the alpha difference is 100% but the pixel difference is 0.1%, we’ll report 0.1%.

(That gets to be a practical problem when we support stuff like fuzzy-matching and accept a tolerance of .1%)

squarejesse (approved):
I love it.

In a follow-up would you mind extracting this new code into a new class like `ImageDiffer` or something, and writing a couple test cases for it directly? I know testing tests is a bit yak-shaving, but in this case the code is subtle enough that I think it’s warranted. Plus I want to extract this code into a standalone library for everyone to use!

squarejesse:
In particular I’d like test cases to cover things like
 * Images match exactly
 * Images differ on RGB by 10%, and confirm we get a 10% back on percent difference
 * Images match on alpha by 10%, and confirm we get a 10% back on percent difference

You can probably use code like this to create sample images, but altering sizes or whatever to create diffs.
```
    val yellowRect = document.createElement("div").apply {
      setAttribute(
        "style",
        """
        |background-color: #ffff00;
        |width: 200px;
        |height: 100px;
        |position: relative
        """.trimMargin(),
      )
      appendChild(
        document.createElement("div").apply {
          setAttribute(
            "style",
            """
            |background-color: #0000ff;
            |position: absolute;
            |width: 50px;
            |height: 25px;
            |top: 10px;
            |right: 20px;
            """.trimMargin(),
          )
        },
      )
    }
```

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
rqcai: we need to convert the delta image to blob so we can write to file.

redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
rqcai: I think you meant take percentDifferenceRgb if it's not 0 else percentDifferenceAlpha

PR #2812, opened 2025-09-12
Extract image differ code into class

rqcai:


swankjesse (approved):
Perfect test. Thank you!

PR #2813, opened 2025-09-12
Put a white background behind layout snapshots

swankjesse:
We're already doing this on Android Views and Compose UI.


redwood-dom-testing/src/commonMain/kotlin/app/cash/redwood/dom/testing/DomSnapshotter.kt
```diff
@@ -72,7 +72,6 @@ public class DomSnapshotter @PublishedApi internal constructor(
           this.canvasWidth = this.width
           this.canvasHeight = this.height
           this.pixelRatio = frame.pixelRatio
-          this.backgroundColor = "transparent"
```
swankjesse, line 75: It turns out this is annoying elsewhere ’cause if any other background is explicitly set, it clears it.

JakeWharton (approved):

PR #2818, opened 2025-09-15
Update dom snapshot output directory

rqcai:
Put snapshots in `/src/test/snapshots/`, diff.png and html files in `/build/snapshots/`


swankjesse (approved):
Nice!

PR #2821, opened 2025-09-17
Use a CSS grid as the layout for Box

swankjesse:
This is a better fit for measuring nested child elements.


JakeWharton (approved):
Pretty funny

swankjesse:
Before, items with wrap were measured as 0px tall because all their children were absolute positioned.

testChildren_Wrap_Center_Center

PR #2823, opened 2025-09-19
Install UI core modifiers for Gradle plugin tests

JakeWharton:
The generator plugin adds a dependency on redwood-compose which depends on redwood-ui-core-modifiers.

---


JakeWharton:
This will fix PRs like #2820 and #2822.

squarejesse (approved):

PR #2842, opened 2025-10-06
Upgrade to AGP 9 

JakeWharton:
We have to disable all the new hotness because we're stuck on an old Paparazzi for now (although its necessary changes remain unreleased at this time).

---

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -71,7 +71,6 @@ jobs:
           kotlinUpgradeYarnLock
           test
           jvmTest
-          -x testReleaseUnitTest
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yaml
```diff
@@ -71,7 +71,6 @@ jobs:
           kotlinUpgradeYarnLock
           test
           jvmTest
-          -x testReleaseUnitTest
```
JakeWharton, line 74: AGP 9 only enables debug tests by default

swankjesse (approved):

swankjesse (approved):

PR #2889, opened 2025-10-31
Fix image diff to properly handle different image sizes

rqcai:
Previously had swapped loop bounds (y used minWidth, x used minHeight) and
only compared overlapping region, causing incomplete diff visualization.


JakeWharton (approved):

JakeWharton:
Weird failure. I'll look into it Monday.

PR #2899, opened 2026-01-08
Fix memory leaks in Redwood UIView components

jhollida24:
This PR fixes multiple memory leaks in Redwood's iOS/UIView implementation that prevented UI components from being deallocated after they were removed from the view hierarchy (e.g., after logout).

I am _not_ a Kotlin, GC, or Treehouse expert. This is the result of a local redwood -> cash-ios build, a test case of opening several receipts, and a conversation with an LLM that can see both sets of source about the references to leaked objects I can see using the iOS memory graph.

This _does_ get rid of a large number of leaked objects on the iOS side.

Root Cause
Kotlin/Native uses reference counting similar to ARC, but lambdas and anonymous objects create strong references by default. Several components were creating retain cycles that prevented proper cleanup.

Fixes
1. UIViewBox and UIViewFlexContainer - Used WeakReference to break retain cycles in SizeListener callbacks that were capturing parent containers strongly.

2. RedwoodUIView - Used WeakReference in the sizeListener to prevent capturing valueRootView strongly.

3. HostProtocolAdapter - The widgetSystem was being retained even after close() was called, creating a Kotlin→Swift reference chain that kept Swift objects alive. Made widgetSystem nullable and cleared it in close() to break this chain.

Impact
These leaks affected all Treehouse-based screens in Cash iOS. After logout, UI components and their associated graphics memory (VM: Memory Tag objects) remained in memory indefinitely. With these fixes, all components are now properly deallocated when screens are dismissed.

This was tested with Receipts on the activity tab.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## Unreleased
+
+Fixed:
+- Fix memory leaks in UIView layout components (`UIViewBox`, `UIViewFlexContainer`, `RedwoodUIView`)
+- Fix memory leak in `HostProtocolAdapter`
+
+These leaks affected some Treehouse-based screens and prevented UI components and their associated memory from being released after dismissal.
+
 ## [0.19.0]
 [0.19.0]: https://github.com/cashapp/redwood/releases/tag/0.19.0
 
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
@@ -31,6 +31,8 @@ import app.cash.redwood.ui.Margin
 import app.cash.redwood.widget.ResizableWidget
 import app.cash.redwood.widget.UIViewChildren
 import app.cash.redwood.widget.Widget
+import kotlin.experimental.ExperimentalNativeApi
+import kotlin.native.ref.WeakReference
 import kotlinx.cinterop.CValue
 import kotlinx.cinterop.convert
 import kotlinx.cinterop.readValue
@@ -89,13 +91,15 @@ internal class UIViewBox :
     var sizeListener: ResizableWidget.SizeListener? = null
     private val measurer = Measurer()
 
+    @OptIn(ExperimentalNativeApi::class)
     val children = UIViewChildren(
       container = this,
       insert = { index, widget ->
         if (widget is ResizableWidget<*>) {
+          val weakView = WeakReference(this@View)
           widget.sizeListener = object : ResizableWidget.SizeListener {
             override fun invalidateSize() {
-              this@View.invalidateSize()
+              weakView.get()?.invalidateSize()
             }
           }
         }
--- a/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
+++ b/redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
@@ -27,6 +27,8 @@ import app.cash.redwood.widget.ResizableWidget.SizeListener
 import app.cash.redwood.widget.UIViewChildren
 import app.cash.redwood.yoga.FlexDirection
 import app.cash.redwood.yoga.Node
+import kotlin.experimental.ExperimentalNativeApi
+import kotlin.native.ref.WeakReference
 import kotlinx.cinterop.convert
 import platform.UIKit.UIView
 import platform.darwin.NSInteger
@@ -119,15 +121,18 @@ private fun Node(view: UIView): Node {
   return result
 }
 
+@OptIn(ExperimentalNativeApi::class)
 private class NodeSizeListener(
   private val node: Node,
   private val view: UIView,
-  private val enclosing: UIViewFlexContainer,
+  enclosing: UIViewFlexContainer,
 ) : SizeListener {
+  private val weakEnclosing = WeakReference(enclosing)
+
   override fun invalidateSize() {
     if (node.markDirty()) {
       view.setNeedsLayout()
-      enclosing.invalidateSize()
+      weakEnclosing.get()?.invalidateSize()
     }
   }
 }
--- a/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
+++ b/redwood-protocol-host/src/commonMain/kotlin/app/cash/redwood/protocol/host/HostProtocolAdapter.kt
@@ -49,7 +49,7 @@ public class HostProtocolAdapter<W : Any>(
   guestVersion: RedwoodVersion,
   container: Widget.Children<W>,
   protocol: HostProtocol,
-  private val widgetSystem: WidgetSystem<W>,
+  widgetSystem: WidgetSystem<W>,
   private val eventSink: UiEventSink,
   private val leakDetector: LeakDetector,
 ) : UiChangesSink {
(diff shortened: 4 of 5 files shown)


redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
```diff
@@ -89,13 +90,15 @@ internal class UIViewBox :
     var sizeListener: ResizableWidget.SizeListener? = null
     private val measurer = Measurer()
 
+    @OptIn(kotlin.experimental.ExperimentalNativeApi::class)
```
jhollida24: I am _not_ clear if this is an okay thing to do, and I'm hoping that someone who knows this area better than I can weigh in.

jhollida24:
Build methodology, if you wanted to try this out locally:

### 1. Configure cash-treehouse to use local Redwood

Create or edit `~/Development/cash-treehouse/local.properties`:

```properties
# Enable Kotlin/Native to build iOS framework
treehouse.kotlin.native.enabled=true

# Point to your local Redwood repository
redwoodPath=/Users/jholliday/Development/redwood
```

### 2. Build cash-treehouse with local Redwood changes

```bash
cd ~/Development/cash-treehouse
./gradlew debugSpm --no-configuration-cache
```

This will:
- Build the CashKt XCFramework with your local Redwood changes
- Automatically update `cash-ios/ThirdParty/SPM/Package.swift` to point to the local build
- Output the framework to `platform/host-ios/build/XCFrameworks/debug/CashKt.xcframework`

### 3. Resolve SPM dependencies in cash-ios

```bash
cd ~/Development/cash-ios
cash spm resolve
```

### 4. Build and run Cash iOS

Build and run from Xcode

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
dnagler: This seems reasonable to me--`WeakReference` was added as an internal `ExperimentalNativeApi` in #2191 cc @JakeWharton

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewFlexContainer.kt
```diff
@@ -119,15 +120,18 @@ private fun Node(view: UIView): Node {
...
 }
 
+@OptIn(kotlin.experimental.ExperimentalNativeApi::class)
 private class NodeSizeListener(
   private val node: Node,
   private val view: UIView,
-  private val enclosing: UIViewFlexContainer,
+  enclosing: UIViewFlexContainer,
 ) : SizeListener {
+  private val weakEnclosing = WeakReference(enclosing)
```
dnagler, line 130: Ooh, great catch here, this would be a strict retain cycle otherwise

dnagler (approved):
These changes make sense to me! I think it's worth getting at least one more set of eyes on them just to be safe though.

Before merging, it'd be good to add a note in CHANGELOG.md. I think @colinrtwhite should be able to create a 0.19.1 release?

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
colinrtwhite: Yep it should be fine to use `ExperimentalNativeApi`s as an implementation detail (like here).

jhollida24:
> These changes make sense to me! I think it's worth getting at least one more set of eyes on them just to be safe though.
> 
> Before merging, it'd be good to add a note in CHANGELOG.md. I think @colinrtwhite should be able to create a 0.19.1 release?

Added a 0.19.1 section in the Changelog

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
dellisd: A style nit here would be to import this type to avoid having the package name written out in the annotation

dellisd (approved):
This looks good to me

colinrtwhite (approved):
Looks good to me, though I'm curious why the size listeners aren't being freed when the associated UI component is removed from the view hierarchy - maybe some ARC intricacies I don't understand.

> HostProtocolAdapter - The widgetSystem was being retained even after close() was called, creating a Kotlin→Swift reference chain that kept Swift objects alive

Is it possible that this change is enough to fix the issue on its own?

jhollida24:
> Looks good to me, though I'm curious why the size listeners aren't being freed when the associated UI component is removed from the view hierarchy - maybe some ARC intricacies I don't understand.
> 
> > HostProtocolAdapter - The widgetSystem was being retained even after close() was called, creating a Kotlin→Swift reference chain that kept Swift objects alive
> 
> Is it possible that this change is enough to fix the issue on its own?

I tried testing with just that commit, rebasing the others away. The leaks of the `UIViewBox` (and a whole bunch of other objects) reappear when I do this.

(The Backtrace here is when this object was allocated)

Screenshot 2026-01-08 at 2 28 59 PM

redwood-layout-uiview/src/commonMain/kotlin/app/cash/redwood/layout/uiview/UIViewBox.kt
jhollida24: Done

jhollida24: Awesome. Thanks for verifying!

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Change Log
 
+## [0.19.1] - 2026-01-08
+[0.19.1]: https://github.com/cashapp/redwood/releases/tag/0.19.1
```
JakeWharton:
```suggestion
## Unreleased
```

Normally there's already an unreleased section here, but I didn't add it when releasing 0.19.0 because I don't expect there to ever be another public release. The version, date, and link are inserted as part of the release process.

You're probably better off just doing an internal release to use this and any future changes. I would not bother doing public releases. But I'm still supportive of documenting the changes using the normal process.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
+
+Fixed:
+- Fix memory leaks in UIView layout components (`UIViewBox`, `UIViewFlexContainer`, `RedwoodUIView`) where `SizeListener` callbacks were creating retain cycles by capturing parent containers strongly. Now use `WeakReference` to break these cycles.
+- Fix memory leak in `HostProtocolAdapter` where the `widgetSystem` was retained after `close()` was called, creating a Kotlin→Swift reference chain that prevented Swift objects from being deallocated. The `widgetSystem` is now cleared in `close()` to break this chain.
+
+These leaks affected some Treehouse-based screens and prevented UI components and their associated memory from being released after dismissal.
```
JakeWharton:
```suggestion
- Fix memory leaks in UIView layout components (`UIViewBox`, `UIViewFlexContainer`, `RedwoodUIView`) where `SizeListener` callbacks were creating retain cycles by capturing parent containers strongly.
- Fix memory leak in `HostProtocolAdapter` where the `widgetSystem` was retained after `close()` was called, creating a Kotlin→Swift reference chain that prevented Swift objects from being deallocated.
```

This log is for users who aren't expected to know implementation details. Those are better served in the commit message for us developers.

JakeWharton (approved):
Curious as to why the built-in leak detection didn't catch these cases. Normally I'd want that blind spot fixed to prevent regression, but it's probably not worth the effort at this point.

CHANGELOG.md
jhollida24: Done.

CHANGELOG.md
jhollida24: Condensed. The same details are indeed in the individual commits.

jhollida24:
> Curious as to why the built-in leak detection didn't catch these cases. Normally I'd want that blind spot fixed to prevent regression, but it's probably not worth the effort at this point.

Yeah, probably not.

Thank you for taking a look at this. It's very much appreciated 🙏 .

